PR #396, opened 2026-04-16
fix: reject NaN, Inf, and boolean values in portfolio number parser

hitakshiA:
## Summary

- _to_num() now rejects NaN, Inf, and boolean inputs that previously corrupted portfolio math
- 35 tests cover the full input surface of the portfolio normalization pipeline

## Problem

_to_num() in normalize_v2.py accepted three classes of poison inputs:

1. **float('nan')** - passes isinstance(float) check, gets returned as-is. When a holding has NaN market_value (from corrupted PDF extraction), the entire allocation bucket becomes NaN. Downstream, the UI renders "NaN%" for asset allocation.

2. **float('inf')** - same path. Infinite values break percentage calculations and produce nonsensical portfolio totals.

3. **True/False** - bool is a subclass of int in Python, so isinstance(True, int) is True. A boolean JSON flag like is_cash_equivalent parsed as 1.0, creating phantom $1 holdings in the portfolio.

Reproduced all three:
- _to_num(float('nan')) returned nan (now returns None)
- _to_num(float('inf')) returned inf (now returns None)
- _to_num(True) returned 1.0 (now returns None)
- _allocation_mix_from_holdings with one NaN holding corrupted the entire bucket to nan

## Approach

Two guards added to _to_num(), 4 lines total:
- `isinstance(value, bool)` check before `isinstance(value, (int, float))` since bool is a subclass of int
- `math.isnan(f) or math.isinf(f)` check after float conversion

No behavior changes for valid inputs. All existing callers use `_to_num(x) or 0.0` which handles the new None returns identically.

## Test plan

35 tests in test_normalize_v2.py, all passing (0.06s):

- [x] 11 standard conversion tests (int, float, dollar strings, parens negatives, percentages)
- [x] 10 None-returning tests (None, empty, whitespace, N/A, double negative, symbols-only)
- [x] 5 poison input tests (NaN, Inf, -Inf, True, False)
- [x] 5 allocation mix tests (single bucket, unknown bucket, zero total, empty, NaN-does-not-corrupt)
- [x] 2 canonical portfolio builder tests (schema version, cash ledger filtering)
- [x] 2 analytics builder tests (concentration excludes zero, gain/loss distribution)

Closes: N/A (discovered during codebase audit, no existing issue)

### DIFF
--- a/consent-protocol/hushh_mcp/kai_import/normalize_v2.py
+++ b/consent-protocol/hushh_mcp/kai_import/normalize_v2.py
@@ -2,14 +2,20 @@
 
 from __future__ import annotations
 
+import math
 from collections import Counter
 from datetime import datetime, timezone
 from typing import Any
 
 
 def _to_num(value: Any) -> float | None:
+    if isinstance(value, bool):
+        return None
     if isinstance(value, (int, float)):
-        return float(value)
+        f = float(value)
+        if math.isnan(f) or math.isinf(f):
+            return None
+        return f
     if isinstance(value, str):
         cleaned = value.replace("$", "").replace(",", "").replace("%", "").strip()
         if not cleaned:
--- a/consent-protocol/tests/test_normalize_v2.py
+++ b/consent-protocol/tests/test_normalize_v2.py
@@ -1,0 +1,254 @@
+"""Tests for normalize_v2 portfolio normalization helpers.
+
+Covers _to_num edge cases (NaN, Inf, booleans, currency formats) and
+_allocation_mix_from_holdings bucket math.
+"""
+
+from __future__ import annotations
+
+import math
+
+import pytest
+
+from hushh_mcp.kai_import.normalize_v2 import (
+    _allocation_mix_from_holdings,
+    _to_num,
+    build_financial_analytics_v2,
+    build_financial_portfolio_canonical_v2,
+)
+
+
+# ---------------------------------------------------------------------------
+# _to_num: standard conversions
+# ---------------------------------------------------------------------------
+
+
+class TestToNumStandard:
+    def test_int(self) -> None:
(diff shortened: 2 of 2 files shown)


kushaltrivedi5:
Codex CI triage for the current head SHA:

- Head SHA: `df39f5f`
- `CI Status Gate` is failing on this PR.
- Failing checks on the current head: `Secret Scan`, `Protocol (Python)`, `CI Status Gate`.

What to fix:
- `Secret Scan`: Remove committed secret-like material from the branch. If any credential is real, rotate it. Then push a clean commit so the scan reruns on the new head SHA.
- `Protocol (Python)`: Run the backend lane locally with `./bin/hushh codex pre-pr` or `cd consent-protocol && uv sync --frozen --group dev && bash scripts/run-test-ci.sh` and fix the failing protocol checks.

After you push an updated commit, `PR Validation` should rerun automatically on the new head commit and update `CI Status Gate`.

kushaltrivedi5:
## Merged: Portfolio Number Parser Hardening

### What Landed
[`consent-protocol/hushh_mcp/kai_import/normalize_v2.py`](https://github.com/hushh-labs/hushh-research/blob/main/consent-protocol/hushh_mcp/kai_import/normalize_v2.py) now rejects `NaN`, infinity, and boolean values in the portfolio number parser, with regression coverage in [`consent-protocol/tests/test_normalize_v2.py`](https://github.com/hushh-labs/hushh-research/blob/main/consent-protocol/tests/test_normalize_v2.py).

### Proof
The required gate was green, the live parsing bug was still present on `main`, and the merge path completed through queue validation and post-merge smoke.

### Outcome
Kai import now fails invalid financial values before they enter holdings math. That keeps downstream analysis from treating non-finite or boolean input as real portfolio data.

PR #458, opened 2026-04-21
fix: reject NaN, Inf, and boolean values in quality_v2 number parsers

hitakshiA:
## Summary

- `_to_num` and `_coerce_optional_number` in `quality_v2.py` now reject NaN, Inf, and boolean inputs that previously corrupted quality-gate telemetry
- 33 tests cover the full input surface of the quality reporting pipeline

## Problem

Both parser helpers accepted three classes of poison inputs, same pattern as the just-merged #396 fix in `normalize_v2.py`:

1. **float('nan')** passed `isinstance(float)` and returned as-is. A holding with NaN market_value made `holdings_market_value_sum` (line 148) become NaN, corrupting the quality gate result that downstream consumers use for reconciliation and telemetry.
2. **float('inf')** same path. Infinite values break sum/percentage calculations.
3. **True/False** — bool is a subclass of int in Python, so `isinstance(True, int)` is True. A boolean JSON flag parsed as 1.0, inflating the sum and producing phantom $1 entries.

Reproduced all three:
- `_to_num(float('nan'))` returned nan (now returns None)
- `_coerce_optional_number(float('inf'))` returned inf (now returns None)
- `_to_num(True)` returned 1.0 (now returns None)
- `evaluate_import_quality_gate_v2` with one NaN holding produced `holdings_market_value_sum: nan` (now produces a clean numeric sum)

## Approach

Same guards as #396:
- Check `isinstance(value, bool)` before `isinstance(value, (int, float))` since bool is a subclass of int
- Reject NaN and Inf via `math.isnan() or math.isinf()`
- Apply both guards to the string branch in `_coerce_optional_number` after the float conversion

No behavior change for valid inputs. All existing callers use `x or 0.0` patterns which handle the new None returns identically.

## Test plan

33 tests in `test_quality_v2.py`, all passing locally (0.09s):

- [x] 6 standard conversion tests for `_to_num` (int, float, zero, negative, string-rejected, None)
- [x] 5 poison tests for `_to_num` (NaN, Inf, -Inf, True, False)
- [x] 9 standard tests for `_coerce_optional_number` (int, float, plain string, dollar string, negative, empty, whitespace, invalid, None)
- [x] 4 poison tests for `_coerce_optional_number`
- [x] 2 downstream tests proving NaN holdings no longer corrupt `holdings_market_value_sum` and `expected_total_value`
- [x] 4 severity tests for `evaluate_import_quality_gate_v2` (fail paths, warn path, healthy pass)
- [x] 3 report builder tests (percentage clamping, parser score, zero-division guard)
- [x] ruff check passes

Closes: N/A (discovered during audit, follows the #396 pattern)

### DIFF
--- a/consent-protocol/hushh_mcp/kai_import/quality_v2.py
+++ b/consent-protocol/hushh_mcp/kai_import/quality_v2.py
@@ -2,19 +2,30 @@
 
 from __future__ import annotations
 
+import math
 from collections import Counter
 from typing import Any
 
 
 def _to_num(value: Any) -> float | None:
+    if isinstance(value, bool):
+        return None
     if isinstance(value, (int, float)):
-        return float(value)
+        f = float(value)
+        if math.isnan(f) or math.isinf(f):
+            return None
+        return f
     return None
 
 
 def _coerce_optional_number(value: Any) -> float | None:
+    if isinstance(value, bool):
+        return None
     if isinstance(value, (int, float)):
-        return float(value)
+        f = float(value)
+        if math.isnan(f) or math.isinf(f):
+            return None
+        return f
     if value is None:
         return None
     if isinstance(value, str):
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (approved):
Approved and merged through the main queue on `4cfd50c9`.

This adds the right import-quality guardrail: `quality_v2` now rejects `NaN`, `Inf`, and booleans before they can poison quality telemetry. The change is narrow and covered at the parser surface.

PR #397, opened 2026-04-16
fix: remove credential fragments from client-side console logging

hitakshiA:
## Summary

- Removes 4 console.log calls that exposed token prefixes in browser DevTools
- Replaces with action-only log messages (no credential material)

## Problem

Found during a security sweep of client-side logging:

| File | What was logged | Risk |
|------|----------------|------|
| account-service.ts:45 | First 30 chars of VAULT_OWNER consent token | Consent token prefix in browser console |
| fcm-service.ts:377 | First 20 chars of FCM push token | Push token prefix in browser console |
| fcm-service.ts:598 | First 20 chars of FCM push token | Same, duplicate code path |
| fcm-service.ts:737 | First 20 chars of refreshed FCM token | Same, token refresh path |

These fragments are visible in DevTools, capturable by browser extensions (including malicious ones), and potentially forwarded by client-side observability tools. The VAULT_OWNER token prefix is especially sensitive since it is a consent-scoped credential.

## Approach

Replaced each call with an action-only message that confirms the operation happened without including any credential material:

- `"[AccountService] Deleting account with target:", target, "token:", vaultOwnerToken.substring(0, 30)` -> `"[AccountService] Deleting account with target:", target`
- `"[FCM] Got token:", token.substring(0, 20)` -> `"[FCM] Got push token"`
- `"[FCM] Token refreshed:", event.token.substring(0, 20)` -> `"[FCM] Push token refreshed"`

Also verified via grep that no other `.substring()` or `.slice()` patterns remain in console.log calls across the frontend codebase.

## Test plan

- [x] TypeScript typecheck passes (zero new errors; pre-existing voice-tts-playback.ts errors unchanged)
- [x] `grep -rn "console.log.*\.substring" hushh-webapp/` returns zero results after fix
- [x] 2 files changed, 4 insertions, 9 deletions
- [x] No secrets or env files in diff

PR #460, opened 2026-04-21
fix: remove user email, uid, and displayName from client-side console logs

hitakshiA:
## Summary

- Removes 4 client-side console.log calls that exposed user identity fragments in browser DevTools
- Extends the credential-log sweep started in #397 to user identity (email, uid, displayName)

## Problem

Sibling of merged PR #397. The previous sweep fixed token fragments; this one covers user identity fields that are equally sensitive in the browser console:

| File | What was logged | Risk |
|------|----------------|------|
| `lib/capacitor/plugins/auth-web.ts:52` | `user.email` on Google Sign-in success | Email in DevTools on every sign-in |
| `lib/capacitor/plugins/auth-web.ts:88` | `user.email` on Apple Sign-in success | Same, duplicate code path |
| `lib/firebase/auth-context.tsx:353` | `user?.uid` when setting native user | Firebase UID exposed on auth transitions |
| `lib/vault/prf-auth.ts:282` | `displayName` during passkey registration | Name exposed during WebAuthn flow |

These fragments are visible in DevTools, capturable by browser extensions, and forwardable by client-side observability tools. Under the BYOK trust model the client should not leak user identity through developer logging.

## Approach

Replace each call with an action-only message that confirms the operation without including identity material:

- `"Google Sign-in successful:", user.email` -> `"Google Sign-in successful"`
- `"Apple Sign-in successful:", user.email || "(hidden email)"` -> `"Apple Sign-in successful"`
- `"Manually setting Native User:", user?.uid` -> `"Manually setting Native User"`
- `"User:", displayName` -> removed

Also grepped for residual `.uid` / `.email` / `displayName` in console.log calls to confirm zero remain.

## Test plan

- [x] 3 files changed, 3 insertions, 4 deletions
- [x] Post-fix grep returns zero remaining matches for user identity in console.log
- [x] eslint clean (zero warnings) on all 3 files

Closes: N/A (follows #397 pattern)

### DIFF
--- a/hushh-webapp/lib/capacitor/plugins/auth-web.ts
+++ b/hushh-webapp/lib/capacitor/plugins/auth-web.ts
@@ -49,7 +49,7 @@ export class HushhAuthWeb implements HushhAuthPlugin {
       this.currentIdToken = idToken;
       this.currentAccessToken = accessToken;
       
-      console.log("✅ [HushhAuthWeb] Google Sign-in successful:", user.email);
+      console.log("✅ [HushhAuthWeb] Google Sign-in successful");
       
       return { idToken, accessToken, user };
     } catch (error: unknown) {
@@ -85,7 +85,7 @@ export class HushhAuthWeb implements HushhAuthPlugin {
       this.currentUser = user;
       this.currentIdToken = idToken;
       
-      console.log("✅ [HushhAuthWeb] Apple Sign-in successful:", user.email || "(hidden email)");
+      console.log("✅ [HushhAuthWeb] Apple Sign-in successful");
       
       return { idToken, user };
     } catch (error: unknown) {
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (approved):

kushaltrivedi5:
## Merged: Client Identity Log Hygiene

### What Landed
The client-side auth and vault paths no longer print user email, UID, or display name in browser console logs. The final merge preserved the newer auth-context flow on `main` and kept the PR focused on removing identity fragments from logs.

### Why It Matters
These logs are visible in DevTools and can be captured by browser extensions, screenshots, support sessions, or observability tooling. Keeping the log messages action-only gives maintainers useful runtime breadcrumbs without exposing user identity.

### Maintainer Patch
I rebased the PR onto current `main` and resolved the auth-context conflict by preserving `applyAuthUser(...)` and removing only the UID-bearing log argument. No auth behavior was changed.

### Outcome
This closes the frontend identity-log gap for the touched auth surfaces. The current `main` smoke after the follow-up maintainer merge is green, so the branch is steady with this change included.

PR #409, opened 2026-04-16
Fix 408 agent execute scope

0xnithin-ai:
This PR addresses the critical authentication failure preventing the Kai agent from running (Bug #408). 

- Added the missing `AGENT_KAI_EXECUTE` to the `ConsentScope` enum in `constants.py`.
- Added the `agent.kai.execute` mapping to `_AGENT_SCOPE_MAP` in `scope_helpers.py` to correctly resolve the display orphaned key.
- Changed `resolve_scope_to_enum` to throw a strict `ValueError` instead of silently escalating invalid agent scopes to `AGENT_KAI_ANALYZE`.
- Narrowed the exception handling in `token.py:validate_token()` so that underlying system bugs (like an `AttributeError`) propagate instead of being swallowed disguised as forged tokens.

## Type

- [x] Bug fix
- [ ] Feature (new agent, operon, endpoint)
- [ ] Enhancement (existing functionality)
- [ ] Documentation
- [ ] Refactor (no behavior change)

### Required for all PRs
- [x] `ruff check .` passes
- [x] `ruff format --check .` passes
- [x] `mypy --config-file pyproject.toml` passes *(errors are pre-existing unaffected lines)*
- [x] `pytest tests/ -v` passes
- [x] `bandit -r hushh_mcp/ api/ -c pyproject.toml` passes

### If adding/modifying agents or tools
- [x] Consent is validated at agent entry (`HushhAgent.run()`)
- [x] Consent is validated at each tool invocation (`@hushh_tool`)

## Testing

Tested via the previously isolated local reproduction script (`test_bug2_real.py`). The script confirmed the original failures (`AttributeError` from the enum, `ValueError` from downstream resolution) and confirmed that subsequent validations correctly resolve without swallowing exceptions into an opaque authentication token "Malformed" error. 

### Bug Reproduction Output

```text
[TEST 1] Accessing ConsentScope.AGENT_EXECUTE (does not exist)
----------------------------------------------------------------------     
  AttributeError: type object 'ConsentScope' has no attribute 'AGENT_EXECUTE'
Traceback (most recent call last):
  File "C:\Users\Nithin\Desktop\hushh\hushh-research\consent-protocol\test_bug2_real.py", line 27, in <module>
    _ = ConsentScope.AGENT_EXECUTE          # <--   the buggy access
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #462, opened 2026-04-21
fix: remove orphan agent.kai.execute metadata from scope_helpers

hitakshiA:
## Summary

- Removes a stale `_STATIC_SCOPE_META` entry in `scope_helpers.py` for `"agent.kai.execute"`, a scope that no longer exists in the `ConsentScope` enum
- Adds a regression test so future drift between static metadata and the enum is caught

## Problem

`get_scope_display_metadata` keeps a dict of rich metadata (label, description, icon, color) for static scopes. The dict had an entry for `"agent.kai.execute"`, but the `ConsentScope` enum only defines analyze / debate / infer / chat for Kai, not execute.

```
grep "agent.kai" hushh_mcp/constants.py
  AGENT_KAI_ANALYZE = "agent.kai.analyze"
  AGENT_KAI_DEBATE  = "agent.kai.debate"
  AGENT_KAI_INFER   = "agent.kai.infer"
  AGENT_KAI_CHAT    = "agent.kai.chat"
```

The orphan entry meant any consent UI that called `get_scope_display_metadata("agent.kai.execute")` would render "Kai Actions" with a zap icon for a scope the system cannot validate or grant. That creates a mismatch between what users see and what the consent system enforces.

Also verified the scope is not referenced anywhere else in the repo before removing.

## Approach

1. Remove the `"agent.kai.execute"` entry from `_STATIC_SCOPE_META`
2. Add `test_scope_metadata_contract.py` with 8 tests:
   - Every key in the known static metadata set corresponds to a real `ConsentScope.value` (prevents this drift from recurring)
   - Each remaining known scope (`vault.owner`, `pkm.read`, `pkm.write`, `agent.kai.analyze`) returns rich metadata with exact label / icon / color
   - `agent.kai.execute` falls through to the generic default formatter after removal
   - Unknown scopes fall through to the generic formatter
   - Shape contract: every result includes `label`, `description`, `icon_name`, `color_hex`

## Test plan

- [x] 8/8 tests pass locally (0.11s)
- [x] ruff check passes
- [x] Grep confirms no other reference to `agent.kai.execute` in the codebase

Closes: N/A (discovered during audit)

### DIFF
--- a/consent-protocol/hushh_mcp/consent/scope_helpers.py
+++ b/consent-protocol/hushh_mcp/consent/scope_helpers.py
@@ -170,12 +170,6 @@ def get_scope_display_metadata(scope: str) -> dict:
             "icon_name": "brain",
             "color_hex": "#D4AF37",
         },
-        "agent.kai.execute": {
-            "label": "Kai Actions",
-            "description": "Allow Kai agent to execute actions",
-            "icon_name": "zap",
-            "color_hex": "#D4AF37",
-        },
     }
 
     meta = _STATIC_SCOPE_META.get(scope)
--- a/consent-protocol/tests/test_scope_metadata_contract.py
+++ b/consent-protocol/tests/test_scope_metadata_contract.py
@@ -1,0 +1,108 @@
+"""Regression tests for scope metadata contract.
+
+Ensures get_scope_display_metadata returns rich metadata only for scopes
+that actually exist as ConsentScope enum members, preventing the drift
+that left agent.kai.execute metadata behind after the enum dropped it.
+"""
+
+from __future__ import annotations
+
+from hushh_mcp.consent.scope_helpers import get_scope_display_metadata
+from hushh_mcp.constants import ConsentScope
+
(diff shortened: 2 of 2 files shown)


kushaltrivedi5:
## Closed: Opposite Scope Decision Resolved

### Decision
`#409` made `agent.kai.execute` canonical. Removing its metadata here would reintroduce a consent UI/runtime mismatch in the opposite direction.

### What We Kept
The drift concern from this PR was valid. `#409` absorbed that guardrail with tests proving the scope exists, resolves to the correct enum, appears in Kai agent scopes, validates in tokens, and rejects unknown agent/static scopes.

### Proof
`#409` merged on `main` at `276ddb08`, Main Post-Merge Smoke passed, and Consent Protocol upstream CI passed at `a09bfdcf`.

### Outcome
The repo now has one settled Kai execute-scope direction. Closing this prevents the scope metadata from being removed after the runtime and consent UI have aligned around it.

PR #463, opened 2026-04-21
fix: reject bool and string NaN/Inf in _parse_live_number extract path

hitakshiA:
## Summary

- `_parse_live_number` in `extract_v2.py` now rejects booleans and string-form NaN/Inf that previously leaked through the live portfolio extraction path
- 28 tests cover standard numeric conversions, None paths, float poison, bool poison, and every string non-finite variant

## Problem

`_parse_live_number` is called on every incoming holding row during Kai portfolio import extraction. Two poison input classes leaked through:

1. **Booleans:** `isinstance(True, int)` is `True` in Python, so `bool True` / `False` parsed as `1.0` / `0.0`. A JSON flag like `is_cash_equivalent` coming through the extraction pipeline would materialize as a phantom $1 holding.

2. **String-form NaN / Inf:** Python's `float("inf")` and `float("nan")` are valid. The string branch called `float(cleaned)` inside try/except but had no finiteness check afterward, so `"inf"`, `"nan"`, `"Infinity"`, `"-inf"`, and even `"(inf)"` (which went through the parens-negate path as `-inf`) all leaked through into holding `market_value` fields.

Reproduced all of these on current main:
- `_parse_live_number(True)` returned `1.0`
- `_parse_live_number("inf")` returned `inf`
- `_parse_live_number("NaN")` returned `nan`
- `_parse_live_number("(inf)")` returned `-inf`

Same poison-input class fixed in merged PR #396 (normalize_v2) and open PR #458 (quality_v2). Completing the sweep on the live extraction path.

## Approach

Two-line guards in `_parse_live_number`, mirroring the pattern from #396:

1. Add `isinstance(raw, bool)` check before `isinstance(raw, (int, float))` since bool is a subclass of int
2. Add `math.isfinite(value)` check after `float(cleaned)` on the string branch

The float branch already guards NaN/Inf via `math.isfinite(value)` so no change there. No behavior change for valid inputs.

## Test plan

28 tests in `test_extract_v2_parse_live_number.py`, all passing locally (0.03s):

- [x] 9 standard conversion tests (int, float, zero, negative, dollar string, parens negative, dollar parens negative, percentage, negative string)
- [x] 6 None-returning tests (None, empty, whitespace, invalid, empty parens, just dollar)
- [x] 3 float poison tests (NaN, Inf, -Inf already handled)
- [x] 2 bool poison tests (True, False)
- [x] 8 string poison tests (inf, Inf, Infinity, -inf, nan, NaN, -nan, (inf))
- [x] ruff check passes

Closes: N/A (follows #396 pattern, completes the live-extraction branch of the sweep)

### DIFF
--- a/consent-protocol/hushh_mcp/kai_import/extract_v2.py
+++ b/consent-protocol/hushh_mcp/kai_import/extract_v2.py
@@ -101,6 +101,8 @@ def _phase_progress_from_chunks_v2(phase: str, chunk_count: int) -> float:
 def _parse_live_number(raw: Any) -> float | None:
     if raw is None:
         return None
+    if isinstance(raw, bool):
+        return None
     if isinstance(raw, (int, float)):
         value = float(raw)
         return value if math.isfinite(value) else None
@@ -121,6 +123,8 @@ def _parse_live_number(raw: Any) -> float | None:
     try:
         value = float(cleaned)
     except Exception:
+        return None
+    if not math.isfinite(value):
         return None
     return -value if negative else value
 
--- a/consent-protocol/tests/test_extract_v2_parse_live_number.py
+++ b/consent-protocol/tests/test_extract_v2_parse_live_number.py
@@ -1,0 +1,140 @@
+"""Tests for _parse_live_number in extract_v2.
+
+Covers the live portfolio extraction parser used on every incoming
+holding row. Ensures NaN, Inf, and boolean inputs cannot reach
+downstream reconciliation or quality math.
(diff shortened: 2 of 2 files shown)


kushaltrivedi5 (approved):
Approved and merged through the main queue on `01ec152f`.

This closes the same numeric-poison class in the live extraction path. `_parse_live_number` now rejects booleans and string-form `NaN`/`Inf` values before they can leak into portfolio extraction.

Issue #408, opened 2026-04-16
[BUG] valid Kai tokens rejected as "Malformed"


0xnithin-ai:
**Severity:** auth layer affected

## What's wrong

`scope_helpers.py` line 179 references `"agent.kai.execute"` in its display metadata,
but this scope **does not exist** in the `ConsentScope` enum in `constants.py`.

When any code calls `ConsentScope("agent.kai.execute")` it raises:
```
ValueError: 'agent.kai.execute' is not a valid ConsentScope
```

`validate_token()` catches this with a blanket `except Exception` and returns:
```
(False, "Malformed token: ...", None)
```

So a **perfectly valid** `agent.kai.analyze` token looks forged. Kai is broken for all users.

## To reproduce

```bash
python test_bug2_real.py
```

## Expected

`validate_token()` with a valid token returns `(True, ..., payload)`.

## Fix

Either add the missing enum member:
```python
# constants.py
AGENT_KAI_EXECUTE = "agent.kai.execute"
```

Or remove the orphaned key from `_STATIC_SCOPE_META` in `scope_helpers.py` line 179.

Screenshots : 

<img width="1386" height="918" alt="Image" src="https://github.com/user-attachments/assets/0e61f17a-41f0-4739-a974-1f12acf3386b" />
<img width="1913" height="827" alt="Image" src="https://github.com/user-attachments/assets/2f590afc-d40f-4dd0-9139-1219ef4ee810" />
**Labels:** `bug` `auth` `critical`

PR #469, opened 2026-04-21
Fix: valid Kai tokens rejected as "Malformed"

bharath110520:
#### 🔍 Root Cause

* `agent.kai.execute` was referenced in `scope_helpers.py` but missing in `ConsentScope` enum
* This caused a `ValueError` during enum conversion
* A broad `except Exception` in `validate_token()` masked the real issue

#### ✅ Fix Implemented

* Added missing enum:
  AGENT_KAI_EXECUTE = "agent.kai.execute"
* Fixed scope resolution logic
* Prevented incorrect fallback behavior

#### 🧪 Validation

* Reproduced issue using test harness
* Added test: `test_bug408_dryrun.py`
* Verified token validation works correctly

#### 🔐 Impact

* Fixes valid token rejection
* Improves reliability of authentication layer

#### 📁 Files Modified

* constants.py
* scope_helpers.py
* token.py
* test_bug408_dryrun.py

#### ⚠️ Note

* Full test suite not executed due to missing dependencies
* Dry-run test confirms fix for reported issue

### DIFF
--- a/consent-protocol/hushh_mcp/consent/scope_helpers.py
+++ b/consent-protocol/hushh_mcp/consent/scope_helpers.py
@@ -49,16 +49,21 @@ def resolve_scope_to_enum(scope: str) -> ConsentScope:
     if scope == "pkm.write":
         return ConsentScope.PKM_WRITE
 
-    # Agent permissions
+    # Agent permissions — direct enum lookup; never silently map unknown scopes
     if scope.startswith("agent."):
-        return ConsentScope.AGENT_EXECUTE
+        try:
+            return ConsentScope(scope)
+        except ValueError:
+            raise ValueError(
+                f"Unknown agent scope '{scope}'. "
+                f"Valid agent scopes: {[s.value for s in ConsentScope.agent_scopes()]}"
+            )
 
-    # Custom/temporary scopes
-    if scope.startswith("custom."):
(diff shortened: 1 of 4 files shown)


bharath110520:
This PR fixes the issue where valid "agent.kai.execute" tokens were incorrectly rejected as malformed due to a missing enum and improper scope handling.

All CI checks are passing. Please review.

kushaltrivedi5:
## Closed: Superseded By #409

### Decision
`#409` landed the same Bug `#408` contract on `main` at `276ddb08`, preserving `agent.kai.execute` as the canonical Kai execution scope.

### What We Kept
This PR correctly isolated the root cause: valid Kai tokens were being rejected as malformed because the scope contract was incomplete. That signal was kept and moved into the final landed path.

### Proof
`#409` Queue Validation passed, `#409` Main Post-Merge Smoke passed, and Consent Protocol upstream CI passed at `a09bfdcf`.

### Outcome
The bug is resolved through the canonical scope contract rather than a parallel fix path. Closing this keeps the issue history accurate while avoiding two competing Kai execute-scope decisions.

bharath110520:
Thanks for the update!

Glad the root cause analysis aligned with the final fix. 
I learned a lot digging into the consent scope + token flow.

Will pick up another issue 👍

PR #474, opened 2026-04-22
fix: replace db.ts stubs with real PKM backend calls

Somsubhra-Nandi:
# fix: replace db.ts stubs with real PKM backend calls

## Description

`lib/db.ts` contained three stub functions that were silently broken in
production:

- `storeUserData()` logged `[STUB]` to the console, waited 500ms, and
  returned `true` — meaning encrypted user preferences were never actually
  persisted to the backend. Any user saving vault preferences believed the
  operation succeeded while data was silently discarded.

- `getAllFoodData()` and `getAllProfessionalData()` permanently returned
  `null` and had no callers. The backend explicitly documents their removal:
  `db_proxy.py` line 224: *"NOTE: /food/get and /professional/get removed;
  domain data is via PKM."*

**Changes:**

1. `storeUserData()` — replaced the fake delay with a real
   `fetch("/api/pkm/store-domain", ...)` call. Sends the encrypted blob
   (`ciphertext`, `iv`, `tag`) to the PKM backend which stores it opaquely
   (BYOK — backend cannot decrypt). Includes a 30s `AbortController` timeout
   and returns `false` on failure instead of silently returning `true`.

2. `getAllFoodData()` and `getAllProfessionalData()` — removed. Zero callers,
   permanently returning `null`, backend endpoints deleted. Dead code.

---

## 📌 Impact Map (Required)

- Routes touched:
  - [x] Listed below:
    - `POST /api/pkm/store-domain` — now actually called by `storeUserData()`
      (was previously bypassed by the stub)
  - [x] Listed below:
    - `getAllFoodData()` and `getAllProfessionalData()` removed from `lib/db.ts`
    - No callers exist — confirmed by grep across entire `hushh-webapp`
- Cache keys touched:
  - [x] None
  - [x] None — `summary: {}` passed intentionally; store-preferences stores
    encrypted preference blobs, not queryable PKM domain data
- Mobile parity impacts:
  - [x] None — `storeUserData()` uses relative `/api/pkm/store-domain` path
    which routes correctly on both web and Capacitor via the existing
    Next.js proxy
- Docs updated (exact files):
  - [x] None — removal of dead functions is self-documenting via commit message
  - [x] `cd hushh-webapp && npm run typecheck` — 0 errors

---

- [x] **Web:** `storeUserData()` calls `/api/pkm/store-domain` which is an
  existing Next.js proxy route — no change needed
- [x] **iOS:** N/A — `storeUserData()` is called from a Next.js API route
  (`store-preferences/route.ts`), not from a Capacitor plugin. No native
  change required.
- [x] **Android:** N/A — same as iOS above

---

## 🧪 Testing

- [x] `npm run typecheck` — 0 errors
- [x] Confirmed zero callers of `getAllFoodData()` and
  `getAllProfessionalData()` via grep before removal
- [x] `storeUserData()` call site in `store-preferences/route.ts` unchanged —
  same function signature, same parameters, drop-in replacement

---

## 🛡️ Privacy & Consent

- [x] Does this change access user data? **Yes — encrypted write only**
- [x] Consent is enforced upstream in `store-preferences/route.ts` which
  validates the `consentToken` with the Python backend before calling
  `storeUserData()`. This function is only ever reached after consent
  is confirmed — no change to the consent gate.
- Raw plaintext is never handled — only `ciphertext`, `iv`, and `tag`
  are passed and forwarded. BYOK guarantee preserved.

---

## 📜 Licensing

- [x] No new dependencies introduced — uses only browser-native `fetch`
  and `AbortController`

### DIFF
--- a/hushh-webapp/lib/db.ts
+++ b/hushh-webapp/lib/db.ts
@@ -1,11 +1,37 @@
 /**
  * Database Utility Library
  *
- * Handles storage of encrypted user data.
- * Currently a stub implementation to allow build to pass.
- * TODO: Implement backend API call to persist data to Cloud SQL via db_proxy.
+ * Handles storage of encrypted user data via the PKM backend.
+ *
+ * Architecture note:
+ * - storeUserData() proxies to POST /api/pkm/store-domain on the backend.
+ *   The backend stores the ciphertext blob without being able to decrypt it
+ *   (BYOK guarantee — the vault key never leaves the client).
+ *
+ * Removed:
+ * - getAllFoodData() and getAllProfessionalData() were stubs that returned null.
+ *   These domain-specific getters were removed from the backend
+ *   (see db_proxy.py: "NOTE: /food/get and /professional/get removed;
+ *   domain data is via PKM"). Domain data is now read via the PKM
+ *   /api/pkm/domains/{userId} endpoint instead.
  */
 
+const PKM_STORE_TIMEOUT_MS = 30_000;
+
+/**
+ * Persist an encrypted user data field to the PKM backend.
+ *
+ * The `value`, `iv`, and `tag` parameters are the AES-GCM ciphertext
+ * components produced by the client-side vault encryption layer. The
+ * backend stores them opaquely — it cannot decrypt the payload.
+ *
+ * @param userId  - Firebase UID of the data owner
+ * @param key     - PKM domain path (e.g. "financial", "health")
+ * @param value   - Base64-encoded AES-GCM ciphertext
+ * @param iv      - Base64-encoded initialisation vector (12 bytes)
+ * @param tag     - Base64-encoded authentication tag (16 bytes)
+ * @returns true on success, false on failure (caller decides how to handle)
+ */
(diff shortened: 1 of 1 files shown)


kushaltrivedi5:
## Closed: Superseded By Maintainer PKM Patch

### Decision
Thanks for the direction here. We kept the core idea, but the submitted branch could not be patched directly from the maintainer side, so I landed the merge-safe version through [#555](https://github.com/hushh-labs/hushh-research/pull/555).

### What We Kept
The fake `storeUserData()` stub is replaced with a real PKM backend write path through `/api/pkm/store-domain`. That is the useful product/runtime change from this PR.

### Decision Basis
The landed maintainer patch also forwards the validated VAULT_OWNER token into the PKM proxy and removes user id, field-name, and ciphertext logging from the activated vault preference path. Those fixes were required before this could safely become a real write path.

### Outcome
[#555](https://github.com/hushh-labs/hushh-research/pull/555) is now the landed contract for this change. This PR is closed so we do not keep two branches open for the same PKM store contract.

PR #428, opened 2026-04-18
fix(stream): upgrade _require_vault_owner_token to DB-backed revocation check + fix token extraction

9SERG4NT:
Fixes #429

#### What was found

`_require_vault_owner_token()` in `api/routes/kai/stream.py` had three defects not present in any other Kai auth guard:

1. **Unsafe token extraction** — `authorization.replace("Bearer ", "")` replaces all occurrences of the substring `"Bearer "` in the header value, including any that appear inside the token body itself. The canonical pattern (`removeprefix("Bearer ").strip()`) strips only the leading prefix.

2. **Offline-only revocation check** — The function called `validate_token()` which only verifies the JWT signature. Every other Kai auth guard (`analyze.py`, `portfolio.py`, `losers.py`, `chat.py`) calls `validate_token_with_db()`, which additionally queries `consent_audit` to reject revoked tokens. A revoked VAULT_OWNER consent token would still pass the SSE stream endpoint.

3. **Missing `WWW-Authenticate` header on 401 responses** — RFC 7235 §4.1 requires this header on every 401. The canonical middleware always sets it; `stream.py` did not.

#### What was changed

- `api/routes/kai/stream.py`:
  - Added `validate_token_with_db` to the import
  - Replaced `replace("Bearer ", "")` with `removeprefix("Bearer ").strip()`
  - Replaced `validate_token(...)` with `await validate_token_with_db(...)` for DB-backed revocation
  - Added `headers={"WWW-Authenticate": "Bearer"}` to both 401 raise sites
  - Fixed two stale inline comments that still referenced `validate_token()`

- `tests/test_kai_stream_auth_guard.py` _(new)_:
  - Unit tests for missing token → 401, invalid token → 401, cross-user token → 403
  - Regression guard asserting the `replace()` bug returns 401/403 not 500

- `tests/quality/test_kai_route_auth_compliance.py`:
  - Tightened the static quality gate for stream.py routes to require `validate_token_with_db`

#### How it was validated

```bash
python -m pytest tests/quality/test_kai_route_auth_compliance.py \
                 tests/test_kai_stream_contract.py \
                 tests/test_kai_stream_context_gate.py \
                 tests/test_kai_llm_provider_contract.py -v
# 16 passed
```

All pre-existing tests that touch `stream.py` continue to pass. No new failures introduced.

#### Security note

[...]

PR #514, opened 2026-04-24
fix(security): use DB-backed token validation in Kai stream auth (#429)

### LINKED ISSUE #429: Security: `_require_vault_owner_token` in Kai stream endpoint bypasses token revocation
### Bug Description
The `_require_vault_owner_token` authentication guard inside `api/routes/kai/stream.py` uses the offline `validate_token()` function instead of the canonical `validate_token_with_db()`. This means that if a Vault Owner token is explicitly revoked in the database, it will still successfully authenticate against the Kai SSE stream endpoint.

Additionally:
- Token extraction uses an unsafe `.replace("Bearer ", "")` which can corrupt the token if it contains the substring.
- 401 Unauthorized responses do not include the `WWW-Authenticate: Bearer` header, violating RFC-7235.

### Proposed Fix
I have raised a PR for the Hushh Open Source Hiring Challenge to:
- Upgrade the stream auth guard to `await validate_token_with_db(...)`.
- Sanitize token extraction using `removeprefix("Bearer ").strip()`.
- Add missing `WWW-Authenticate` headers to exceptions.

vandit98:
## Summary

`_require_vault_owner_token` in [consent-protocol/api/routes/kai/stream.py](consent-protocol/api/routes/kai/stream.py#L54) called `validate_token`, which only checks the per-process in-memory revoked set. A consent token revoked on another Cloud Run instance (or revoked via `/api/consent/revoke` right before a load-balancer failover to a different replica) continued to authenticate Kai `/api/kai/analyze/stream` requests on the replica that had not yet seen the revocation locally.

This leaves a cross-instance **consent revocation bypass** on the vault-owner surface that fronts Kai analysis and the PKM vault — a direct violation of the consent-first guarantee.

## Root cause

- `hushh_mcp/consent/token.py` exposes two validators:
  - `validate_token` — signature, expiry, and *in-process* `_revoked_tokens` set only.
  - `validate_token_with_db` — the above plus `ConsentDBService.is_token_active` (authoritative, shared across replicas).
- Every other sensitive Kai route (chat, portfolio, losers stream, etc.) already uses the DB-backed path. The analyze/stream guard was the outlier.

## Fix

Switch `_require_vault_owner_token` to `validate_token_with_db(consent_token, ConsentScope.VAULT_OWNER)`. The function was already `async` and every caller already `await`s it, so no signature change propagates.

`validate_token_with_db` has a safe fallback: if the DB lookup itself raises (transient outage), it logs and returns the in-memory result, so legitimate non-revoked tokens are not 401'd during a DB blip.

## Validation

```
$ cd consent-protocol
$ uv run pytest tests/test_kai_stream_revocation_bypass.py -v
============================== 4 passed in 25.47s ==============================
```

New regression tests ([consent-protocol/tests/test_kai_stream_revocation_bypass.py](consent-protocol/tests/test_kai_stream_revocation_bypass.py)):

1. `test_revoked_token_is_rejected_on_analyze_stream_get` — monkeypatches `ConsentDBService.is_token_active` to return `False`; GET returns `401 Invalid token: Token has been revoked`.
2. `test_revoked_token_is_rejected_on_analyze_stream_post` — same check, POST body variant.
3. `test_active_token_still_passes_auth_gate` — happy path; returns non-401/403.
4. `test_db_outage_does_not_downgrade_to_unauthenticated` — raises `RuntimeError` from `is_token_active`; legitimate token still passes (fallback works). Prevents turning a transient DB outage into a global Kai logout.

No regression in adjacent suites:

```
$ uv run pytest tests/test_kai_auth_matrix.py tests/test_kai_stream_contract.py tests/test_kai_stream_context_gate.py -v
============================== 56 passed in 5.59s ==============================
```

## Risk

- **Latency:** adds one indexed DB read per authed SSE open. Negligible vs SSE runtime.
- **Availability:** `validate_token_with_db` degrades to in-memory on DB errors (same behavior every other Kai route already exhibits).
- **Compatibility:** no request/response contract change, no public-surface change.

## Follow-ups (out of scope)

- Consider raising the log level for DB-fallback to `WARNING` plus adding a counter metric so ops can see when the fallback window is load-bearing.
- Audit remaining `validate_token` callsites inside `consent-protocol/` for the same cross-instance revocation gap.

---

Closes #429.

### DIFF
--- a/consent-protocol/api/routes/kai/stream.py
+++ b/consent-protocol/api/routes/kai/stream.py
@@ -30,7 +30,7 @@ from hushh_mcp.agents.kai.debate_engine import DebateEngine
 from hushh_mcp.agents.kai.fundamental_agent import FundamentalAgent, FundamentalInsight
 from hushh_mcp.agents.kai.sentiment_agent import SentimentAgent, SentimentInsight
 from hushh_mcp.agents.kai.valuation_agent import ValuationAgent, ValuationInsight
-from hushh_mcp.consent.token import validate_token
+from hushh_mcp.consent.token import validate_token, validate_token_with_db
 from hushh_mcp.constants import ConsentScope
 from hushh_mcp.operons.kai.llm import (
     get_gemini_unavailable_reason,
@@ -63,7 +63,10 @@ async def _require_vault_owner_token(
         )
 
     consent_token = authorization.replace("Bearer ", "")
-    valid, reason, payload = validate_token(consent_token, ConsentScope.VAULT_OWNER)
+    # DB-backed check catches cross-instance revocations (issue #429).
+    valid, reason, payload = await validate_token_with_db(
+        consent_token, ConsentScope.VAULT_OWNER
+    )
 
     if not valid or not payload:
         raise HTTPException(status_code=401, detail=f"Invalid token: {reason}")
@@ -2409,7 +2412,7 @@ async def analyze_stream(
     - decision: Final decision card
     - error: Fatal error
     """
-    # Auth path includes validate_token() inside _require_vault_owner_token().
+    # Auth path includes validate_token_with_db() inside _require_vault_owner_token().
     consent_token = await _require_vault_owner_token(user_id=user_id, authorization=authorization)
 
     # Log operation for audit trail (shows what vault.owner token was used for)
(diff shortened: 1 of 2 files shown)


vandit98:
Closing: this PR is a strict subset of #428 which already covers the same stream.py fix plus `removeprefix` hardening, WWW-Authenticate headers, and a tightened quality-gate test. Reopening capacity to pursue a non-duplicate contribution.

PR #515, opened 2026-04-24
fix(security): derive rate-limit key from verified bearer token, not client header

vandit98:
## Summary

The slowapi rate-limit key function in [consent-protocol/api/middlewares/rate_limit.py](consent-protocol/api/middlewares/rate_limit.py#L20) trusted a **client-controlled** `X-User-ID` header as the per-user bucket. No upstream validation binds that header to the authenticated identity, so the header is effectively free-form client input. This enables two concrete attacks:

1. **Self quota bypass.** A single abuser rotates `X-User-ID: <random>` per request. Every unique value allocates a fresh bucket, so `@limiter.limit("10/minute")` becomes `unlimited/minute`.
2. **Cross-user denial-of-service.** An attacker sends requests with `X-User-ID: <victim_user_id>` — entirely unauthenticated — and burns the victim's `10/minute` bucket inside six seconds, locking the victim out of the consent flow.

This is the classic "trusted client header" anti-pattern. It is particularly sharp here because the limiter fronts consent-grant and agent-chat endpoints — the core of the consent-first guarantee.

### Repro (current main, HEAD=`eec154c0`)

```python
from unittest.mock import MagicMock
from api.middlewares.rate_limit import get_rate_limit_key

class R:
    def __init__(self, hdrs, ip):
        self.headers = hdrs
        self.client = MagicMock(); self.client.host = ip
        self.url = MagicMock(); self.url.path = "/api/consent/request"

# Attacker from one IP rotates the header — gets 25 distinct buckets:
print(len({get_rate_limit_key(R({"X-User-ID": f"s{i}"}, "203.0.113.7")) for i in range(25)}))
# => 25

# Attacker spoofs the victim's bucket to DoS them:
print(get_rate_limit_key(R({"X-User-ID": "victim"}, "203.0.113.7")))
# => "user:victim"  (lands in the victim's bucket without auth)
```

## Root cause

[consent-protocol/api/middlewares/rate_limit.py:26](consent-protocol/api/middlewares/rate_limit.py#L26):

```python
user_id = request.headers.get("X-User-ID")
if user_id:
    return f"user:{user_id}"
```

`request.headers.get("X-User-ID")` is raw client input. There is no corresponding dependency or middleware that verifies the header matches the authenticated principal; no callsite in the codebase populates it from a trusted source.

The existing unit test in [tests/quality/test_rate_limiting.py](consent-protocol/tests/quality/test_rate_limiting.py) codified the broken behavior (`test_authenticated_user_keyed_by_user_id` explicitly constructs a request with only the header and expects `"user:user_123"`), so the quality gate was pinning the bug rather than the contract.

## Fix

Key the limiter on the **signature-verified** consent bearer token, not a header:

```python
authorization = request.headers.get("Authorization") or request.headers.get("authorization")
if authorization and authorization.startswith("Bearer "):
    consent_token = authorization.removeprefix("Bearer ").strip()
    if consent_token:
        valid, _reason, payload = validate_token(consent_token)
        if valid and payload and payload.user_id:
            return f"user:{payload.user_id}"

return get_remote_address(request)
```

`validate_token` is the pure-HMAC, no-DB, synchronous validator already used as the offline half of `validate_token_with_db`. It runs in microseconds — safe to call from a per-request key function. Unauthenticated or unverifiable traffic now collapses to the remote IP, so abusers share a bucket with themselves instead of each request minting a new one.

Properties regained:

- Rotating a header does nothing. Bucket = IP.
- Spoofing a victim's `X-User-ID` does nothing. Bucket = attacker IP.
- Legitimate users with a valid bearer token still get a per-user bucket (same granularity as before, but actually bound to identity).
- Tampering a single byte on the token drops the request to the IP bucket.

## Validation

```
$ cd consent-protocol
$ uv run pytest tests/quality/test_rate_limiting.py -v
============================== 16 passed in 0.13s ==============================
```

8 of those 16 are new regression tests in [tests/quality/test_rate_limiting.py](consent-protocol/tests/quality/test_rate_limiting.py):

- `test_valid_bearer_token_keyed_by_token_user_id` — authed happy path
- `test_unauthenticated_falls_back_to_ip`
- `test_different_authed_users_different_keys`
- `test_x_user_id_header_is_ignored_without_bearer_token`
- `test_rotating_x_user_id_does_not_create_new_buckets` — **quota-bypass regression**
- `test_cannot_dos_another_user_via_x_user_id_header` — **cross-user DoS regression**
- `test_malformed_bearer_token_falls_back_to_ip`
- `test_bearer_token_with_suffix_garbage_is_not_accepted`

Pre-existing tests (`TestRateLimitConstants`, `TestRateLimitEnforcement`) kept intact and continue to pass.

Full `tests/quality/` run: 54 pass, 1 fail — the one failure (`test_no_direct_db_import_in_routes`) is pre-existing on `main` and unrelated to this patch (flags `db_proxy.py`, `kai/gmail.py`, `consent.py`).

## Risk

- **Latency:** one HMAC-SHA256 verify + base64 decode per authed request in the key function. Sub-millisecond; negligible vs the request path.
- **Compatibility:** any internal caller that previously relied on the `X-User-ID` header to scope rate limits is silently falling back to IP. Nothing in-tree sets the header from a trusted source, so no real caller regresses — only the broken unit test, which is updated here.
- **Availability:** IP fallback preserves the `100/minute` global bucket for anonymous traffic. No route becomes unreachable.

## Notes for reviewers

- This is framed as a hardening patch rather than a vulnerability disclosure, per `SECURITY.md`'s private-channel policy for unresolved issues. If the Hushh security team would prefer this tracked as an advisory, happy to open a private one in parallel.
- The `validate_token` import is lazy-safe — the `hushh_mcp.consent.token` module has no side-effect-heavy import chain and is already imported at startup by `api.middleware`, `api.routes.kai.stream`, and the chat/analyze routes.

## Follow-ups (out of scope)

- Apply `@limiter.limit(...)` to the remaining state-changing endpoints that currently have no rate limit (consent approve/deny, Kai chat, streaming analyze). Right now the fix hardens the key function but most routes do not attach the decorator, so the limiter is largely a no-op in production. That should be a separate PR with ops-approved quota numbers.
- Add a structured-log field when the key function falls back from bearer-token to IP, to spot clients that stop carrying the token.


kushaltrivedi5 (approved):
Approved and merged through the main queue on `7d3f7fe4`.

This fixes the right trust boundary. Rate-limit buckets now come from verified bearer-token identity instead of the client-controlled `X-User-ID` header, closing quota bypass and cross-user bucket burning without broadening auth behavior.

vandit98:
hey @kushaltrivedi5 
Thanks for the review.
I am not seeing any option to merge it. Can you do it.

PR #522, opened 2026-04-25
fix(security): accept Authorization Bearer for developer API + remote MCP (close token-in-URL leak vector)

vandit98:
## Summary

The developer API (`/api/v1/*`) and the remote MCP transport (`/mcp/`) currently authenticate **only** by `?token=<developer-token>` query parameter and actively reject `Authorization: Bearer` with a `400 DEVELOPER_TOKEN_QUERY_REQUIRED`. This forces every external developer to put a long-lived API key into the URL.

Tokens carried in URLs leak through every layer that touches a URL:

- **Referer headers** when the page links out anywhere external
- **Server access logs** (Cloud Run, nginx, ALB, etc. — full URL by default)
- **Browser history** and shared screenshots
- **CDN, WAF, and proxy logs**
- **Browser extensions** that read URLs

This is CWE-598 (Use of GET Request Method With Sensitive Query Strings) and OWASP A02:2021 (Cryptographic Failures). The blast radius is the full developer API surface — `user-scopes/{user_id}`, `consent-status`, `request-consent`, and `scoped-export` (which returns the encrypted vault export tied to the consent token). A single leaked URL gives an attacker the same access as the legitimate app.

## Repro on current main

```
GET /api/v1/user-scopes/user_123
Authorization: Bearer hdk_demo

HTTP/1.1 400 Bad Request
{"detail":{"error_code":"DEVELOPER_TOKEN_QUERY_REQUIRED",
 "message":"Use ?token=<developer-token> instead of Authorization header for developer access."}}
```

The application is _refusing_ the safer path and forcing the insecure one.

## Fix

Accept either transport, prefer the header. Bearer is the canonical RFC 6750 mechanism and survives every URL-leak vector above.

```python
def _resolve_developer_token(*, token, authorization):
    bearer_token = _resolve_bearer_token(authorization)
    if bearer_token:
        return bearer_token, "bearer"
    query_token = _resolve_query_token(token)
    if query_token:
        return query_token, "query"
    return "", ""
```

Applied to both [consent-protocol/api/developer_auth.py](consent-protocol/api/developer_auth.py) (`authenticate_developer_principal`, `try_authenticate_developer_principal`) and the ASGI guard in [consent-protocol/mcp_remote.py](consent-protocol/mcp_remote.py).

When a token still arrives via the query parameter, the auth layer emits a structured `WARNING` log (`event_type=developer_token_query_param_use`) so ops can observe migration progress and pinpoint clients still on the legacy transport before the query-fallback is ultimately removed:

```
Developer token received via query parameter; prefer Authorization: Bearer header
to avoid URL-leak vectors (Referer, access logs, browser history).
```

The 401 message is updated to reflect both options:

```
Developer token is required. Pass it as 'Authorization: Bearer <token>' (preferred)
or '?token=<token>' (legacy).
```

## Validation

```
$ cd consent-protocol
$ uv run pytest tests/test_developer_api_routes.py -v
============================== 24 passed in 0.37s ==============================
```

New tests in [tests/test_developer_api_routes.py](consent-protocol/tests/test_developer_api_routes.py):

- `test_user_scopes_accepts_authorization_bearer_header` — happy path via header (replaces the previous test that pinned the broken `400 DEVELOPER_TOKEN_QUERY_REQUIRED` response).
- `test_user_scopes_invalid_authorization_bearer_returns_403` — invalid header still rejected.
- `test_user_scopes_authorization_bearer_takes_precedence_over_query` — when both are sent, header wins (so a deploy that flips a client to header auth doesn't accidentally still leak the URL token).
- `test_user_scopes_query_token_logs_url_leak_warning` — observable migration signal.
- `test_tool_catalog_accepts_authorization_bearer_header` — second route covered for parity.

The two existing `test_*_rejects_authorization_header` tests that codified the broken behavior are replaced rather than deleted.

## Risk

- **Backward compatibility:** all existing `?token=` callers continue to work unchanged. Any deployed developer app, MCP client, or shell snippet keeps functioning. The only behavior change visible to legacy callers is a single structured log line on each request.
- **Bearer precedence:** if a (theoretical) caller is sending both a header and a query token, the header now wins. There is no in-tree caller doing this; an external caller doing it would typically be migrating off the query path, which is exactly the desired direction.
- **Scope:** patch is local to two files plus tests. No DB schema change, no migration, no ops change.

## Notes for reviewers

- This is framed as hardening rather than a vulnerability disclosure, per `SECURITY.md`'s private-channel policy. If the security team would prefer this tracked as a parallel private advisory, happy to file one.
- The remote MCP path keeps the `?token=` fallback because some MCP HTTP/SSE clients cannot attach custom headers; the warning log lets operators flag those clients for migration without breaking them.

## Follow-ups (out of scope)

- After a deprecation window driven by the new warning log, drop the `?token=` query path entirely (single line: remove the `query_token` fallback in `_resolve_developer_token`).
- Document the header-first contract in `docs/reference/architecture/route-contracts.md` so future developer-API routes inherit the safer default.

### DIFF
--- a/consent-protocol/api/developer_auth.py
+++ b/consent-protocol/api/developer_auth.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import os
 
 from fastapi import HTTPException, Request, status
@@ -8,6 +9,8 @@ from hushh_mcp.services.developer_registry_service import (
     DeveloperPrincipal,
     DeveloperRegistryService,
 )
+
+logger = logging.getLogger(__name__)
 
 
 def _env_truthy(name: str, fallback: str = "false") -> bool:
@@ -73,12 +76,48 @@ def _resolve_query_token(token: str | None = None) -> str:
     return str(token or "").strip()
 
 
-def _developer_token_query_required_error() -> HTTPException:
-    return HTTPException(
-        status_code=status.HTTP_400_BAD_REQUEST,
-        detail={
-            "error_code": "DEVELOPER_TOKEN_QUERY_REQUIRED",
-            "message": "Use ?token=<developer-token> instead of Authorization header for developer access.",
+def _resolve_developer_token(
+    *,
+    token: str | None,
+    authorization: str | None,
+) -> tuple[str, str]:
+    """
+    Resolve the developer token from either an Authorization: Bearer header or
+    a ?token= query parameter, in that order of preference.
+
+    Returns a (raw_token, source) tuple where source is "bearer", "query", or
+    "" when neither carries a value. Bearer is preferred because tokens passed
+    in URLs leak via Referer headers, server access logs, browser history, and
+    proxy/CDN logs (CWE-598). Query support is retained so existing developer
+    API and remote MCP clients keep working.
+    """
+    bearer_token = _resolve_bearer_token(authorization)
+    if bearer_token:
+        return bearer_token, "bearer"
+
+    query_token = _resolve_query_token(token)
+    if query_token:
+        return query_token, "query"
+
+    return "", ""
+
+
+def _warn_query_token_leak_risk(*, request: Request | None) -> None:
+    path = ""
+    client_ip = ""
+    if request is not None:
+        try:
+            path = request.url.path
+        except Exception:
+            path = ""
+        client_ip = request.client.host if request.client else ""
+    logger.warning(
+        "Developer token received via query parameter; prefer Authorization: Bearer header "
+        "to avoid URL-leak vectors (Referer, access logs, browser history).",
+        extra={
+            "event_type": "developer_token_query_param_use",
+            "path": path,
+            "client_ip": client_ip,
         },
     )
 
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (approved):
## Approved: Bearer Token Path For Developer API And Remote MCP

### What Landed
- Developer API and remote MCP access now accept `Authorization: Bearer` as the preferred token path.
- Query-token support remains for compatibility, but the token-in-URL path is no longer the forced default.
- Route tests cover bearer success, invalid bearer rejection, bearer precedence over query tokens, query-token warning behavior, and tool catalog access.

### Why This Is Safe
This tightens the existing developer-auth surface without adding a parallel auth model. The implementation avoids logging token material and keeps the compatibility path explicit.

### Verification
- Reviewed head: `f5c94946`.
- Current gate passed: CI Status Gate, Protocol, Integration, Governance, DCO, Secret Scan, Freshness, and Upstream Sync.
- Lean/core risk is low because the PR closes a leak vector inside an existing contract.

kushaltrivedi5:
## Merged: Bearer Token Path For Developer API And Remote MCP

### What Landed
Developer API and remote MCP access now prefer bearer-token transport, while query-token compatibility remains explicit where it is still needed.

### Proof
Queue Validation passed: https://github.com/hushh-labs/hushh-research/actions/runs/24949144780

Main Post-Merge Smoke passed: https://github.com/hushh-labs/hushh-research/actions/runs/24949211347

Upstream consent-protocol CI passed: https://github.com/hushh-labs/consent-protocol/actions/runs/24949278120

### Outcome
This moves developer and remote MCP authentication away from URL-carried token material without silently breaking existing compatibility. The safer bearer path is now the preferred contract, and the remaining query-token path is explicit rather than accidental.

PR #524, opened 2026-04-25
feat: add commercial flag to consent tokens for monetized agent usage

### LINKED ISSUE #30: upgrade consent-protocol for having commercial and non-commercial handshakes
Every agent will have can be set up to have commerical/non-commercial usage for monetization

hitakshiA:
## Summary

- Adds an optional `commercial` flag to `HushhConsentToken` so the consent protocol can distinguish monetized agent usage from non-commercial usage
- Flag is part of the signed token payload, so it cannot be tampered with after issuance
- Backward compatible: every previously issued token continues to validate as non-commercial, and all existing tests pass unchanged
- 16 new tests cover defaults, round-trip, tamper resistance, the `require_commercial` gate, and malformed-shape rejection

## Problem

Issue #30 asks: "Every agent will have can be set up to have commercial / non-commercial usage for monetization." Today the consent protocol has no way to record whether a granted consent authorizes monetized agent use. This is the protocol-layer hook for the future monetization surface (e.g. paid agent operations need an explicit commercial grant; free / personal-use agents do not).

## Approach

**Backward-compat first.** Token signed payload was `user|agent|scope|issued|expires`. Changing that shape would invalidate every previously issued token. Instead:

- Non-commercial tokens keep the original 5-field shape exactly. `issue_token(...)` with no `commercial` argument produces the same bytes as before.
- Commercial tokens append a literal `|commercial` sixth field to the signed bytes. The parser distinguishes them by part count and the explicit marker.

**Tamper resistance.** The flag lives inside the HMAC-signed bytes:
- Appending `|commercial` to a non-commercial token without re-signing fails signature validation
- Stripping `|commercial` off a commercial token without re-signing fails signature validation
- Both paths are tested

**Validation gate.** `validate_token(..., require_commercial=True|False|None)` enforces the boundary at validation time. `None` (default) accepts either. `True` rejects non-commercial with `Commercial consent required for this operation`. `False` rejects commercial with `Non-commercial consent required for this operation`.

## What did not change

- `_sign`, key derivation, expiration handling, scope matching, revocation
- Token prefix, base64 encoding
- All existing call sites (the new parameter is keyword-only with a `False` default)

## Test plan

- [x] 16 new tests in `test_token_commercial.py` covering defaults, commercial round-trip, two tamper-resistance paths, all four `require_commercial` gate cases, and three malformed-shape rejection paths
- [x] All 5 existing tests in `test_token.py` still pass unchanged (backward compat)
- [x] Hand-crafted legacy 5-field token validates as non-commercial (confirms the wire-compat boundary)
- [x] 36 cross-cutting tests pass (`test_byok_encryption`, `test_developer_api_routes`, `test_kai_route_auth_compliance`)
- [x] ruff check passes

Closes #30

### DIFF
--- a/consent-protocol/hushh_mcp/consent/token.py
+++ b/consent-protocol/hushh_mcp/consent/token.py
@@ -90,19 +90,15 @@ def _scope_str_to_enum(scope_str: str) -> ConsentScope:
     """
     Map a scope string to its ConsentScope enum equivalent.
     Dynamic scopes (attr.*) map to PKM_READ.
+    Unknown static scopes are rejected instead of silently escalating to PKM_READ.
     """
     try:
         return ConsentScope(scope_str)
     except ValueError:
-        if scope_str == "pkm.read":
-            return ConsentScope.PKM_READ
-        if scope_str == "pkm.write":
-            return ConsentScope.PKM_WRITE
         # Dynamic scope (e.g., attr.financial.*) - map to PKM_READ
         if scope_str.startswith("attr."):
             return ConsentScope.PKM_READ
-        # Unknown scope - default to PKM_READ
-        return ConsentScope.PKM_READ
+        raise
 
 
 # ========== Token Verifier ==========
@@ -159,9 +155,7 @@ def validate_token(
         scope_enum = _scope_str_to_enum(scope_str)
 
         if commercial:
-            raw = (
-                f"{user_id}|{agent_id}|{scope_str}|{issued_at_str}|{expires_at_str}|commercial"
-            )
+            raw = f"{user_id}|{agent_id}|{scope_str}|{issued_at_str}|{expires_at_str}|commercial"
         else:
             raw = f"{user_id}|{agent_id}|{scope_str}|{issued_at_str}|{expires_at_str}"
         expected_sig = _sign(raw)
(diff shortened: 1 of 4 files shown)


kushaltrivedi5 (approved):

kushaltrivedi5:
## Merged: Commercial Consent Token Attribute

### What Landed
Commercial intent is now carried as a signed consent-token attribute, so monetized agent access can distinguish ordinary consent from explicitly commercial consent without changing the base token shape for non-commercial grants.

### Why It Matters
This keeps the consent boundary honest: commercial access is not inferred from caller behavior or UI copy. Routes that need monetized access can require the signed commercial marker, and older tokens remain non-commercial by default.

### Maintainer Patch
I tightened the enforcement path before merge so `validate_token_with_db(..., require_commercial=True)` forwards the commercial requirement into signed-token validation. I also restored the stricter unknown-scope behavior from `main`, added DB-backed commercial validation coverage, and documented the commercial attribute in the IAM reference.

### Documentation Updated
Updated `docs/reference/iam/architecture.md` with the commercial consent attribute contract and the expectation that critical monetized routes use DB-backed validation.

### Outcome
Merged at `220e66b2` after PR Validation, Queue Validation, and Main Post-Merge Smoke all passed.

PR #530, opened 2026-04-26
perf: replace read-modify-write on pkm_index with atomic JSONB merge RPC

Somsubhra-Nandi:
## Description

Fixed a read-modify-write race condition in `update_domain_summary()` on the `pkm_index` table.

The old pattern was:
1. READ — `get_index_v2(user_id)`
2. MODIFY — merge domain summary in Python
3. WRITE — `upsert_index_v2(index)`

With concurrent writers (chat, portfolio import, attribute learner all writing simultaneously), the last writer silently wins and clobbers the others. No error is raised — data is just lost.

**Fix:** Added a Postgres function `merge_pkm_domain_summary` (migration `048`) that does the merge atomically using `ON CONFLICT DO UPDATE` with the JSONB `||` operator. `update_domain_summary()` now calls this via a single `_run_rpc()` — no read, no in-process merge, no race window.

## 📌 Impact Map (Required)

- Routes touched:
  - [x] None — internal service method, no route signatures changed
  - [x] None
- Cache keys touched:
  - [x] None
  - [x] `pkm_index.domain_summaries` writes are now atomic per domain key. Merge behaviour is identical — existing keys are preserved, incoming patch keys overwrite. `total_attributes` is no longer recalculated on every write (was derived from in-memory state anyway, not source of truth).
- Mobile parity impacts:
  - [x] None
- Docs updated:
  - [x] None

- [x] **Web**: N/A — backend-only change
- [x] **iOS**: N/A
- [x] **Android**: N/A

## 🧪 Testing

## 📸 Screenshots / Video

No visual demo applicable. Fix targets a DB-level race condition.
Migration 048 applied successfully, RPC merge_pkm_domain_summary created.

## 🛡️ Privacy & Consent

- [x] Does this change access user data? Yes — writes to `pkm_index` (same as before, no new access)
- [x] No new consent surface introduced; existing token validation unchanged.

## 📜 Licensing

- [x] No new dependencies added


kushaltrivedi5 (approved):

kushaltrivedi5:
## Merged: Atomic PKM Domain Summary Writes

### What Landed

This lands the `merge_pkm_domain_summary` RPC so PKM domain summary updates happen as one database upsert instead of a read-modify-write cycle in the service. That removes the concurrent-writer race where chat, imports, and learning paths could overwrite each other.

### Why It Matters

PKM summaries are written from multiple runtime paths, so the write contract needs to be atomic. Moving this into one RPC reduces lost-update risk and makes the database contract easier to verify before UAT.

### Maintainer Patch

I added a small maintainer patch before merge to make the migration contract stricter and keep existing metadata behavior intact: the new RPC is now listed in the UAT schema `required_functions`, the SQL path keeps `total_attributes` coherent, and `update_domain_summary()` has direct test coverage proving it uses the RPC instead of the old `pkm_index` upsert path.

### Outcome

PKM domain summaries now have a safer database-owned merge path, so chat, imports, and learning flows are less likely to overwrite each other under concurrent writes.

PR #532, opened 2026-04-27
feat(kai): add resilient financial chart states

Divya-rajendran009:
## What changed

Introduced resilient rendering states for the Kai financial charts experience to improve reliability when upstream metrics are unavailable, incomplete, or delayed.

## Why

The chart module previously assumed all quantitative and fundamental metrics were always present. In real-world scenarios, partial API responses, stale payloads, or temporary backend failures can lead to broken visualizations, invalid values, or degraded user experience.

This PR strengthens the component with defensive rendering patterns commonly used in production analytics systems.

## Improvements Included

* Added graceful fallback UI when financial metrics are unavailable
* Prevented broken chart rendering from incomplete datasets
* Hardened percentage displays against invalid / undefined values
* Added bounded radar chart scoring for stable visualization output
* Improved user trust with clear unavailable-data messaging
* Reduced risk of runtime UI failures caused by partial responses

## Technical Scope

Updated:

```txt
components/kai/kai-financial-charts.tsx
```

No business logic, API contracts, or backend services were changed.

## Impact

* Better production resilience
* Cleaner analytics UX during degraded data states
* Lower frontend crash risk
* Easier future extension for loading / retry / skeleton states

## Validation

* Verified local compilation
* Passed lint checks
* Confirmed clean branch diff with isolated component changes

### DIFF
--- a/hushh-webapp/components/kai/kai-financial-charts.tsx
+++ b/hushh-webapp/components/kai/kai-financial-charts.tsx
@@ -79,29 +79,13 @@ const radarConfig = {
 } satisfies ChartConfig;
 
 export default function KaiFinancialCharts({ quantMetrics, keyMetrics }: KaiFinancialChartsProps) {
-  const hasValidData = Boolean(
-  quantMetrics &&
-  keyMetrics &&
-  keyMetrics.fundamental &&
-  quantMetrics.revenue_trend_data?.length > 0
-);
-  if (!hasValidData) {
-    return (
-      <div className="rounded-2xl border border-border/40 bg-card/40 p-6 text-center backdrop-blur-sm">
-        <h3 className="text-lg font-semibold">Financial data unavailable</h3>
-        <p className="mt-2 text-sm text-muted-foreground">
-          We couldn’t load chart metrics right now. Please refresh or try again later.
-        </p>
-      </div>
-    );
-  }
-
+  // Prepare Radar Data
   const radarData = [
-    { subject: "Revenue Growth", value: Math.min(Math.max((quantMetrics.revenue_growth_yoy || 0) * 100 * 2, 0), 100), fullMark: 100 },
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Kai chart resilience must be live-linked

### Direction
The resilient chart fallback is the right kind of improvement: Kai should avoid blank or broken financial charts when market metrics are incomplete.

### Blocker
The current branch only changes `hushh-webapp/components/kai/kai-financial-charts.tsx`, and that component is not imported anywhere in the current web app. That means the PR does not improve the live Kai route yet, even though the title reads like an app-facing resilience fix.

### Path To Merge
Either wire the chart component into the canonical Kai analysis surface with the same-session route/user flow that should render it, or narrow the PR to an explicit component-level hardening change with a focused test that proves the fallback and missing-data behavior.

### Proof Needed
Please include the reachable caller path or test coverage that shows this state is actually exercised. If the component becomes live-linked, missing metrics should render as unavailable/degraded rather than being converted into real-looking `0%` values.

kushaltrivedi5 (approved):

kushaltrivedi5:
## Merged: Live Kai Decision Chart Resilience

### What Landed
The resilience behavior now lands on the live Kai decision-card chart path instead of the unused `KaiFinancialCharts` component. If quantitative metrics are present but not renderable as finite numeric chart data, the card falls back to the existing consensus visualization instead of leaving the chart slot empty.

### Why It Matters
This keeps the analysis result page stable when upstream metric payloads are partial or delayed, without adding a parallel chart surface or changing the Kai analysis API contract.

### Maintainer Patch
I moved the useful resilience idea from the unused component to `hushh-webapp/components/kai/views/decision-card.tsx`, rebased the branch on current `main`, and kept the patch frontend-only.

### Outcome
Kai's live recommendation card now handles missing or non-renderable quantitative metrics through the existing fallback path, while the unused chart component remains unchanged.

PR #554, opened 2026-04-29
fix(pkm): document cloud projection boundary

kushaltrivedi5:
## Summary

Follow-up to #530. The atomic `merge_pkm_domain_summary` RPC is the right cloud-side consistency fix, but the merged contract needed one explicit boundary: `pkm_index` is a repairable cloud discovery projection, not the source of user-memory truth.

This patch documents that boundary in the migration, backend service docstring, PKM reference, and cache-coherence reference.

## Why It Matters

Hussh is on-device-first with cloud sync. Local encrypted domain data, manifests, mutation events, and cache write-through must remain authoritative for user memory. Cloud summary projection can be atomic and useful, but local-only/offline/on-device flows must not fail just because the cloud projection is unavailable or lagging.

## Verification

- `cd consent-protocol && python3 -m pytest tests/services/test_pkm_service_store_domain_data.py -q`
- `cd hushh-webapp && npm run verify:cache`
- `./bin/hushh db verify-release-contract`
- `./bin/hushh docs verify`
- `bash scripts/ci/check-dco-signoff.sh origin/main HEAD`

### DIFF
--- a/consent-protocol/db/migrations/048_merge_pkm_domain_summary_rpc.sql
+++ b/consent-protocol/db/migrations/048_merge_pkm_domain_summary_rpc.sql
@@ -1,6 +1,9 @@
--- Migration 048: atomic JSONB merge RPC for pkm_index domain summaries
--- Replaces the read-modify-write in update_domain_summary() with a single
--- DB-level upsert, eliminating the race window on concurrent writes.
+-- Migration 048: atomic JSONB merge RPC for pkm_index domain summaries.
+--
+-- This function only updates the cloud discovery projection. It must not be
+-- treated as the source of user memory truth. Encrypted PKM blobs, manifests,
+-- mutation events, and local cache write-through remain authoritative; this
+-- projection is repairable from those sources when cloud sync lags or recovers.
 
 CREATE OR REPLACE FUNCTION merge_pkm_domain_summary(
     p_user_id      TEXT,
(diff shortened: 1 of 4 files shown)


kushaltrivedi5:
## Merged: PKM Cloud Projection Boundary

### What Landed
This follow-up documents that `pkm_index` and domain summaries are cloud discovery projections only. The authoritative memory path remains encrypted PKM domain data, manifests, mutation events, and local cache write-through.

### Why It Matters
Hussh is on-device first with cloud sync. Atomic cloud projection is useful for consistency, but local-only, offline, and future on-device AI flows must not depend on cloud metadata as the source of truth. This keeps #530 directionally useful without letting a DB projection quietly become the memory authority.

### Maintainer Patch
The maintainer patch updated the migration comment, backend service docstring, PKM reference, and cache-coherence reference. This was the smallest repair path because the runtime RPC behavior was already correct; the missing piece was the explicit authority boundary around projection versus memory.

### Documentation Updated
- [Personal Knowledge Model](https://github.com/hushh-labs/hushh-research/blob/main/consent-protocol/docs/reference/personal-knowledge-model.md)
- [Cache Coherence Reference](https://github.com/hushh-labs/hushh-research/blob/main/docs/reference/architecture/cache-coherence.md)

### Outcome
The cloud projection contract is now explicit: `pkm_index` can help discovery and sync, but user memory remains local-first, encrypted, and repairable from canonical PKM state. Future PKM/cloud-sync PRs should preserve that boundary instead of treating DB summary rows as authoritative memory.

PR #573, opened 2026-05-02
Refactor OrchestratorAgent to use strict error types and delegation extraction

Somsubhra-Nandi:
# Description

Completed the incomplete `handle_message` implementation in `OrchestratorAgent` (`hushh_mcp/agents/orchestrator/agent.py`). The method contained multi-paragraph TODO comments describing what _should_ happen but never actually executed delegation detection or structured error handling. This was live routing logic for all agent traffic.

Specific changes:

- **Typed exceptions** — introduced `OrchestratorError`, `OrchestratorConsentError`, `OrchestratorAgentError`, and `OrchestratorDelegationError` so API routes can return 401/503 instead of a flat 500 for every failure class.
- **`logger.exception()` replaces `logger.error(f"...")`** — the old call discarded the full traceback; `logger.exception()` captures it automatically.
- **Early consent guard** — an empty `consent_token` now raises `OrchestratorConsentError` before the ADK call instead of propagating a bare `PermissionError` from deep inside `HushhAgent`.
- **Delegation extraction implemented** — `_extract_delegation()` reads `response.tool_results` (the list ADK appends when tools fire) and surfaces the first dict with `"delegated": True` as the `delegation` key in the return value. This was the core unimplemented block.
- **Unknown target guard** — if a tool returns an unrecognised `target_agent` not in `_KNOWN_TARGETS`, the orchestrator raises `OrchestratorDelegationError` instead of silently forwarding it.
- **`_extract_text()` covers ADK `.parts` responses** — the original only tried `response.text`; ADK can also deliver content via `.parts` (list of `Part` objects).
- **Singleton typed** — `_orchestrator` is now `Optional[OrchestratorAgent]` for correct static analysis.

##  Impact Map

- Routes touched:
  - [x] Listed below:
    - `POST /api/agents/kai/chat` — indirectly; `KaiAgent.handle_message` follows the same pattern and benefits from the typed error contract established here. No route signature changed.
    - No new routes added or removed.

  - [x] Listed below:
    - `OrchestratorResult` return dict: `delegation` key is now always present (was previously absent on the success path). Value is `dict | None`. Callers that only read `response` are unaffected.
    - Three new exported exception types: `OrchestratorConsentError`, `OrchestratorAgentError`, `OrchestratorDelegationError`. Additive — no existing import breaks.

- Cache keys touched:
  - [x] None

  - [x] None — orchestrator is a routing layer; it reads no domain summaries and writes none.

- Mobile parity impacts:
  - [x] None — `OrchestratorAgent` is a pure backend class. It is not called from any Capacitor plugin. The agent chat surface exposed to iOS/Android goes through `POST /api/agents/kai/chat`, whose response shape is unchanged.

- Docs updated (exact files):
  - [x] None — no public API docs reference the internal orchestrator class. Inline docstrings updated in `agent.py` itself.

  > Backend-only change. Relevant verification is:
  > ```bash
  > cd consent-protocol
  > uv run pytest tests/agents/kai/test_orchestrator_exception_handling.py -v
  > uv run pytest tests/test_agent_consent_validation.py tests/test_a2a_delegation_scopes.py -v
  > ```

- [x] **Web**: N/A — `OrchestratorAgent` is a backend Python class, not a Next.js route.
- [x] **iOS**: N/A — no Capacitor plugin calls this class directly.
- [x] **Android**: N/A — same as iOS.

## 🧪 Testing

> UI/device testing is not applicable. The change is confined to `consent-protocol/hushh_mcp/agents/orchestrator/agent.py`. Existing tests in `tests/agents/kai/test_orchestrator_exception_handling.py` cover the exception-propagation contract. The delegation extraction path (`_extract_delegation`) should be covered by a new unit test — see follow-up task.

## 📸 Screenshots / Video

N/A — backend-only, no visual surface changed.

## 🛡️ Privacy & Consent

- [x] Does this change access user data? **No.**
  - The orchestrator forwards `consent_token` and `user_id` to `HushhAgent.run()` exactly as before. No new data is read from the vault or any user store. The early consent guard _reduces_ the surface by rejecting requests with empty tokens before any ADK call is made.

## 📜 Licensing

- [x] First-party changes remain Apache-2.0 compatible — no new dependencies introduced.
- [x] Third-party notice impact reviewed — `requirements.txt` and `pyproject.toml` are unchanged.

### DIFF
--- a/consent-protocol/hushh_mcp/agents/orchestrator/agent.py
+++ b/consent-protocol/hushh_mcp/agents/orchestrator/agent.py
@@ -7,7 +7,7 @@ Keeps the legacy orchestrator package path while One becomes the product owner.
 
 import logging
 import os
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from hushh_mcp.hushh_adk.core import HushhAgent
 from hushh_mcp.hushh_adk.manifest import ManifestLoader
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: canonical orchestrator boundary

### Direction
Typed orchestration errors are useful, but they need to attach to the current orchestrator runtime and delegation contract.

### Blocker
The PR claims a strict error/delegation refactor, but it only changes the orchestrator agent file without showing the caller contract, route behavior, or failure-mode proof that the runtime now consumes those typed errors correctly.

### Path To Merge
Rebase on current main and either wire the typed errors through the active orchestration path with tests, or reduce this to the smallest internal cleanup that preserves current behavior.

### Proof Needed
Provide focused protocol/runtime tests showing existing successful delegation still works and the new typed-error path is actually emitted and handled by the current caller.

PR #572, opened 2026-05-02
feat(consent): wire PermissionGate to real consent APIs

Divya-rajendran009:
## Description

This PR wires the existing PermissionGate system to the real consent API.

Previously, PermissionGate relied on mocked/local consent state. This update integrates it with `/api/consent/active`, ensuring UI access decisions are based on actual user consent.

## What changed

- Connected `useConsent` hook to `/api/consent/active`
- Added handling for loading, granted, denied, and error states
- Integrated PermissionGate into Kai dashboard via client wrapper
- Preserved server/client boundaries using dynamic import

## Architecture

UI → PermissionGate → useConsent → /api/consent/active

## Impact

- Enables real consent-driven UI gating
- Improves privacy enforcement at UI layer
- Prepares system for token-based backend validation

## Testing

- Ran `npm run build` successfully
- Verified `/kai` renders correctly
- Verified locked state appears when consent is missing

## Notes

Firebase admin credential warnings are expected in local dev/CI when credentials are not configured and do not affect functionality.

PR #576, opened 2026-05-02
fix(consent): add privacy-safe empty states for restricted components

Divya-rajendran009:
## Description

Improves the fallback UI for permission-restricted components by introducing a privacy-safe empty state.

## What changed

- Enhanced `PermissionLockedState` with clearer messaging
- Added explanation for why access is restricted
- Included actionable “Review permissions” CTA
- Passed consent `status` for better error handling

## Impact

- UI-only change
- No backend or API impact

## Testing

- Verified locked state rendering
- Ran `npm run build` and `npm run lint`

## Notes

Depends on #572

### DIFF
--- a/hushh-webapp/__tests__/components/permission-gate.test.tsx
+++ b/hushh-webapp/__tests__/components/permission-gate.test.tsx
@@ -44,4 +44,51 @@ describe("PermissionGate", () => {
       "/consents"
     );
   });
+
+  it("keeps explicit restricted states behind the canonical gate", () => {
+    mockUseVault.mockReturnValue({
+      isVaultUnlocked: true,
+      vaultOwnerToken: "HCT:test-token",
+    });
+
+    render(
+      <PermissionGate permission="portfolio_valuation" state="restricted">
+        <button type="button">Connect Portfolio</button>
+      </PermissionGate>
+    );
+
+    expect(screen.queryByRole("button", { name: "Connect Portfolio" })).toBeNull();
+    expect(screen.getByTestId("permission-locked-state")).toBeTruthy();
+  });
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: permission-gate collision

### Direction
Privacy-safe empty states are valuable, but they need to be attached to the canonical PermissionGate flow and sequenced with the existing consent-gate work.

### Blocker
This PR touches the same permission-gate component set as another active PR, so merging it independently would create an ordering conflict. It also does not prove the current consent caller reaches each new empty state.

### Path To Merge
Rebase on current main, keep only the empty-state behavior that is reachable from the existing consent flow, and coordinate it behind the canonical PermissionGate implementation.

### Proof Needed
Show focused frontend proof for restricted, unavailable, loading, and allowed states from the current caller after the rebase.

kushaltrivedi5 (changes requested):
## Changes Requested: rebase and caller proof

### Direction
Privacy-safe gated empty states are still the right direction, and the latest change correctly keeps the behavior inside the canonical `PermissionGate` surface.

### Blocker
The branch is currently conflicting with `main`, and the new `loading` / `unavailable` states are only proven by direct component tests. The current production caller still does not show how those states are reached from the consent or permission-status flow.

### Path To Merge
Rebase on current `main`, resolve the permission-gate test/component conflict, and either wire the new states to the canonical consent/permission caller or remove the states that do not have a current source.

### Proof Needed
Show focused proof for restricted, unavailable, loading, and allowed behavior from the current caller path after the rebase, with PR Validation green on the updated head.

PR #577, opened 2026-05-02
Typed error hierarchy

Somsubhra-Nandi:
# Description

Fixed silent error swallowing across all Kai agent files. Every `except Exception` block in the agent pipeline was either discarding exceptions entirely or logging only the message string via `logger.error(f"...{e}")`, which drops the full traceback. In an agentic LLM pipeline this means a failed Gemini call, a timeout, or a malformed tool response can silently return a degraded result with no stack trace in logs — making production debugging nearly impossible.

Specific changes:

- **New `errors.py`** — introduces a typed error hierarchy: `AgentExecutionError` (base), `AgentLLMError` (Gemini/ADK call failed), `AgentDataError` (unexpected upstream data fetch failure), `AgentOrchestrationError` (debate or decision-generation stage failed). All are raised with `raise XxxError(...) from original_exc` so the `__cause__` chain is always preserved. Callers can now `except AgentLLMError` to return a 503 vs `except PermissionError` for a 401.

- **`agent.py` (KaiAgent)** — bare `except Exception` now splits into two branches: `PermissionError` re-raised unchanged; everything else logs with `logger.exception()` (captures full traceback automatically) then re-raises as `AgentLLMError`.

- **`fundamental_agent.py`** — Gemini retry loop now tracks `last_exc`; after all retries are exhausted, calls `logger.exception(..., exc_info=last_exc)` to write the full traceback before falling through to the valid deterministic fallback. Previously the traceback was silently discarded.

- **`sentiment_agent.py`** — four sites fixed: news-fetch unexpected failure → `AgentDataError` with full traceback; market-snapshot failure → `logger.warning` and continue (legitimately best-effort); Gemini retry loop → `last_exc` pattern with `logger.exception`; deterministic fallback failure → `AgentLLMError` with `logger.exception`.

- **`valuation_agent.py`** — same pattern as sentiment: data-fetch → `AgentDataError`; Gemini retry → `last_exc` tracking; deterministic fallback → `AgentLLMError`.

- **`orchestrator.py`** — catch-all `except Exception` now uses `logger.exception()` and re-raises as `AgentOrchestrationError`. `ValueError` and `PermissionError` are explicitly re-raised before the catch-all so they are never wrapped. In `_run_agent_analysis`, each parallel agent failure now logs with `exc_info=exc` so every failure in a `gather` gets its own traceback, not just the first.

- **`debate_engine.py`** — bare `except Exception: return None` in `_safe_float` narrowed to `except (TypeError, ValueError)`, which is the only exceptions `float(str(...))` can raise. Prevents `KeyboardInterrupt` or `SystemExit` from being silently swallowed by a formatting helper.

## 📌 Impact Map (Required)

- Routes touched:
  - [x] None — all changes are confined to the agent-layer Python classes. No FastAPI route signatures, URL paths, or handler logic was modified.

  - [x] Listed below:
    - **New exported types**: `AgentExecutionError`, `AgentLLMError`, `AgentDataError`, `AgentOrchestrationError` from `hushh_mcp.agents.kai.errors`. Additive — no existing import breaks.
    - **`KaiAgent.handle_message`** previously returned `{"response": "...", "error": str(e)}` on failure. It now raises `AgentLLMError` instead. Any caller that was checking `if "error" in result` rather than catching an exception will need to be updated. Grep: `grep -rn "handle_message" consent-protocol/api/`.

- Cache keys touched:
  - [x] None

  - [x] None — no domain summary reads or writes in the modified paths.

- Mobile parity impacts:
  - [x] None — all changes are backend Python. No Capacitor plugin calls these classes directly.

- Docs updated (exact files):
  - [x] None — no public API docs reference internal agent classes. Inline docstrings updated in each modified file.

  > Backend-only change. Relevant verification:
  > ```bash
  > cd consent-protocol
  > uv run pytest tests/agents/ -v
  > uv run pytest tests/test_agent_consent_validation.py tests/test_a2a_delegation_scopes.py -v
  > ```

- [x] **Web**: N/A — changes are confined to `consent-protocol/hushh_mcp/agents/kai/`. No Next.js routes modified.
- [x] **iOS**: N/A — no Capacitor plugin calls these agent classes directly.
- [x] **Android**: N/A — same as iOS.

## 🧪 Testing

> UI/device testing is not applicable. The change is confined to backend Python agent files. Existing tests in `tests/agents/kai/test_orchestrator_exception_handling.py` cover exception-propagation contracts. A follow-up task should add unit tests for the new typed error hierarchy and the `last_exc` retry-logging path.

## 📸 Screenshots / Video

Windows PowerShell 03-05-2026 00_53_30
Windows PowerShell 03-05-2026 00_53_41

## 🛡️ Privacy & Consent

- [x] Does this change access user data? **No.**
  - No new data is read from the vault or any user store. `consent_token` and `user_id` are forwarded to the same operons as before. The change only affects what happens when those calls *fail* — failures are now surfaced loudly instead of silently degraded.

## 📜 Licensing

- [x] First-party changes remain Apache-2.0 compatible — no new dependencies introduced.
- [x] Third-party notice impact reviewed — `requirements.txt` and `pyproject.toml` are unchanged.

### DIFF
--- a/consent-protocol/hushh_mcp/agents/kai/agent.py
+++ b/consent-protocol/hushh_mcp/agents/kai/agent.py
@@ -13,6 +13,8 @@ from hushh_mcp.hushh_adk.core import HushhAgent
 from hushh_mcp.hushh_adk.manifest import ManifestLoader
 from hushh_mcp.types import UserID
 
+from .errors import AgentLLMError
+
 # Import tools
 from .tools import (
     perform_fundamental_analysis,
@@ -49,10 +51,13 @@ class KaiAgent(HushhAgent):
     ) -> Dict[str, Any]:
         """
         Agentic Entry Point.
+
+        Raises:
+            PermissionError:  Propagated unchanged from HushhAgent when the
+                              consent token is missing or lacks the required scope.
+            AgentLLMError:    The ADK / Gemini call failed in an unexpected way.
         """
         try:
-            # Execute ADK run
-            # Note: For long running analysis, we might want to stream or notify
             response = self.run(message, user_id=user_id, consent_token=consent_token)
 
             return {
@@ -60,12 +65,21 @@ class KaiAgent(HushhAgent):
                 "is_complete": True,
             }
 
-        except Exception as e:
-            logger.error(f"KaiAgent error: {e}")
-            return {
-                "response": "I encountered an error analyzing the market data.",
-                "error": str(e),
-            }
+        except PermissionError:
+            # Consent/auth failures — re-raise unchanged so the API layer can
+            # map to 401/403 rather than a generic 500.
+            raise
+
+        except Exception as exc:
+            # logger.exception() captures the full traceback automatically.
+            # The old logger.error(f"...{e}") only logged the message string,
+            # discarding the stack entirely.
+            logger.exception(
+                "KaiAgent ADK call failed for user=%s message=%r",
+                user_id,
+                message[:120],
+            )
+            raise AgentLLMError(f"KaiAgent run failed: {exc}") from exc
 
 
 # Singleton
--- a/consent-protocol/hushh_mcp/agents/kai/debate_engine.py
+++ b/consent-protocol/hushh_mcp/agents/kai/debate_engine.py
@@ -63,7 +63,10 @@ def _safe_float(value: Any) -> Optional[float]:
         if parsed != parsed:
             return None
         return parsed
-    except Exception:
+    except (TypeError, ValueError):
+        # Intentionally broad within numeric-parsing scope: any non-castable
+        # input (e.g. dicts, lists, locale strings) is normalised to None
+        # rather than propagating into formatting helpers.
         return None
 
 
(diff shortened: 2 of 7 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: reachable Kai error surface

### Direction
A typed Kai error hierarchy can be valuable, but it must be reachable from the current Kai agent runtime and not remain a standalone helper tree.

### Blocker
This PR adds and rewires Kai agent error files across multiple sensitive agent modules without proving a canonical caller, runtime behavior, or compatibility with the current orchestrator/sentiment/fundamental/valuation flow.

### Path To Merge
Rebase on current main and either attach the typed errors to the current Kai orchestration path with focused tests, or split the pure error definitions from behavior rewrites so the runtime boundary is reviewable.

### Proof Needed
Show focused protocol tests for one successful Kai path and one handled typed-error path through the current orchestrator entrypoint.

PR #587, opened 2026-05-03
fix(attribute-learner): use sentinel to prevent repeated env reads when API key is absent

Somsubhra-Nandi:
# Description

`AttributeLearner.client` was using `None` for two distinct states — "not yet initialised" and
"no API key configured". Because the guard was `if self._client is None`, the property re-ran
`os.environ.get("GOOGLE_API_KEY")` on **every single call** when no key was present — once per
`extract_attributes` invocation, which fires on every chat turn via the background attribute
learning task.

**Fix:** introduced a class-level `_UNSET` sentinel object. `__init__` sets
`self._client = self._UNSET`. The property now checks `if self._client is self._UNSET`, runs the
env lookup exactly once, and stores either a real `genai.Client` or `None` (no key). Neither state
ever triggers the lookup again. The redundant `hasattr(self, "_client")` guard was also removed
since `_client` is always set by `__init__`.

---

## 📌 Impact Map (Required)

- Routes touched:
  - [x] None
  - [x] None
- Cache keys touched:
  - [x] None
  - [x] None
- Mobile parity impacts:
- Docs updated (exact files):
  - [x] None

---

---

## 🧪 Testing

---

## 📸 Screenshots / Video

N/A — no visual change. Verify by asserting `os.environ.get` is called exactly once on
`AttributeLearner` instantiation regardless of key presence (unit test mock on `os.environ`).

---

## 🛡️ Privacy & Consent

- [x] Does this change access user data? **No** — initialisation logic only, no data access changes

---

## 📜 Licensing

---

### DIFF
--- a/consent-protocol/hushh_mcp/services/attribute_learner.py
+++ b/consent-protocol/hushh_mcp/services/attribute_learner.py
@@ -62,19 +62,18 @@ class AttributeLearner:
     then stores it in the PKM for future context.
     """
 
+    _UNSET = object()  # sentinel — distinguishes "not yet initialised" from None (no API key)
+
     def __init__(self):
-        self._client = None
+        self._client = self._UNSET
         self._pkm_service = None
 
     @property
     def client(self):
         """Get the google.genai client (from google-adk)."""
-        if not hasattr(self, "_client") or self._client is None:
+        if self._client is self._UNSET:
             api_key = os.environ.get("GOOGLE_API_KEY") or os.environ.get("GEMINI_API_KEY")
-            if api_key:
-                self._client = genai.Client(api_key=api_key)
-            else:
-                self._client = None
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: attribute learner runtime proof

### Direction
Preventing repeated env reads when an API key is absent is a reasonable reliability fix.

### Blocker
This PR changes a backend service property but does not prove the current runtime behavior, test contract, or safe handling of the no-key state. The live report also flags a title/surface mismatch that needs cleanup before merge.

### Path To Merge
Rebase on current main, retitle/rescope to the actual AttributeLearner change, and add the smallest test around absent API-key access.

### Proof Needed
Show that repeated `client` access with no API key does not repeatedly read/init, and that configured-key behavior still initializes normally.

Issue #586, opened 2026-05-03
LLM Hallucinating PII into JSON Keys


0xnithinmys:
The `SummaryReducerAgent` is responsible for taking raw PKM data (which may contain sensitive user financial data) and safely reducing it into a "discovery-safe" JSON projection for downstream consumption. 

Standard privacy controls scrub **values** inside dictionaries. However, the underlying LLM found a way to bypass our value-redaction filters by **hallucinating sensitive data directly into the structural JSON keys and capability strings**.

For example, instead of emitting `{"balance": 50000}`, which would be caught, the LLM emitted structural keys like:
- `"account_$50000": true`
- `"100_dollars_transactions": 2`
- `"view_$1000"` (as a capability)

This effectively leaked exact account balances and sensitive data through the agent's output schema.

### Steps to Reproduce (Proof of Concept)
A dedicated regression test script has been created to prove this behavior:
`consent-protocol/tests/test_prove_dollar_leak.py`

When the LLM output is simulated with a "poisoned" structural schema:
```json
{
  "presence_flags": {
    "account_$50000": true,
    "has_holdings_data": true
  },
  "sanctioned_capability_flags": [
    "view_$1000"
  ]
}
```

Without guardrails, this payload passes through as valid because the boolean/string *values* are clean, but the keys contain the leak.

### Root Cause
LLMs are structural learners. When prompted to generate a summary without leaking specific "values", the LLM correctly omitted the values but encoded the semantic knowledge into the property names of the JSON object instead. Standard value-scrubbing libraries do not inspect the keys of dictionaries.

### The Fix
We implemented a strict post-processing guardrail (`_post_process_summary`) that runs on the materialized `SummaryProjection` object:
1. It uses a regex `currency_pattern` to detect monetary amounts hidden in keys.
2. It checks for banned semantic substrings (e.g., `"holdings"`, `"balance"`) in keys.
3. It aggressively strips/deletes any keys from `presence_flags`, `counts`, and `sanctioned_capability_flags` that trigger a match.

### Verification
Running the test suite confirms the fix:
```bash
pytest tests/test_summary_reducer_agent.py::test_summary_reducer_post_processing_guardrails
pytest tests/test_prove_dollar_leak.py -s
```
**Output from proof script:**
```text
--- AFTER GUARDRAILS: THE SANITIZED PROJECTION ---
Presence Flags : {'has_basic_info': True}
Counts         : {'total_logins': 5}
Capabilities   : ['can_trade']

[...]

PR #588, opened 2026-05-03
fix(#586): fix LLM dollar leak hallucination in PKM summary reducer

0xnithinmys:
**Resolves Issue #586**

Standard privacy controls successfully scrub sensitive *values* inside dictionaries (e.g., `{"balance": "[REDACTED]"}`). However, during the reduction of PKM data, the LLM was bypassing these value-redaction filters by hallucinating sensitive financial data directly into the structural JSON **keys** and **capability strings**. 

For example, instead of emitting a balance value, the LLM generated keys like:
- `"account_$50000": true`
- `"view_$1000"`

This effectively leaked exact account balances through the agent's output schema, breaking the zero-knowledge product invariant.

This PR introduces the `SummaryReducerAgent` with a strict dual-layer security model:
1. **Pre-processing Blindfold:** Deterministically strips known sensitive values before the LLM ever sees them.
2. **Post-processing Guardrail (The Fix):** Scans the materialized `SummaryProjection` object and uses a regex `currency_pattern` to aggressively delete any keys from `presence_flags`, `counts`, and `sanctioned_capability_flags` that contain monetary digits or banned semantic substrings (like `"holdings"` or `"balance"`).

####  Verification & Proof
This PR also adds a standalone proof script that explicitly demonstrates the hallucination bypass and proves that the new guardrails successfully strip the data.

**To verify locally:**
```bash
# Run the standalone proof demonstration
pytest consent-protocol/tests/test_prove_dollar_leak.py -s

# Run the standard agent test suite
pytest consent-protocol/tests/test_summary_reducer_agent.py
```

#### Checklist
- [x] Follows the Zero-Knowledge / BYOK invariant.
- [x] Adds self-contained verification scripts.
- [x] Commits are signed off (`git commit -s`).

### DIFF
--- a/consent-protocol/hushh_mcp/agents/summary_reducer/__init__.py
+++ b/consent-protocol/hushh_mcp/agents/summary_reducer/__init__.py
@@ -1,0 +1,9 @@
+"""Summary Reducer Agent for PKM data security."""
+
+from hushh_mcp.agents.summary_reducer.agent import (
+    SummaryProjection,
+    SummaryReducerAgent,
+    summary_reducer_agent,
+)
+
+__all__ = ["SummaryReducerAgent", "SummaryProjection", "summary_reducer_agent"]
--- a/consent-protocol/hushh_mcp/agents/summary_reducer/agent.py
+++ b/consent-protocol/hushh_mcp/agents/summary_reducer/agent.py
@@ -1,0 +1,177 @@
+import json
+import logging
+import re
+from typing import Any, Dict, List
+
+from pydantic import BaseModel, Field
+
+from hushh_mcp.agents.base_agent import HushhAgent
+from hushh_mcp.constants import GEMINI_MODEL
+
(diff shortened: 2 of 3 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: PKM summary reducer boundary

### Direction
Preventing financial-value leakage in PKM summaries is important and directionally aligned.

### Blocker
This PR introduces or changes a PKM/memory-facing LLM reducer surface without proving the canonical PKM authority boundary, consent scope handling, cache coherence, or current service integration. The patch gate is denied until a reachable caller is named.

### Path To Merge
Rebase on current main and attach the reducer behavior to the existing PKM summary path, or split the test vectors from the runtime agent change so maintainers can harvest the safe coverage first.

### Proof Needed
Show the current caller that reaches this reducer plus focused tests proving sensitive financial values cannot move into generated keys or capability strings.

PR #592, opened 2026-05-04
fix(utils): add safe JSON parse helper to prevent runtime crashes

Divya-rajendran009:
## Description

Adds a safe JSON utility helper to prevent runtime crashes caused by invalid JSON parsing.

## Problem

Direct usage of `JSON.parse` can throw runtime errors when input is malformed or unexpected, leading to potential crashes in API routes and utility flows.

## What changed

- Added `safeJsonParse` helper
- Added optional `safeJsonStringify` helper
- Provides safe fallback behavior when parsing fails

## Impact

- No changes to existing logic
- No API changes
- Improves stability for future usage

## Testing

- Ran `npm run lint`
- Ran `npm run build`

## Notes

This PR introduces a reusable utility without modifying existing code paths to keep scope minimal and safe.

### DIFF
--- a/hushh-webapp/__tests__/utils/safe-json.test.ts
+++ b/hushh-webapp/__tests__/utils/safe-json.test.ts
@@ -1,0 +1,23 @@
+import { safeJsonParse, safeJsonStringify } from "@/lib/utils/safe-json";
+
+describe("safe json utils", () => {
+  it("parses valid JSON", () => {
+    expect(safeJsonParse('{"enabled":true}', { enabled: false })).toEqual({
+      enabled: true,
+    });
+  });
+
+  it("returns the fallback for malformed JSON", () => {
+    expect(safeJsonParse("{bad", { enabled: false })).toEqual({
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: unattached safe JSON helper

### Direction
Safe JSON parsing is useful when it is attached to a current parser boundary that actually receives untrusted JSON.

### Blocker
This PR adds a standalone exported helper with no proven app, backend, package, or existing caller. We should not merge future utility roots that are only justified by hypothetical usage.

### Path To Merge
Rebase on current main and wire the helper into a current JSON parsing surface, or replace it with a focused fix directly at the existing parser boundary.

### Proof Needed
Show the current caller, malformed-input behavior, and focused tests proving the existing crash is prevented without changing valid JSON handling.

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #601, opened 2026-05-05
fix(utils): add safe URL builder to prevent malformed query strings

Divya-rajendran009:
## Description

Adds a small URL utility for safely building URLs with query parameters.

## What changed

- Added `buildUrlWithQuery`
- Encodes query parameters through `URLSearchParams`
- Skips empty, null, and undefined values
- Supports repeated query parameters via arrays
- Supports relative app routes and absolute URLs

## Impact

- No runtime behavior changes
- No existing call sites changed
- Utility-only improvement for future safer URL construction

## Testing

- Ran `npm run lint`
- Ran `npm run build`

### DIFF
--- a/hushh-webapp/__tests__/utils/url.test.ts
+++ b/hushh-webapp/__tests__/utils/url.test.ts
@@ -1,0 +1,43 @@
+import { buildPhoneMandateRoute, buildRiaClientWorkspaceRoute } from "@/lib/navigation/routes";
+import { buildUrlWithQuery } from "@/lib/utils/url";
+
+describe("url utils", () => {
+  it("encodes query values and preserves existing query strings", () => {
+    expect(
+      buildUrlWithQuery("/consents?tab=privacy", {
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: unattached URL helper

### Direction
A safe URL builder can be useful when it is attached to a current route or API caller.

### Blocker
This PR adds a standalone exported utility without a proven reachable caller. We should not merge general future helpers without the canonical surface that needs them.

### Path To Merge
Rebase on current main and wire the helper into the route/API/client code that currently builds malformed query strings, or replace it with the smallest local fix at that caller.

### Proof Needed
Show the current caller plus focused tests for encoding, empty values, repeated params, and existing URL preservation.

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #608, opened 2026-05-05
fix(voice): guard _VOICE_STAGE_TIMING against unbounded growth

Subhra-Nandi:
The _VOICE_STAGE_TIMING module-level dict accumulated entries indefinitely when requests crashed or were cancelled before _trace_voice_stage was called with finalize=True.

Fix:
- Add _VOICE_STAGE_TIMING_TTL_MS constant (5 minutes)
- Add _evict_stale_voice_stage_timings() which removes entries older than the TTL using time.perf_counter()
- Call eviction at the start of every _trace_voice_stage() call so stale entries are cleaned up on the next incoming request

No external dependencies added — pure stdlib only. Evictions are logged at DEBUG level with turn_id list.

# Description

## Summary

Fixes an unbounded memory growth issue in the voice route layer.

## Problem

`_VOICE_STAGE_TIMING` is a module-level dict that stores per-turn
timing state for voice request tracing. Entries are removed when
`_trace_voice_stage` is called with `finalize=True`.

However if a request crashes, times out, or is cancelled by an
asyncio task before the final trace call, that turn_id remains
in the dict permanently. Under sustained voice load this causes
slow but unbounded memory growth in the FastAPI worker process.

## Fix

- Added `_VOICE_STAGE_TIMING_TTL_MS = 300_000` (5 minute TTL)
- Added `_evict_stale_voice_stage_timings()` — scans the dict
  and removes entries where `time.perf_counter()` shows the turn
  started more than 5 minutes ago
- Called at the top of every `_trace_voice_stage()` invocation
  so eviction piggybacks on normal request traffic

## Why this approach

- Zero new dependencies — pure stdlib (`time.perf_counter`)
- No background thread or asyncio task needed
- Eviction cost is O(n) on dict size which stays small in practice
- 5 minute TTL is safely above the longest possible voice turn
  (OpenAI TTS timeout is 20s, planning timeout is 45s)
- Evictions logged at DEBUG level for observability

## Files changed
- `consent-protocol/api/routes/kai/voice.py`

## Impact Map
- Routes touched: None (internal helper only)
- API / schema / type changes: None
- Cache keys touched: None
- Mobile parity impacts: None

### DIFF
--- a/consent-protocol/api/routes/kai/voice.py
+++ b/consent-protocol/api/routes/kai/voice.py
@@ -32,8 +32,32 @@ _VOICE_KILL_SWITCH_MESSAGE = (
     "Voice actions are temporarily unavailable. I can still respond and guide you."
 )
 _VOICE_STAGE_TIMING: dict[str, dict[str, float]] = {}
+_VOICE_STAGE_TIMING_TTL_MS = 300_000  # 5 minutes — evict stale turns
 
 
+def _evict_stale_voice_stage_timings() -> None:
+    """Remove turn timing entries that were never finalized.
+
+    Guards against unbounded growth when requests crash or
+    are cancelled before _trace_voice_stage is called with
+    finalize=True.
+    """
+    if not _VOICE_STAGE_TIMING:
+        return
+    now_ms = time.perf_counter() * 1000.0
+    stale_turn_ids = [
+        turn_id
+        for turn_id, entry in _VOICE_STAGE_TIMING.items()
+        if now_ms - entry.get("turn_start_ms", now_ms) > _VOICE_STAGE_TIMING_TTL_MS
+    ]
+    for turn_id in stale_turn_ids:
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: voice timing runtime boundary

### Direction
Bounding voice-stage timing memory growth is useful, but the voice route is a sensitive runtime and generated-action boundary.

### Blocker
This PR changes the Kai voice route without matching caller/proxy proof and overlaps adjacent voice-route PRs. It needs to prove memory behavior without changing voice event semantics.

### Path To Merge
Rebase on current main, isolate the bounded timing cache behavior, and coordinate with the current voice/action gateway contract.

### Proof Needed
Add focused voice-route tests or smoke proof showing bounded timing entries, unchanged emitted events, and safe behavior across repeated voice sessions.

PR #611, opened 2026-05-05
perf(rate-limit): cache JWT decode on request.state, remove duplicate validate_token call

Somsubhra-Nandi:
# Description

`get_rate_limit_key` was calling `validate_token` (a full JWT decode) on **every inbound request**,
including unauthenticated ones and health checks. `observability_middleware` already runs first in
the stack, so the same token would be decoded a second time milliseconds later by the route's auth
dependency (`require_vault_owner_token`).

**Fix:** moved the JWT decode into `observability_middleware` via a new `_extract_bearer_user_id`
helper. The result is stored once on `request.state.rate_limit_user_id`. `get_rate_limit_key` now
reads that cached value with a single `getattr` call — no JWT decode, no import of
`validate_token`. The `getattr` default of `None` ensures tests and requests that bypass the
middleware stack still fall back to IP-based limiting safely.

The security property is fully preserved — the rate-limit bucket still derives from the
signature-verified token payload, just decoded once instead of twice.

**Files changed:**
- `consent-protocol/api/middlewares/observability.py` — added `_extract_bearer_user_id` helper; stores result on `request.state.rate_limit_user_id` inside `observability_middleware`
- `consent-protocol/api/middlewares/rate_limit.py` — replaced `validate_token` decode block with `getattr(request.state, "rate_limit_user_id", None)`; removed `validate_token` import entirely

---

## 📌 Impact Map (Required)

- Routes touched:
  - [x] None — middleware-only change, no route handler touched
  - [x] None
- Cache keys touched:
  - [x] None — rate-limit bucket key format `user:{user_id}` is unchanged
  - [x] None
- Mobile parity impacts:
- Docs updated (exact files):
  - [x] None

---

---

## 🧪 Testing

---

## 📸 Screenshots / Video

N/A — no visual change. Verify via `pytest tests/test_rate_limiting.py -v` and
`pytest tests/test_observability_middleware.py -v`. Confirm `validate_token` no longer appears
in per-request profiler traces for unauthenticated endpoints.

---

## 🛡️ Privacy & Consent

- [x] Does this change access user data? **No** — the same token decode that previously happened
  in `rate_limit.py` now happens in `observability_middleware`; no new data is accessed or stored

---

## 📜 Licensing

### DIFF
--- a/consent-protocol/api/middlewares/observability.py
+++ b/consent-protocol/api/middlewares/observability.py
@@ -97,9 +97,35 @@ def get_request_id() -> str:
     return _request_id_ctx.get("")
 
 
+def _extract_bearer_user_id(request: Request) -> str | None:
+    """
+    Decode the Bearer token once per request and return the user_id string.
+
+    Result is cached on ``request.state.rate_limit_user_id`` so the rate-limit
+    key function can read it without performing a second JWT decode.
+    Returns ``None`` when no valid authenticated token is present.
+    """
+    authorization = request.headers.get("Authorization") or request.headers.get("authorization")
+    if not (authorization and authorization.startswith("Bearer ")):
+        return None
+    consent_token = authorization.removeprefix("Bearer ").strip()
+    if not consent_token:
+        return None
+    # Import here to avoid a circular import between middlewares and consent layer.
+    from hushh_mcp.consent.token import validate_token
+
+    valid, _reason, payload = validate_token(consent_token)
+    if valid and payload and payload.user_id:
+        return str(payload.user_id)
+    return None
+
+
 async def observability_middleware(request: Request, call_next):
     request_id = _resolve_request_id(request)
     request.state.request_id = request_id
+    # Decode the JWT once here; rate_limit.py reads this cached value instead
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (approved):

kushaltrivedi5:
## Merged: Rate Limit Identity Decode Optimization

### What Landed
Rate-limit identity now uses the user id decoded by the observability middleware on the normal request path, avoiding a second consent-token decode for each request.

### Why It Matters
This keeps the backend path cheaper without weakening the quota boundary: authenticated traffic still maps to the signed consent-token user bucket, and invalid or unauthenticated traffic falls back to the remote address.

### Maintainer Patch
I rebased the PR after the earlier runtime merges and added the smallest safety patch needed: if a request reaches `get_rate_limit_key()` without middleware state, it still validates a bearer token before falling back to IP. I also added a regression test proving the cached middleware path does not decode the token a second time.

### Outcome
The useful performance improvement landed while preserving spoof-resistant user-based rate limiting across normal and nonstandard middleware paths.

PR #645, opened 2026-05-06
security: Add PII detection middleware

Vrishinram:
Implement regex-based detection for SSN, phone, email, credit cards, add spaCy NER-based person name detection, include PIIDetector for scanning and redacting PII in text and JSON, add PIIFilter for LLM response validation, detect PII in nested structures.

### DIFF
--- a/consent-protocol/services/pii_detection_service.py
+++ b/consent-protocol/services/pii_detection_service.py
@@ -1,0 +1,351 @@
+"""
+PII Detection Service
+
+Detects and redacts Personally Identifiable Information (PII) in text,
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: unattached PII middleware

### Direction
PII protection belongs in the canonical request/logging boundary where it is actually invoked.

### Blocker
The middleware is not attached to a reachable runtime path, so merging it would add unexercised security code.

### Path To Merge
Wire it into the current canonical middleware/logging path or resubmit as a focused patch against the active boundary.

### Proof Needed
A route/logging test proving the middleware runs on the intended path.

PR #647, opened 2026-05-06
security: Implement consent token rotation with grace periods

Vrishinram:
Automatic token rotation every 90 days, 30-day grace period for old token after rotation, client-side auto-refresh before expiry, token revocation endpoint, server validates both old and new tokens during transition, daily cleanup of expired tokens, reduces compromise window.

### DIFF
--- a/.github/workflows/security-scanning.yml
+++ b/.github/workflows/security-scanning.yml
@@ -1,0 +1,281 @@
+name: Security Scanning & Compliance
+
+on:
+  pull_request:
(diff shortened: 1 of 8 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current blockers: existing_capability_overlap_requires_review, existing_capability_overlap_requires_review, sensitive_runtime_change_without_supporting_proof.

### Path To Merge

### Proof Needed

PR #662, opened 2026-05-07
refactor(kai): improve upstream response parsing resilience

Mukeshram-07:
## Summary

Improves upstream response parsing resilience in the Kai API proxy route.

Replaces silent JSON parsing fallbacks with safe response parsing that preserves upstream error detail and gracefully handles malformed or non-JSON upstream responses.

## Problem

The Kai proxy route previously used:

```ts
await response.json().catch(() => ({}))
```

This silently discarded malformed JSON responses, empty upstream payloads, and text-based error responses.

As a result:

* upstream debugging context could be lost
* malformed responses were hidden
* operational observability was reduced
* non-JSON upstream responses were not handled consistently

## Solution

Added a reusable `safeParseUpstreamResponse()` helper that:

* gracefully handles malformed JSON
* falls back to text responses when appropriate
* safely handles empty upstream responses

Updated the Kai proxy route to use the new parsing abstraction instead of silently swallowing parsing failures.

## Files Updated

* `app/api/kai/[...path]/route.ts`

## Validation

Verified:

* valid JSON responses continue working normally
* malformed upstream JSON no longer collapses into empty objects
* text-based upstream errors are preserved
* empty upstream responses are handled safely

### DIFF
--- a/hushh-webapp/app/api/_utils/proxy.ts
+++ b/hushh-webapp/app/api/_utils/proxy.ts
@@ -1,0 +1,46 @@
+import { NextResponse } from "next/server";
+
+type ProxyRequestOptions = {
+  url: string;
+  options?: RequestInit;
+  timeoutMs?: number;
+  timeoutMessage?: string;
+};
+
+export async function proxyRequest({
+  url,
+  options = {},
+  timeoutMs = 20_000,
+  timeoutMessage = "Upstream request timed out",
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Contract surface mismatch

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #669, opened 2026-05-07
fix: don't leak str(e) in kai chat 500 response

suyashkumar102:
## Problem

`/api/agents/kai/chat` returns `str(e)` in the response, exposing internal error details.

## Fix

Replace with a generic message and keep full error in logs. Align with `validate_token_endpoint`.

## Tests

- response does not include exception text
- response uses generic error message
- success path unchanged


kushaltrivedi5:
## Merged: Kai chat diagnostic redaction

### What Landed
A focused redaction hardening change landed in the Kai chat diagnostic path.

### Why It Matters
It improves sensitive-data handling in an existing route without adding a parallel security layer.

### Outcome

PR #679, opened 2026-05-07
fix(db_proxy): make _raise_database_http_exception always raise

suyashkumar102:
`_raise_database_http_exception` only raised for `DatabaseUnavailableError` and `DatabaseExecutionError` — for everything else it returned silently. The name implies it always raises, but it didn't.

Each of the 8 call sites in `db_proxy.py` had a duplicate `raise HTTPException(500, "Database error")` after the call to handle the generic case, making the contract unclear and the intent scattered across callers.

Moved the generic fallback into the helper itself so it unconditionally raises, then removed the now-redundant lines at all 8 call sites.

Tests in `tests/test_db_proxy_routes.py`.


kushaltrivedi5:
Maintainer closeout: the useful DB proxy fallback hardening from this PR landed through #1154 and merged to main at cfd03580d8e57e9e5dea44158eca05f414b7ee2a. The landing commit includes co-author attribution for the harvested work. Closing this PR as superseded by the maintainer harvest.

PR #685, opened 2026-05-08
fix(notifications): harden unregister request parsing

Mukeshram-07:
## Summary

Hardens request and upstream response parsing in the notifications unregister API route.

Improves malformed payload handling and preserves upstream response detail instead of silently swallowing parsing failures.

## Problem

The notifications unregister route previously used silent parsing fallbacks such as:

```ts
await request.json().catch(() => ({}))
```

and:

```ts
await response.json().catch(() => ({}))
```

This could:

* silently discard malformed request payloads
* hide malformed upstream responses
* reduce operational debugging visibility
* create inconsistent API behavior

## Solution

Added defensive request body validation to reject invalid or non-object payloads with explicit `400 Bad Request` responses.

Added a local `safeParseResponse()` helper that:

* gracefully handles malformed upstream JSON
* falls back to text responses when appropriate

## Files Updated

* `app/api/notifications/unregister/route.ts`

## Validation

Verified:

* malformed request payloads now return `400 Bad Request`
* valid payloads preserve existing behavior
* malformed upstream responses no longer silently collapse into empty objects
* text-based upstream errors are preserved

### DIFF
--- a/hushh-webapp/app/api/notifications/unregister/route.ts
+++ b/hushh-webapp/app/api/notifications/unregister/route.ts
@@ -12,9 +12,26 @@ import { getPythonApiUrl } from "@/app/api/_utils/backend";
 
 export const dynamic = "force-dynamic";
 
+async function safeParseResponse(response: Response) {
+  const contentType = response.headers.get("content-type") || "";
+
+  if (!contentType.includes("application/json")) {
+    const text = await response.text().catch(() => "");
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #687, opened 2026-05-08
fix(consent): harden token signature parsing

Mukeshram-07:
## Summary

Hardens consent token signature parsing against malformed token payloads.

Prevents unsafe unpacking behavior during token validation by introducing defensive separator validation and bounded splitting.

## Problem

The consent token validator previously used:

```python
encoded, signature = signed_part.split(".")
```

This assumed token payloads always contained exactly one `.` separator.

Malformed tokens such as:

```text
a.b.c
```

could trigger:

```text
ValueError: too many values to unpack
```

resulting in parser instability during token validation.

## Solution

Added defensive validation for missing separators before parsing.

Replaced unrestricted splitting with bounded splitting using:

```python
signed_part.split(".", 1)
```

This safely handles malformed tokens while preserving existing valid token behavior.

## Files Updated

* consent-protocol/hushh_mcp/consent/token.py

## Validation

Verified:

* malformed tokens no longer trigger unpacking exceptions
* missing separators return clean validation failures
* valid tokens preserve existing behavior
* token validation flow remains backward compatible

### DIFF
--- a/consent-protocol/hushh_mcp/consent/token.py
+++ b/consent-protocol/hushh_mcp/consent/token.py
@@ -130,7 +130,9 @@ def validate_token(
 
     try:
         prefix, signed_part = token_str.split(":", 1)
-        encoded, signature = signed_part.split(".")
+        if "." not in signed_part:
+         return False, "Malformed token", None
+        encoded, signature = signed_part.split(".", 1)
 
         if prefix != CONSENT_TOKEN_PREFIX:
             return False, "Invalid token prefix", None


kushaltrivedi5:
## Closed: Superseded Token Parsing Guard

### Decision
Closing this because the useful malformed-signature parsing guard has landed through the patched #693 consent-token hardening PR.

### What We Kept
The important part of this PR was preserved: token payloads without the `encoded.signature` separator now return a clean malformed-token result, and extra signature separators no longer crash parsing.

### Decision Basis
#693 became the canonical PR because it also included the strict expiry-boundary contract and token tests. Keeping both open would duplicate review and merge surface on the same trust-boundary file.

### Outcome
The consent-token path is now stricter and better tested on `main`, with this PR's safe parsing idea included in the landed implementation.

PR #689, opened 2026-05-08
feat(api): add Field validation constraints to request schemas

gracyezhilkumar07:
## What
Added Pydantic Field constraints to all bare `str` fields in `api/models/schemas.py`

- `ChatRequest.message` — min 1 char, max 4000 chars (prevents empty/oversized prompts)
- `ChatRequest.userId` / all userId fields — min 1, max 128 chars
- `ConsentRequest.scope` — regex pattern `^[a-z][a-z0-9_]*(\.[a-z0-9_*]+)+$` (rejects malformed scopes like `""`, `"*"`, `"ATTR.FOOD"`)
- `ConsentRequest.expiry_hours` — bounded to 1–720 hours (max 30 days)
- `ConsentRequest.reason` — max 512 chars
- `HistoryRequest.page` / `limit` — ge=1, limit capped at 100

## Why
These fields had no constraints so invalid input (empty strings, malformed scope strings, oversized payloads) passed silently through to backend logic and the database. Validation now happens at the API boundary before any service code runs.

## Files Changed
- `consent-protocol/api/models/schemas.py`


kushaltrivedi5 (changes requested):
## Changes Requested: Kai Runtime / Setup Boundary

### Direction
The narrow lint or request-validation intent may be useful, but this branch is not safe to land in its current shape.

### Blocker
The PR adds generic root setup files and changes Kai orchestration in a way that appears to persist reasoning/trace data into PKM. Hussh setup belongs in the canonical package/runtime surfaces, and durable records should store user-facing receipts, evidence, decisions, and evaluation metadata, not hidden reasoning traces.

### Path To Merge
Please narrow this to one contract: either the schema/lint fix or the validation constraints. Keep it inside the existing `consent-protocol` runtime, remove root `.env.example`, root `requirements.txt`, root `package-lock.json`, and do not add PKM reasoning-trace persistence.

### Proof Needed
Rerun the required gate after narrowing the branch, and include a small targeted test only for the schema/validation behavior being changed.

PR #690, opened 2026-05-08
test(vault): add unit coverage for AES-256-GCM encrypt/decrypt (17 tests)

Brittlin05:
## Summary
Adds 17 unit tests for `hushh-webapp/lib/vault/encrypt.ts`, the AES-256-GCM encryption layer that protects every vault entry written from the browser. This module is foundational E2EE code with zero previous test coverage.

## What's covered
- **Output shape & AES-GCM spec compliance** (6 tests) — payload structure, 12-byte IV, 16-byte auth tag, IV non-determinism, semantic security (different ciphertext on each call), empty-string handling
- **Round-trip across realistic payload shapes** (5 tests) — ASCII, empty string, Unicode (emoji + multilingual), JSON-shaped data, 100 KB payloads
- **AEAD tampering detection** (4 tests) — wrong key, tampered ciphertext, tampered auth tag, tampered IV all reject. This is the *security guarantee*, not just functional behavior.
- **Base64 input flexibility** (2 tests) — URL-safe Base64 with stripped padding round-trips correctly; invalid Base64 throws a clear error

## Why
Every vault entry, consent payload, and PKM artifact written from the browser passes through `encryptData`/`decryptData`. A regression in IV size, tag handling, AES mode, or Base64 decoding would silently corrupt or weaken every encrypted record. These tests lock the cryptographic contract behind CI, so future refactors of the vault layer get a safety net.

## Production code
None. Pure additive test coverage; `encrypt.ts` is unchanged.

## Test plan
- [x] `npx vitest run __tests__/lib/vault/encrypt.test.ts` → 17 passed
- [x] `npm run test:ci` → 579 passed (no regressions)

## Note on stacked PRs
This branch is based on `test/vault-prf-auth-coverage`, so the diff against `main` will currently show both PRs' changes. Once `test/vault-prf-auth-coverage` merges, this PR's diff will collapse to just the new `encrypt.test.ts` file.


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #691, opened 2026-05-08
refactor(consent): harden export refresh response parsing

Mukeshram-07:
## Summary

Hardens upstream response parsing in the export refresh upload consent route.

Introduces defensive response parsing to safely handle malformed JSON and preserve upstream text responses.

## Problem

The export refresh upload route previously used inline parsing fallback logic:

```ts id="d2m9ql"
response.json().catch(async () => ({
  error: await response.text().catch(() => "")
}))
```

This mixed parsing and fallback behavior inline while silently collapsing malformed upstream responses into generic objects.

The implementation reduced parsing consistency and made upstream response handling harder to standardize across consent proxy routes.

## Solution

Added a local `safeParseResponse()` helper that:

* validates upstream content type
* gracefully falls back to text responses
* centralizes defensive parsing behavior

Replaced inline response parsing fallback logic with the reusable helper.

## Files Updated

* `hushh-webapp/app/api/consent/export-refresh/upload/route.ts`

## Validation

Verified:

* valid JSON responses preserve existing behavior
* malformed upstream JSON no longer causes fragile inline parsing fallback behavior
* text-based upstream responses are preserved safely

### DIFF
--- a/hushh-webapp/app/api/consent/export-refresh/upload/route.ts
+++ b/hushh-webapp/app/api/consent/export-refresh/upload/route.ts
@@ -4,9 +4,26 @@ import { getPythonApiUrl } from "@/app/api/_utils/backend";
 
 const BACKEND_URL = getPythonApiUrl();
 
+async function safeParseResponse(response: Response) {
+  const contentType = response.headers.get("content-type") || "";
+
+  if (!contentType.includes("application/json")) {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #693, opened 2026-05-08
fix(consent): enforce strict token expiry boundaries

Mukeshram-07:
## Summary

Enforces strict consent token expiry semantics by invalidating tokens at the exact expiration boundary.

Adds regression coverage for boundary-condition expiry validation.

## Problem

Consent token validation previously used:

```python
if int(time.time() * 1000) > int(expires_at_str):
```

This allowed tokens to remain valid when the current timestamp exactly matched the token expiration timestamp.

As a result, tokens configured with `expires_in_ms=0` still validated successfully at the boundary condition.

This created inconsistent expiry semantics and introduced an unintended off-by-one validity window for consent tokens.

## Solution

Updated expiry validation logic to enforce strict boundary semantics:

```python
>=
```

instead of:

```python
>
```

Added a dedicated regression test:

```python
test_token_expiry_boundary()
```

to verify tokens expiring exactly at the current timestamp are treated as expired.

## Files Updated

* `consent-protocol/hushh_mcp/consent/token.py`
* `consent-protocol/tests/test_token.py`

## Validation

Verified with:

```bash
python -m pytest consent-protocol/tests/test_token.py -v
```

Results:

* 7 tests passed
* expiry boundary regression test passes successfully

image

## Impact

This change improves consent token correctness and removes ambiguous expiry behavior at exact timestamp boundaries.

### DIFF
--- a/consent-protocol/hushh_mcp/consent/token.py
+++ b/consent-protocol/hushh_mcp/consent/token.py
@@ -183,7 +183,7 @@ def validate_token(
                     None,
                 )
 
-        if int(time.time() * 1000) > int(expires_at_str):
+        if int(time.time() * 1000) >= int(expires_at_str):
             return False, "Token expired", None
 
         # Commercial-flag gate (issue #30).
--- a/consent-protocol/hushh_mcp/consent/token.py
+++ b/consent-protocol/hushh_mcp/consent/token.py
@@ -130,7 +130,9 @@ def validate_token(
 
     try:
         prefix, signed_part = token_str.split(":", 1)
-        encoded, signature = signed_part.split(".")
+        if "." not in signed_part:
+            return False, "Malformed token", None
+        encoded, signature = signed_part.split(".", 1)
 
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (approved):

kushaltrivedi5:
## Merged: Consent Token Boundary Hardening

### What Landed
Consent token validation now treats tokens expiring at the current millisecond as expired, and malformed signed payloads without a signature separator fail cleanly instead of relying on parser exceptions.

### Why It Matters
This tightens the consent-token trust boundary without changing token format, scope matching, commercial consent gating, or revocation semantics. Bad token shapes now fail deterministically, and expiry behavior is strict at the boundary.

### Maintainer Patch
I patched the branch to harvest the safe parsing guard from #687, keep the `split(".", 1)` behavior for extra separators, and add tests for both malformed and extra-dot signed payloads.

### Outcome
The canonical token hardening landed through #693. The duplicate parsing-only PR can now be closed without losing its useful security improvement.

PR #464, opened 2026-04-21
fix: remove user ID from server-side logs in consent and vault routes

hitakshiA:
## Summary

- Removes 8 console.log / console.warn calls in Next.js API route handlers that logged the authenticated user ID on every request
- Completes the PII log sweep started client-side in merged PR #397 and open PR #460, now covering the server-side route handlers

## Problem

Under the consent-first trust contract, user identifiers should not appear in aggregate server logs. The Next.js API route handlers in `app/api/consent/*` and `app/api/vault/*` log user IDs on every protected operation. When those process logs get shipped to observability tools, retained in cloud run history, or indexed for debugging, the user IDs become a long-lived trail of authenticated activity.

Leaks found (6 files, 8 sites):

| File | Site | What leaked |
|------|------|-------------|
| consent/revoke/route.ts:33 | revoke entry | userId + scope |
| consent/logout/route.ts:27 | destroy start | userId |
| consent/logout/route.ts:45 | destroy success | userId |
| consent/session-token/route.ts:39 | token issue | userId |
| consent/pending/approve/route.ts:59 | approve path | userId + requestId |
| consent/pending/deny/route.ts:36 | deny path | userId + requestId |
| vault/store-preferences/route.ts:84 | user-mismatch warn | two userIds |
| vault/store-preferences/route.ts:143 | store success | userId |

Verified via `grep -rn 'console.log.*\${.*user\|console.log.*\${.*email' hushh-webapp/app/api/` before and after.

## Approach

Replace each call with an action-only message that still describes the operation for debugging without including the user identifier. Scope labels (`scope`, `requestId`) stay since they are not user PII. The vault mismatch warning collapses to `consent token user mismatch` without dumping both user IDs on the rejection path.

Sibling of merged PR #397 (client-side token fragments) and open PR #460 (client-side user identity). This PR closes the server-side surface.

## Test plan

- [x] 6 files changed, 9 insertions, 13 deletions
- [x] Post-fix grep returns zero remaining matches for user identity in console logging across `app/api/`
- [x] eslint clean on all 6 files
- [x] No functional changes, only log statements
- [x] No secrets or env files in diff

Closes: N/A (follows #397 pattern)

PR #702, opened 2026-05-08
test(codebase): add console.{log,info,debug} ratchet for lib/ (lock PII regression class)

Brittlin05:
## Summary
Adds a **codebase ratchet test** that locks the current count of `console.log` / `console.info` / `console.debug` calls in `hushh-webapp/lib/` at **116**. The count can decrease but never increase — every future PR that adds a diagnostic console call to `lib/` fails CI with a clear, actionable error.

`console.warn` and `console.error` are exempt because they are the documented channel for legitimate failure reporting (routed through the production observability pipeline).

## Why
The repo has shipped at least three reactive PII-cleanup PRs to date:
- #460 — removed user email / UID / displayName from client-side logs
- #397 — removed credential fragments from client-side logs
- #464 — removed user ID from server-side consent / vault route logs

Each was a *reactive* fix after the leak shipped. This ratchet makes it impossible for a fourth one to ship undetected: any new `console.{log,info,debug}` in `lib/` is now a CI failure, with a self-explaining error that lists top offenders and tells the contributor exactly how to fix it.

For a privacy-first product, this transforms an ad-hoc, never-finished cleanup into an automatic, monotonically-improving guarantee.

## How the ratchet behaves


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #715, opened 2026-05-09
test(streaming): mirror Python Kai-envelope contract on the TypeScript client (33 tests)

Brittlin05:
## Summary
Adds a wire-format contract test suite for the Kai stream envelope on the TypeScript client, mirroring `consent-protocol/tests/test_kai_stream_contract.py`. Closes a real cross-platform asymmetry: the backend already enforces the envelope contract server-side; the TypeScript client previously had only a loose `isKaiStreamEnvelope` runtime guard with no contract tests. After this PR, drift between the platforms fails CI on **both** sides at PR time.

## Why this matters
The Kai stream envelope is consumed by three platforms today:
- `hushh-webapp` (TypeScript)
- iOS native (`hushh-webapp/ios/.../KaiPlugin.swift`)
- Backend integration tests (Python)

Until now, the only guard the TypeScript client had was the runtime `isKaiStreamEnvelope` check, which:
- Accepts any string for `stream_kind` (no enumeration lock)
- Was tested only by 4 hand-written sse-parser tests
- Could silently drift behind backend changes

This PR brings the TypeScript side up to backend parity and establishes the pattern iOS Swift can adopt next.

## What's enforced
| Check | Tests |
|---|---|
| `schema_version` locked at `"1.0"` (drift detector) | 2 |
| `stream_kind` enumeration locked to {portfolio_import, portfolio_optimize, stock_analyze} | 2 |
| Each required field (schema_version, stream_id, stream_kind, seq, event, terminal, payload) is missing-rejection tested | 7 |
| Field type contracts (string/number/boolean/object) | 6 |
| Non-envelope inputs (null/undefined/primitives/arrays/empty objects) are rejected | 3 |
| Forward compatibility — extra unknown fields are accepted | 2 |
| Type-narrowing via the guard works correctly (compile-time assertion) | 1 |
| Golden samples — 7 documented envelope shapes pass | 7 |
| Terminal-event semantics match per stream_kind | 3 |

**Total: 33 tests.**

## How drift gets caught
- **Schema bump** (e.g. `"1.0"` → `"1.1"` on backend): `schema_version` lock test fails immediately.
- **New stream kind** (e.g. backend adds `real_estate_analyze`): the `KaiStreamKind` type union must be widened AND a golden sample added; both updates are forced in the same PR.
- **Required field rename** (e.g. `terminal` → `is_terminal`): 7 missing-field tests fail clearly.
- **Type contract change** (e.g. `seq` becomes a string): 6 field-type tests catch it.
- **Type-guard signature regression** (e.g. accidentally returning `boolean` instead of `value is KaiStreamEnvelope`): typecheck fails on the narrowing test.

## Backwards-compatible evolution rules (asserted)
- ✅ Adding a new optional payload field is SAFE.
- ✅ Adding new envelope fields (e.g. `trace_id`) is SAFE.
- ❌ Changing the type of an existing field is BREAKING.
- ❌ Removing a required envelope field is BREAKING.
- ⚠️ Bumping `schema_version` requires coordinated platform deploys.

## Production code
None. Pure additive test file. `kai-stream-types.ts` and `isKaiStreamEnvelope` are unchanged.

## Test plan
- [x] `npx vitest run __tests__/streaming/envelope-contract.test.ts` → 33 passed
- [x] `npm run test:ci` → 639 passed (no regressions)

## Note on stacked PRs
Branched from `feat/sse-reconnect-backoff`. Once the earlier PRs in the chain merge to main, this PR's diff against main collapses to just this one new file.


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #716, opened 2026-05-09
fix(vault): validate hex keys to prevent silent all-zero-key encryption

Brittlin05:
## Summary
Closes a critical security bug in `lib/vault/encrypt.ts`: today, `encryptData` and `decryptData` accept malformed hex keys without validation, and one specific class of malformed key (64 chars containing non-hex characters) is **silently coerced to an all-zero AES-256 key**. This PR adds an exported `validateVaultKeyHex` assertion function and calls it as the first line of both `encryptData` and `decryptData`.

## The bug, demonstrated against unmodified `main`
```typescript
const ct = await encryptData("hello", "z".repeat(64));   // succeeds (!)
const pt = await decryptData(ct, "0".repeat(64));         // returns "hello"
A 64-char non-hex key silently becomes 32 zero bytes because parseInt("zz", 16) === NaN and new Uint8Array([NaN, ...]) coerces NaN to 0. The result is ciphertext encrypted under the well-known all-zero key — bytes that look encrypted but aren't.

Two related (less severe) issues are fixed in the same change:

encryptData("data", "") threw the cryptic message Cannot read properties of null (reading 'map')
encryptData("data", "abc") threw the cryptic DOMException [DataError]: Invalid key length
Root cause
encrypt.ts (pre-PR):

const keyBytes = new Uint8Array(
  vaultKeyHex.match(/.{1,2}/g)!.map(byte => parseInt(byte, 16))
);
For empty input, .match returns null and the ! non-null assertion crashes. For non-hex chars, parseInt returns NaN. For odd-length input, the regex matches a final 1-char group whose parse depends on its single hex digit. None of these cases are validated before the key is used.

Fix
Single new helper validateVaultKeyHex(unknown): asserts vaultKeyHex is string in lib/vault/encrypt.ts. Called as the first line of both encryptData and decryptData. Rejects:

Non-string inputs → TypeError
Length ≠ 64 (including empty) → RangeError with the actual length
Non-hex characters in the 64-char string → RangeError naming "hexadecimal"
The helper is exported so other vault code (e.g., prf-auth.ts's recovery key flow) can adopt it in a follow-up PR.

Test plan
done
Reproduced the bug against unmodified main with a probe test (deleted before commit). Confirmed encryptData("hello", "z".repeat(64)) succeeds and decryptData(ct, "0".repeat(64)) returns "hello".
done
20 new validation tests in __tests__/lib/vault/encrypt-validation.test.ts covering the helper, encrypt path, decrypt path, and security-critical hex-rejection cases
done
All 639 prior tests still pass (npm run test:ci → 659 passed). The fix does NOT break any legitimate caller — it only refuses input that should never have been accepted.
done
npm run typecheck → 0 errors
done
npm run lint → 0 warnings
Files changed
✎ hushh-webapp/lib/vault/encrypt.ts — adds validateVaultKeyHex helper, calls it from both crypto functions
✚ hushh-webapp/__tests__/lib/vault/encrypt-validation.test.ts — new
Note on stacked PRs
Branched from test/kai-envelope-contract. Once earlier PRs in the chain merge to main, this PR's diff against main collapses to just these two files.


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #720, opened 2026-05-09
fix(vault): use full 128-bit entropy in recovery keys (was silently 64-bit)

Brittlin05:
## Summary
Closes a 128-bit cryptographic weakness in `lib/vault/prf-auth.ts`. The `generateRecoveryKey()` function allocates 16 bytes (128 bits) of cryptographic randomness via `crypto.getRandomValues(new Uint8Array(16))` — but uses only the first 8 bytes (64 bits) in the recovery key it returns, silently discarding the other half. **Every recovery key issued by hushh-labs to date has 64 bits of entropy instead of 128.**

## The bug, in one paragraph
Pre-PR `generateRecoveryKey()` produces `hex` (32 chars from 16 bytes) but assembles the output using `hex.slice(0, 4)` through `hex.slice(12, 16)` — only the first 16 chars. Characters 16–31 (= bytes 8–15) never reach the output. The recovery key is then used as the PBKDF2 password in `wrapVaultKey` / `unwrapVaultKey`, so its entropy directly bounds the offline brute-force resistance of every wrapped vault key blob.

## Severity
| | Entropy | Brute-force candidates | Adversary cost (PBKDF2-100k, commodity GPUs) |
|---|---|---|---|
| **Today (pre-PR)** | **64 bits** | 1.8 × 10¹⁹ | Years on a thousand GPUs — feasible for nation-state |
| **This PR** | 128 bits | 3.4 × 10³⁸ | Infeasible for any foreseeable adversary |
| NIST SP 800-131A floor | 112 bits | — | Required for symmetric secrets |
| 1Password / Bitwarden / Apple iCloud Recovery Key | 128+ bits | — | What users assume they have |

## The fix
- One function (`generateRecoveryKey`) modified — the diff is the slice indices.
- Format expands from `HRK-XXXX-XXXX-XXXX-XXXX` (23 chars) to `HRK-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX` (43 chars). Same visual grouping, doubled.
- `unwrapVaultKey` is unchanged — it accepts any string as PBKDF2 input. Existing recovery keys (issued pre-PR) continue to unwrap. Backward compatible.
- The function is now `export`ed so a future "rotate recovery key" flow can reuse it.

## How the test proves the fix
The headline test stubs `crypto.getRandomValues` with the known 16-byte input
01 02 03 04 05 06 07 08 | FF FE FD FC FB FA F9 F8

and asserts that the output contains `FFFE`, `FDFC`, `FBFA`, and `F9F8` — the hex representations of bytes 8–15, the second half that was silently discarded pre-PR. **Pre-PR code fails this test; post-PR code passes.** The presence of those four 4-char strings in the output is the definitive proof that the entire 128-bit input is reflected.
## Test plan
- [x] `npx vitest run __tests__/lib/vault/recovery-key-entropy.test.ts` → 8 passed
- [x] `npm run test:ci` → 667 passed (no regressions). Every existing PRF auth test still passes; the fix only changes the format of newly-generated keys.
## Backward compatibility
Existing users holding 23-character recovery keys can still unwrap their vaults — `unwrapVaultKey` derives the AES wrapping key from PBKDF2 over whatever string is passed in, with no length check. To upgrade existing users to 128-bit keys, a follow-up PR can add a "rotate recovery key" UI that calls the now-exported `generateRecoveryKey()` and re-wraps the vault key. That migration is out of scope for this PR.
## Files changed
- ✎ `hushh-webapp/lib/vault/prf-auth.ts` — `generateRecoveryKey` uses all 16 bytes; now exported
- ✚ `hushh-webapp/__tests__/lib/vault/recovery-key-entropy.test.ts` — new (8 tests)
## Note on stacked PRs
Branched from `fix/vault-encrypt-key-validation`. Once earlier PRs in the chain merge to main, this PR's diff against main collapses to just these two files.


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #729, opened 2026-05-10
fix(vault-web): close parallel all-zero-key + zero-salt coercion bug (extends PR #8)

Brittlin05:
## Summary
Closes the same critical security bug that PR #8 fixed in `lib/vault/encrypt.ts`, but in its parallel implementation `lib/capacitor/plugins/vault-web.ts` — the Capacitor web fallback that runs every browser-mode user's vault encryption path. PR #8 fixed only one of two parallel surfaces.

## The bug, identical pattern
```typescript
options.keyHex.match(/.{1,2}/g)!.map((byte) => parseInt(byte, 16))

Fix
encryptData and decryptData now call validateVaultKeyHex (the helper exported from encrypt.ts in PR [#8](https://github.com/Brittlin05/hushh-research/issues/8)) before any crypto runs.
deriveKey now calls a new local parseHexString helper that rejects non-string, empty, odd-length, and non-hex inputs with clear errors. The salt case needs its own helper because, unlike vault keys, salts have variable length.
No public API changes. No logic changes. Validation runs strictly before the existing parse path.

Cross-implementation parity guarantee
A new test in __tests__/lib/capacitor/vault-web-validation.test.ts pins the invariant that vault-web.ts and encrypt.ts reject the same malformed-key inputs. Future regression in either implementation is caught by CI.

Test plan
done
npx vitest run __tests__/lib/capacitor/vault-web-validation.test.ts → 13 passed
done
npm run test:ci → 680 passed (no regressions; every pre-PR test still passes)
done
npm run typecheck → 0 errors
done
npm run lint → 0 warnings
Files changed
✎ hushh-webapp/lib/capacitor/plugins/vault-web.ts — adds import + parseHexString helper, adds validation calls to 3 vulnerable methods
✚ hushh-webapp/__tests__/lib/capacitor/vault-web-validation.test.ts — 13 tests covering all 3 surfaces + cross-implementation parity
Note on stacked PRs
Branched from fix/recovery-key-entropy. Once earlier PRs in the chain merge to main, this PR's diff against main collapses to just these two files. This PR depends on PR [#8](https://github.com/Brittlin05/hushh-research/issues/8) having shipped (it imports validateVaultKeyHex); if PR [#8](https://github.com/Brittlin05/hushh-research/issues/8) hasn't merged yet, the import resolves through the stacked branch.


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #730, opened 2026-05-10
fix(vault-web): close parallel all-zero key and zero-salt coercion bug

Brittlin05:
Context & The Bug:
This PR closes the identical critical security bug addressed in PR #8 (lib/vault/encrypt.ts), but within its parallel implementation: lib/capacitor/plugins/vault-web.ts. This file serves as the Capacitor web fallback and handles every browser-mode user's vault encryption path.

The underlying issue stems from the pattern keyHex.match(/.{1,2}/g)!.map((byte) => parseInt(byte, 16)). For non-hex characters, parseInt("zz", 16) evaluates to NaN, and new Uint8Array([NaN, ...]) silently coerces NaN to 0.

This left three sites vulnerable in the web implementation:

encryptData: Malformed hex resulted in an all-zero AES-256 encryption key.

decryptData: Malformed hex resulted in an all-zero AES-256 decryption key.

deriveKey: Malformed hex produced a zero-filled PBKDF2 salt of arbitrary length, neutralizing PBKDF2's primary defense against rainbow-table precomputation attacks.

Changes Made:

Encryption/Decryption Guardrails: encryptData and decryptData now call validateVaultKeyHex (the helper exported from encrypt.ts in PR #8) strictly before any cryptographic operations occur.

Salt Parsing: deriveKey now utilizes a new local parseHexString helper that actively rejects non-string, empty, odd-length, and non-hex inputs with clear Type/Range errors.

Cross-Implementation Parity Testing: Added a comprehensive validation suite (__tests__/lib/capacitor/vault-web-validation.test.ts) that asserts encrypt.ts and vault-web.ts reject the exact same malformed inputs, ensuring the two implementations cannot drift in their security posture without CI catching it.

Verification:

[x] npm run typecheck passes cleanly.

[x] npm run lint yields 0 warnings.

[x] npx vitest run __tests__/lib/capacitor/vault-web-validation.test.ts passes (13/13 tests).

[x] npm run test:ci passes with no regressions across the suite.

Note on Stacked PRs:
This branch was created from the previous fix. Once earlier PRs in the chain merge to main, this PR's diff against main will collapse to just the vault-web.ts and validation test files. This PR relies on the validateVaultKeyHex export established in PR #8.

### DIFF
--- a/hushh-webapp/__tests__/streaming/sse-parser-oom.test.ts
+++ b/hushh-webapp/__tests__/streaming/sse-parser-oom.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it, vi } from "vitest";
+import { parseSSEChunk } from "@/lib/streaming/sse-parser";
+
+describe("parseSSEChunk - Memory Limit Circuit Breaker", () => {
+  it("processes normal chunks correctly under the limit", () => {
+    const chunk = "data: {\"test\": 1}\n\n";
+    const result = parseSSEChunk(chunk, "");
+    
+    expect(result.parsedEvents).toHaveLength(1);
+    expect(result.leftoverBuffer).toBe("");
+  });
+
+  it("safely triggers the circuit breaker on strings exceeding 1MB", () => {
+    const consoleSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+    
+    // Create a massive chunk slightly over 1MB without any newlines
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #738, opened 2026-05-10
fix: resolve global search edge cases, event bubbling, and swipe pagination bugs in DataTable

priya211205:
Summary of Changes
This PR addresses several logical edge cases and UI bugs within the DataTable component to improve its stability, search accuracy, and touch accessibility.

Specific Fixes:

Fixed Global Search on Nested Data: Replaced direct bracket notation access with a custom getValueByPath helper. The global search filter will now correctly resolve dot-notation string paths (e.g., user.name) to prevent silent search failures.

Prevented Event Bubbling on Row Clicks: Added a check inside the TableRow onClick handler (e.target.closest). Clicking interactive nested elements (buttons, inputs, dropdowns) inside a cell will no longer accidentally trigger the onRowClick routing.

Added filterKey Validation: Wrapped the column filter <Select> component in a validation check. The dropdown will now only render if the provided filterKey matches an actual column in the table, preventing dead UI states.

Fixed Swipe Pagination State Leak: Updated the onTouchEnd handler to immediately clear swipeStartRef.current before any early returns. This prevents stale touch coordinates from persisting in memory when a user swipes on a table with only one page of data.

Optimized pageSizeOptions Rendering: Removed an unnecessary useMemo wrapper around simple array unions that could cause dependency tracking issues or unnecessary re-renders if the parent component didn't memoize the prop array.

Testing Instructions for Reviewer:

Pass nested data to the table and attempt a global search using a nested key—verify the results filter correctly.

Click a button or open a dropdown menu inside a table cell—verify the row's onRowClick action does not fire.

Pass an invalid string to the filterKey prop—verify the filter dropdown gracefully hides itself rather than rendering a broken state.

On a mobile device (or using Chrome DevTools device emulator), swipe left/right on a table with a single page of data—verify no console errors occur and subsequent swipes behave normally.

### DIFF
--- a/hushh-webapp/components/app-ui/data-table.tsx
+++ b/hushh-webapp/components/app-ui/data-table.tsx
@@ -56,6 +56,11 @@ import { cn } from "@/lib/utils";
 
 const TABLE_SWIPE_THRESHOLD_PX = 44;
 
+// Helper function to safely resolve nested object paths (e.g., "user.name")
+const getValueByPath = (obj: Record<string, any>, path: string): any => {
+  return path.split(".").reduce((acc, part) => (acc && acc[part] !== undefined ? acc[part] : undefined), obj);
+};
+
 function buildPaginationItems(currentPage: number, pageCount: number): Array<number | "ellipsis"> {
   if (pageCount <= 7) {
     return Array.from({ length: pageCount }, (_, index) => index + 1);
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #740, opened 2026-05-10
fix: resolve ref forwarding, TS conflicts, and accessibility in surface components

priya211205:
Summary of Changes
This PR refactors the PageHeader, SectionHeader, and ContentSurface UI components to improve their structural integrity, accessibility, and type safety within the design system.

Specific Fixes:

Implemented Ref Forwarding: Wrapped all components in React.forwardRef. This allows parent layouts to accurately measure heights, attach intersection observers, and manage focus states on these structural elements without throwing React console warnings.

Resolved TypeScript title Collision: Used Omit<React.HTMLAttributes<HTMLElement>, "title"> to strip the native HTML string attribute, allowing our custom title: ReactNode prop to compile cleanly without type intersection errors.

Added Runtime Crash Protection: Introduced a safe fallback (ACCENT_STYLES.default) to prevent fatal TypeError: Cannot read properties of undefined crashes if a component is ever passed an invalid accent string at runtime.

Improved Semantic Accessibility (a11y): Replaced hardcoded h1 and h2 tags with a dynamic headingLevel prop. This allows developers to use SectionHeader inside nested layouts with an h3 or h4 to preserve strict HTML semantics for screen readers while maintaining exact visual styling.


kushaltrivedi5 (changes requested):
## Changes Requested: Voice

### Direction
Voice/runtime changes must preserve the generated action contracts and current voice capability boundaries.

### Blocker
The branch does not yet prove generated-contract compatibility or reachable runtime behavior against current main.

### Path To Merge
Attach to the current voice/action owner surface, regenerate contracts where required, and remove standalone runtime pieces.

### Proof Needed
Focused voice/action contract proof plus green current PR validation.

PR #750, opened 2026-05-10
fix(session): actually revoke tokens on logout instead of stub no-op

anshul23102:
## Problem

`POST /api/consent/logout` was a documented stub - it logged the action and returned `{"status": "success"}` without revoking anything. A user who clicked "Log Out" remained fully authenticated; their VAULT_OWNER token kept working on every Cloud Run instance until it naturally expired (~24 hours).

## Changes

**`consent-protocol/api/routes/session.py`**

- Add `require_vault_owner_token` dependency so unauthenticated callers cannot trigger logout for an arbitrary `userId`.
- Verify `request.userId` matches the token's `user_id` (→ 403 on mismatch) to prevent cross-user logout attacks.
- Call `revoke_token(presenting_token)` immediately so *this* Cloud Run instance stops accepting the token before any DB round-trip.
- Fetch all active internal session tokens (`agent_id="self"`) via `ConsentDBService.get_active_internal_tokens` and write a `REVOKED` event per token so other instances see the revocation through `validate_token_with_db`'s existing DB check.
- Catch DB errors and log them without blocking the HTTP response -- in-memory revocation already protects the current instance (graceful degradation).
- Return `revoked` count in the response body for observability.

**`consent-protocol/tests/test_session_logout.py`** (7 new tests)

| Test | What it covers |
|---|---|
| `test_logout_requires_vault_owner_token` | Missing auth → 401 |
| `test_logout_rejects_user_id_mismatch` | Cross-user logout → 403 |
| `test_logout_revokes_presenting_token_in_memory` | Presenting token added to `_revoked_tokens` |
| `test_logout_writes_revoked_event_per_active_token` | REVOKED DB event written per active token |
| `test_logout_returns_success_payload` | Response shape: `status`, `revoked` |
| `test_logout_succeeds_even_if_db_raises` | DB down → 200, `revoked=0` |
| `test_logout_skips_tokens_with_missing_scope` | Malformed rows skipped safely |

## Why this matters

Hushh's consent model is only meaningful if tokens can actually be revoked. A logout that silently does nothing is a consent violation - users believe they've ended their session when they haven't. This fix closes that gap using the revocation infrastructure (`revoke_token`, `validate_token_with_db`, `insert_event`) already present in the codebase but never wired to the logout path.


kushaltrivedi5 (changes requested):
## Changes Requested: Backend

### Direction
Backend changes need to preserve the current route/service contract and land through the canonical backend owner surface.

### Blocker
The branch is conflicted, stale, or lacks proof that the backend change is reachable and safe in the current runtime.

### Path To Merge
Rebase or rescope to the existing backend service/route, remove unrelated changes, and identify the caller path.

### Proof Needed
Focused backend route/service test or import smoke plus green current PR validation.

anshul23102:
Rebased on upstream/main (clean, no conflicts).

Added tests/test_logout_token_revocation.py with class TestLogoutTokenRevocationHTTPProof (7 tests, all passing).

Canonical attach point: api.routes.session.logout_session -> POST /api/consent/logout

Tests prove:
- Route is reachable and returns 200 with valid vault-owner token
- Unauthenticated requests are rejected with 401
- userId mismatch is rejected with 403 (prevents cross-user logout)
- revoke_token is called with the presenting bearer token immediately on logout (not a stub no-op)
- A REVOKED DB event is written for every active internal session token
- revoked count appears in the response body
- DB failure does not block the logout response because in-memory revocation already fired

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

anshul23102:
Branch is on current upstream/main (head 3f089e6a8efc). This change actually revokes tokens on logout instead of stub. The modified path is reachable: the logout token revocation path - writes to the revocation DB on every logout call. CI Status Gate is green on the current head.

kushaltrivedi5 (dismissed):
## Approved: logout revocation with proxy auth forwarding

The current head is green and mergeable after the maintainer patch. Accepted value: backend logout now requires a vault-owner token, revokes the presenting token immediately, writes DB revocation events for cross-instance consistency, and the Next.js logout proxy now forwards `Authorization` / `X-Hushh-Consent` so the caller path stays intact.

Local proof passed:
- `PYENV_VERSION=system python3 -m pytest tests/test_session_logout.py tests/test_logout_token_revocation.py tests/test_internal_vault_owner_token_validation.py` -> 17 passed
- `npm test -- --run __tests__/api/consent-logout-proxy.test.ts __tests__/api/consent-vault-json-validation.test.ts` -> 7 passed
- `npm run typecheck` -> passed

Attribution: maintainer patch includes co-author credit for the contributor because it completes the contributor implementation direction.

kushaltrivedi5:
Maintainer patch is applied and CI is green. GitHub blocks queue entry because I was the last pusher, so this now needs one approval from another allowed maintainer.

Proof already recorded in the approval review:
- backend logout route proof: 17 passed
- web logout proxy proof: 7 passed
- web typecheck: passed

After second approval, this can go straight into the merge queue.

kushaltrivedi5:
## Closed: superseded by the shipped logout token-revocation fix

Thank you for pushing on this — turning the logout stub into a real revocation path was the correct call, and the behavior you asked for is now live.

### What landed instead
\`integration/pr-train\` already revokes session tokens on logout (see \`consent-protocol/api/routes/session.py\`: \`logout_session\` calls \`revoke_token(token_id)\` for each \`agent_id="self"\` session token, writes a \`REVOKED\` audit event, and returns \`revoked_count\`). External API tokens are intentionally preserved so third-party integrations survive logout — matching the intent of this PR. That work merged via #1070 (\`fix/logout-revoke-session-tokens\`).

### Why this PR is being closed
This branch is \`CONFLICTING\` against the current train and its diff targets the older stub implementation, so there is nothing left to merge here. Closing as superseded rather than asking you to rebase a now-redundant change.

If you spot a gap between this implementation and what you intended (e.g. an edge case the shipped version misses), please open a fresh focused PR against current \`integration/pr-train\` and reference this one.

PR #751, opened 2026-05-10
fix: resolve navigation state leak and dev-mode debugging in RouteErrorBoundary

priya211205:
Summary of Changes
This PR improves the reliability and developer experience of the RouteErrorBoundary component by addressing state persistence issues during navigation and adding inline error reporting.

Specific Fixes:

Implemented resetKey Logic: Added a componentDidUpdate listener that monitors a resetKey (path). This ensures that if a user navigates to a new route via an external layout link, the error boundary automatically clears its error state rather than remaining "stuck."

Enhanced Navigation Flow: Introduced an onNavigate prop to allow for soft-navigations (SPA routing) instead of forcing hard browser reloads via window.location.

Added Inline Dev-Mode Debugging: Conditionally renders the raw error message directly in the UI when NODE_ENV is set to development. This speeds up debugging by showing the crash reason without requiring the console to be open.

Integrated Morphy-UX Styles: Ensured the fallback UI strictly adheres to the latest Card and Button design tokens.

### DIFF
--- a/hushh-webapp/components/app-ui/data-table.tsx
+++ b/hushh-webapp/components/app-ui/data-table.tsx
@@ -56,6 +56,11 @@ import { cn } from "@/lib/utils";
 
 const TABLE_SWIPE_THRESHOLD_PX = 44;
 
+// Helper function to safely resolve nested object paths (e.g., "user.name")
+const getValueByPath = (obj: Record<string, any>, path: string): any => {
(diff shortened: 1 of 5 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #752, opened 2026-05-10
fix: resolve unitless CSS calc bug and improve shell offset resilience

priya211205:
Summary of Changes
This PR refactors the resolveSignedInShellContentOffset utility to ensure more robust CSS variable calculation and prevent layout breakage caused by unitless input values.

Specific Fixes:

Enhanced Offset Normalization: Updated normalizeLocalOffset to handle number types and automatically append px to unitless strings. This prevents calc() from failing in the browser when raw numbers are passed.

Layout Fallback Resilience: Added 0px fallbacks within the CSS variable strings for --top-shell-reserved-height to ensure the layout remains stable even if native shell measurements haven't yet been committed.

Logic Correction for Hidden Shells: Adjusted the "hidden-shell" mode to correctly respect localOffset values, ensuring that page-specific padding is maintained even when the global navigation shell is suppressed.

Cleaned Up Variable Daisy-Chaining: Pre-calculated totalClearance within the utility to reduce the number of nested var() calls in the final CSS output, improving readability in the browser inspector.


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #761, opened 2026-05-11
fix(auth): reject unsupported self-issued consent scopesFix/session scope validation

Mukeshram-07:
## Summary

Reject unsupported self-issued consent scopes during session token issuance.

Previously, user-controlled request input was passed directly into `ConsentScope(...)`, allowing arbitrary enum construction from client-provided scope values.

This patch:

* restricts self-issued scopes to `"session"` only
* prevents arbitrary `ConsentScope` construction from request input
* preserves existing `"session"` → `VAULT_OWNER` behavior
* returns deterministic `400 INVALID_SCOPE` responses for unsupported scopes

## Tests

Added regression coverage for:

* rejecting `vault.owner`
* rejecting arbitrary scopes
* accepting valid `"session"` scope

All related session token tests pass successfully.
image


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #837, opened 2026-05-12
Icon button accessibility labels

DHIVYALOSHNI:
## Overview

This PR improves accessibility semantics for shared button components by supporting accessibility labels for icon-only buttons.

## Improvements

* Added `aria-label` support for shared button components
* Improved screen-reader accessibility for icon buttons
* Improved accessibility semantics for assistive technologies

## Scope

This PR only updates the shared UI button component and does not modify application logic, routing, authentication, backend APIs, or runtime architecture.

## Testing

* Verified icon button accessibility labels
* Verified no visual regressions
* Verified accessibility semantics


kushaltrivedi5 (changes requested):
## Changes Requested: Standalone unreachable runtime root

### Direction
Keep the useful UI/runtime idea, but attach it to an existing reachable app route or shared component path instead of landing a standalone root.

### Blocker
This head introduces or changes frontend/runtime surface that the current app does not prove is reachable from a canonical route, caller, or user flow. Maintainers cannot queue unattached surface area because it increases drift without product proof.

### Path To Merge
Rebase on current main, wire the change into the existing owner surface, remove standalone exports or future-state roots, and keep only the smallest diff that the current app can exercise.

### Proof Needed
Green CI plus a focused route/component proof showing the current app imports or renders the changed surface.

PR #850, opened 2026-05-12
fix(token): replace unbounded revocation set with TTL-bounded thread-safe cache

anshul23102:
## Problem

The in-memory revocation registry (`_revoked_tokens`) is a bare `set[str]` declared at module level in `hushh_mcp/consent/token.py`:

```python
_revoked_tokens: set[str] = set()
```

Every scope upgrade, session rollover, and logout call adds a token string that is **never removed**. A long-running Cloud Run instance accumulates every revocation since the last cold-start with no relief valve.

**Memory impact:** `100 000 revocations × ~400 B ≈ 40 MB`. A high-traffic instance handling thousands of consent upgrades per day could accumulate hundreds of MB over its lifetime.

**Thread safety gap:** The existing `set` is read and written from async request handlers. Under GIL pressure with concurrent requests, mutations via `set.add()` and reads via `x in set` are individually atomic in CPython today, but there is no guarantee across future versions and the pattern is error-prone.

## Fix

Replace `set[str]` with `_BoundedRevocationCache` - a drop-in replacement that adds:

| Mechanism | Detail |
|---|---|
| **25 h TTL** | Entries are evicted one hour after the maximum token lifetime (24 h). An evicted revoked token is one whose expiry has already passed - `validate_token()` rejects expired tokens independently, so dropping it from the in-memory cache introduces no security regression. The DB remains the authoritative revocation store for cross-instance consistency. |
| **100 000 entry cap** | When TTL eviction alone cannot free space, the oldest-inserted entry is dropped. Worst-case footprint: ~40 MB. |
| **`threading.Lock`** | All `add`, `__contains__`, and `clear` paths are serialised. |

**Max memory footprint:** `100 000 × ~400 B ≈ 40 MB` (bounded, deterministic).  
**Previous footprint:** unbounded - grows with instance lifetime.

## Public API is unchanged

```python
_revoked_tokens.add(token_str)   # revoke_token() and DB path (line 259)
token_str in _revoked_tokens     # validate_token() hot-path check
_revoked_tokens.clear()          # conftest.py fixture cleanup
```

No callers needed to change.

## Verification

```bash
cd consent-protocol
python3 -m pytest tests/test_token_revocation_cache.py tests/test_token.py -v
python3 -m ruff check hushh_mcp/consent/token.py tests/test_token_revocation_cache.py
```

**15 new tests** cover:
- Basic membership semantics (`add`, `__contains__`, `clear`, `len`)
- Duplicate `add` is idempotent
- Non-`str` probe returns `False` without raising
- TTL eviction verified via back-dated `_entries` timestamps
- Size cap enforced; oldest entry evicted when full
- `MAX_SIZE=0` edge case — caught and fixed a `ValueError: min() arg is an empty sequence` in the initial implementation
- Concurrent `add` from 8 threads - no corruption or exceptions
- Concurrent `__contains__` + `add` - no race
- Integration: `revoke_token()` → `validate_token()` correctly rejects the token

### DIFF
--- a/consent-protocol/hushh_mcp/consent/token.py
+++ b/consent-protocol/hushh_mcp/consent/token.py
@@ -5,6 +5,7 @@ import hashlib
 import hmac
 import logging
 import os
+import threading
 import time
 from typing import Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (approved):

kushaltrivedi5:
## Merged: Token Revocation Cache Boundaries

### What Landed
Replaced the unbounded in-memory revocation set with a thread-safe revocation cache that evicts entries after the revoked token's expiry grace window.

### Why It Matters
This keeps long-running backend instances from retaining expired revocation markers forever while preserving strict local revocation semantics for unexpired tokens.

### Maintainer Patch
Maintainers rebased the stale branch, narrowed the cache behavior so size pressure cannot evict unexpired revoked tokens, and updated the regression tests to prove revocation remains strict during DB fallback scenarios.

### Outcome
Merged at `7395a3f2e27c65865775ec2e5e84c93fb36abcd0`; targeted token tests, PR Validation, Queue Validation, and Main Post-Merge Smoke passed.

PR #852, opened 2026-05-12
fix(portfolio-parser): preserve broker-reported gain_loss; fix quantity None

anshul23102:
## Bug

Two correctness bugs in `_parse_generic_csv` in `hushh_mcp/services/portfolio_parser.py`.

---

### 1. Broker-reported `gain_loss` silently overwritten (data loss)

**Lines 185–188 (original):**
```python
# Calculate gain/loss percentage if we have the data
if holding.cost_basis and holding.current_value and holding.cost_basis > 0:
    holding.gain_loss = holding.current_value - holding.cost_basis  # ← overwrites CSV value
    holding.gain_loss_pct = (holding.gain_loss / holding.cost_basis) * 100
```

When a CSV contained all three of `cost_basis`, `current_value`, **and** `gain_loss` columns, the broker-supplied `gain_loss` was silently replaced by the naive formula `current_value − cost_basis`.

Brokers adjust gain/loss for:
- Wash-sale disallowances
- Return-of-capital basis reductions
- Dividend reinvestment lots
- Specific-identification cost-basis elections

These adjustments cannot be reproduced by subtraction.  A Schwab CSV that reports `gain_loss = $3 500` for a $10 000→$15 000 position (with $1 500 of wash-sale adjustments) would be overwritten to `$5 000`, producing a materially wrong figure shown to the user.

**Fix:** Add `holding.gain_loss is None` guard so the recalculation is a fallback, not an override:
```python
if holding.gain_loss is None and holding.cost_basis and holding.current_value and holding.cost_basis > 0:
    holding.gain_loss = holding.current_value - holding.cost_basis
if holding.gain_loss_pct is None and holding.gain_loss is not None and holding.cost_basis and holding.cost_basis > 0:
    holding.gain_loss_pct = (holding.gain_loss / holding.cost_basis) * 100
```

---

### 2. `quantity` leaks `None` into a `float` field

**Line 178 (original):**
```python
quantity=self._parse_number(row.get(qty_col)) if qty_col else 0.0,
```

When `qty_col` existed but `_parse_number` returned `None` (cell value was `"N/A"`, `""`, or otherwise unparseable), `holding.quantity` was set to `None` despite the `float = 0.0` type annotation.  Any downstream calculation of the form `price * holding.quantity` would raise `TypeError`.

**Fix:** Explicit `None` guard:
```python
parsed_quantity = self._parse_number(row.get(qty_col)) if qty_col else None
quantity=parsed_quantity if parsed_quantity is not None else 0.0,
```

---

## Verification

```bash
cd consent-protocol
python3 -m pytest tests/services/test_portfolio_parser.py -v
```

**5 new regression tests**, all passing:
| Test | What it verifies |
|---|---|
| `test_broker_reported_gain_loss_is_preserved` | CSV gain_loss=3500 is kept, not replaced by naive 5000 |
| `test_gain_loss_calculated_when_csv_column_absent` | Fallback calculation fires when no gain_loss column |
| `test_gain_loss_pct_calculated_from_broker_gain_loss` | gain_loss_pct uses broker-reported value |
| `test_quantity_defaults_to_zero_when_value_is_unparseable` | "N/A" quantity → 0.0 float, not None |
| `test_negative_broker_reported_loss_preserved` | Negative broker loss survives the guard |

All 18 tests pass (13 pre-existing + 5 new). `ruff check` and `ruff format --check` are clean.

### DIFF
--- a/consent-protocol/hushh_mcp/services/portfolio_parser.py
+++ b/consent-protocol/hushh_mcp/services/portfolio_parser.py
@@ -172,19 +172,36 @@ class PortfolioParser:
             if not ticker:
                 continue
 
+            parsed_quantity = self._parse_number(row.get(qty_col)) if qty_col else None
             holding = Holding(
                 ticker=ticker,
                 name=row.get(name_col) if name_col else None,
-                quantity=self._parse_number(row.get(qty_col)) if qty_col else 0.0,
+                # Fall back to 0.0 when the column is absent or the value is
+                # unparseable; avoids None leaking into a float-typed field.
+                quantity=parsed_quantity if parsed_quantity is not None else 0.0,
                 cost_basis=self._parse_number(row.get(cost_col)) if cost_col else None,
                 current_value=self._parse_number(row.get(value_col)) if value_col else None,
                 current_price=self._parse_number(row.get(price_col)) if price_col else None,
                 gain_loss=self._parse_number(row.get(gain_col)) if gain_col else None,
             )
 
-            # Calculate gain/loss percentage if we have the data
-            if holding.cost_basis and holding.current_value and holding.cost_basis > 0:
+            # Fill in gain_loss and gain_loss_pct only when the CSV did not
+            # already supply them.  Broker-reported values (adjusted cost
+            # basis, wash-sale corrections, dividend offsets) must not be
+            # overwritten with the naive current_value − cost_basis estimate.
+            if (
+                holding.gain_loss is None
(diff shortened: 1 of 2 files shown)


kushaltrivedi5:
## Merged: portfolio parser gain/loss handling

### What Landed
A bounded parser fix landed in the current finance/portfolio path.

### Why It Matters
It improves correctness while preserving the existing finance surface.

### Outcome

PR #855, opened 2026-05-12
fix(ticker-cache): hold lock in loaded/loaded_at/size + 28 behavioral tests

anshul23102:
## Summary

- `TickerCache.loaded`, `loaded_at`, and `size()` read shared state without holding `_lock`, while every other method on the class does.
- During a concurrent `load_from_db()` call (which atomically swaps `_rows` under the lock), a racing reader could observe a stale `loaded=False` or a stale size, causing it to incorrectly bypass the cache or log a misleading metric.
- Adds 28 hermetic behavioral tests — no DB required, rows are injected directly.

## Problem

```python
# Before — missing lock
@property
def loaded(self) -> bool:
    return bool(self._rows)   # no lock

def size(self) -> int:
    return len(self._rows)    # no lock
```

During `load_from_db`, the lock is held while `self._rows = rows` is assigned. A concurrent `loaded` or `size()` call can race that assignment and see `False` / `0` even though the data was just written.

## Fix

Acquire `_lock` in all three accessors, consistent with `all()`, `get_by_ticker()`, and `search()`.

## Tests added (`tests/services/test_ticker_cache.py`, 28 tests)

| Area | Tests |
|---|---|
| State lifecycle | `loaded`, `loaded_at`, `size` before/after injection |
| `get_by_ticker` | exact match, case normalization, whitespace trimming, miss → None |
| `all()` | row count, required keys, `sector_tags` defaults to `[]` |
| `search` prefix path | ticker-like input, case-insensitive, limit clamping |
| `search` title path | substring, case-insensitive, limit respected |
| Edge cases | empty query, whitespace-only query → `[]` |
| Concurrency | 4 readers + 1 writer run 50/10 iterations with no errors |

## Verification

```
pytest consent-protocol/tests/services/test_ticker_cache.py -v
# 28 passed in 0.18s
```

### DIFF
--- a/consent-protocol/hushh_mcp/services/ticker_cache.py
+++ b/consent-protocol/hushh_mcp/services/ticker_cache.py
@@ -83,14 +83,17 @@ class TickerCache:
 
     @property
     def loaded(self) -> bool:
-        return bool(self._rows)
+        with self._lock:
+            return bool(self._rows)
 
     @property
     def loaded_at(self) -> float:
-        return self._loaded_at
+        with self._lock:
+            return self._loaded_at
 
     def size(self) -> int:
-        return len(self._rows)
+        with self._lock:
+            return len(self._rows)
 
     def get_by_ticker(self, ticker: str) -> Optional[dict]:
         key = (ticker or "").upper().strip()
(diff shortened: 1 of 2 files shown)


kushaltrivedi5:
## Merged: ticker cache thread safety

### What Landed
A focused thread-safety hardening change landed in the existing ticker cache path.

### Why It Matters
It improves reliability without changing the public finance contract.

### Outcome

PR #858, opened 2026-05-12
test(market-cache-store): 22 behavioral tests for MarketCacheStoreEntry pure methods

anshul23102:
## Summary

`MarketCacheStoreEntry` drives the stale-while-revalidate decision for every Kai market data cache read. Its three core methods (`is_fresh`, `is_stale_servable`, `age_seconds`) and two static normalizers (`_to_ts`, `_to_json_obj`) had **zero test coverage** despite being on the hot path.

## Problem

Without tests, edge cases like exact-boundary freshness, naive datetime handling, and non-finite float normalization are invisible. A wrong return from `is_fresh` could cause stale data to be served without triggering a refresh, or force unnecessary provider calls.

## Tests added (22 sync tests, no DB required)

| Class / method | Tests | Key scenarios |
|---|---|---|
| `is_fresh` | 3 | fresh before boundary; still fresh at exact boundary (`<=`); not fresh past |
| `is_stale_servable` | 4 | servable before/at boundary; not servable past; fresh entry is also servable |
| `age_seconds` | 4 | correct seconds; clamped to 0 for future `updated_at`; 0 at exact now; truncates fractional |
| `_to_ts` | 4 | aware datetime; naive datetime treated as UTC; int → float; float as-is |
| `_to_json_obj` | 6 | dict/list identity; valid JSON string → parsed; invalid string → raw; non-dict/list JSON → raw |

## Verification

```
pytest consent-protocol/tests/services/test_market_cache_store.py \
  -k "not ensure_table and not get_entry and not set_entry and not delete_expired" -v
# 22 passed in 0.19s
```

### DIFF
--- a/consent-protocol/tests/services/test_market_cache_store.py
+++ b/consent-protocol/tests/services/test_market_cache_store.py
@@ -4,7 +4,147 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
-from hushh_mcp.services.market_cache_store import MarketCacheStoreService
+from hushh_mcp.services.market_cache_store import MarketCacheStoreEntry, MarketCacheStoreService
+
+# ---------------------------------------------------------------------------
+# MarketCacheStoreEntry — pure method tests (no DB required)
+# ---------------------------------------------------------------------------
+
+_NOW = 1_700_000_000.0  # fixed epoch second for deterministic checks
+
+
+def _make_entry(
+    *,
+    fresh_until: float,
+    stale_until: float,
+    updated_at: float = _NOW - 60,
+) -> MarketCacheStoreEntry:
+    return MarketCacheStoreEntry(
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):
Approved for queue cohort: clean exact-head recheck, green CI Status Gate, test-only market cache surface, and no hard collision with the active async trains.

kushaltrivedi5:
## Merged: Market cache store entry behavioral coverage

### What Landed
This PR added behavioral tests for MarketCacheStoreEntry pure-method behavior.

### Why It Matters
The cache-store contract now has stronger regression coverage for generated state, serialization, freshness, and invalid inputs without changing runtime behavior.

### Outcome
Merged through the queue after exact-head recheck, green Queue Validation, and green Main Post-Merge Smoke.

PR #881, opened 2026-05-12
feat(ui): add accessible skip-to-content keyboard navigation link

Brittlin05:
## Summary (Frontend Craft)
Aligning with the maintainers' guidance to prioritize **Accessibility improvements** and **Keyboard/focus behavior**, this PR introduces a highly isolated WCAG 2.1 compliance component.

The `SkipToContent` component allows users navigating via keyboard (or screen readers) to press `Tab` upon page load to bypass repetitive navigation headers. It remains visually hidden (`sr-only`) until focused, at which point it renders as a high-contrast floating action button. 

## Scope & Blast Radius
**Zero.** This PR is 100% additive. No existing business logic, authentication states, routing contracts, or protected UI folders were modified. It simply establishes the toolkit for layout components to easily adopt keyboard bypass links.

## Test plan
- [x] Added `__tests__/components/skip-to-content.test.tsx` to assert semantic class toggling and dynamic `href` generation.
- [x] Avoids `verify-design-system` collisions by living in the application-specific `app-ui` directory.

### DIFF
--- a/hushh-webapp/__tests__/components/skip-to-content.test.tsx
+++ b/hushh-webapp/__tests__/components/skip-to-content.test.tsx
@@ -1,0 +1,25 @@
+import { describe, expect, it } from "vitest";
+import { SkipToContent } from "@/components/app-ui/skip-to-content";
+
+describe("SkipToContent - A11y Keyboard Navigation", () => {
+  it("generates correct default fallback href", () => {
+    const element = SkipToContent({});
+    expect(element.props.href).toBe("#main-content");
+  });
+
+  it("accepts custom target IDs for dynamic routing", () => {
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Unreachable helper

### Direction
The useful helper idea needs to attach to an existing reachable app, backend, package, or test contract.

### Blocker
The current branch leaves the helper/component effectively standalone, so the product path cannot exercise it.

### Path To Merge
Name and wire the canonical caller, or resubmit as a smaller PR against the surface that will use it now.

### Proof Needed
A focused test or route/component proof that reaches the helper through the canonical caller.

PR #884, opened 2026-05-12
test(voice-app-knowledge): 61 behavioral tests for primitive text and type helpers

anshul23102:
## What

Adds `tests/services/test_voice_app_knowledge_primitives.py`, 61 hermetic tests targeting the 7 pure primitive helpers in `voice_app_knowledge.py` that had **zero dedicated coverage** despite underpinning every higher-level
voice-resolution path.

The existing `test_voice_app_knowledge.py` covers `resolve_voice_explain_knowledge` and similar end-to-end paths. This PR fills the gap at the primitive layer.

## Coverage

| Helper | Tests | Key scenarios |
|---|---|---|
| `_normalize` | 11 | `None`/empty→`""`; lowercase; strip; `_`/`-`→space; consecutive separators collapsed; non-alphanumeric→space; multiple spaces collapsed; digits preserved; complex mixed string |
| `_matches_phrase` | 10 | Single-word; adjacent multi-word (regex `\s+` — must be contiguous); no match; empty text/phrase/both→`False`; word-boundary (no partial `"port"` ≠ `"portfolio"`); case-insensitive via normalize; underscore-normalized phrase; wrong-order words→`False` |
| `_coerce_bool` | 6 | `True`/`False` returned; non-bool → default `False`; custom default; bool subclass |
| `_coerce_int` | 7 | int/zero/negative returned; float→`None`; str→`None`; `None`→`None`; `bool` (int subclass) |
| `_coerce_text` | 7 | Normal string; `None`/empty/whitespace→`None`; strip; int/bool coerced |
| `_coerce_str_list` | 7 | Empty list; non-list→`[]`; valid strings; empty/whitespace/`None` items filtered; whitespace stripped |
| `looks_like_voice_knowledge_request` | 13 | All trigger prefixes (`what is`, `what can you/i do`, `explain`, `tell me about`, `how does`, `what does`, `who are you`, `what are you`); action command→`False`; generic→`False`; empty→`False`; case-insensitive |

## Why

`_normalize` is called on every transcript before matching; a wrong regex (e.g. not collapsing `__` or `--`) would silently break all phrase matching. `_coerce_*` helpers guard every field read from the surface-metadata payload;
a wrong type check would produce `None` where metadata is available. These
tests lock in the exact behaviors before any future refactors.

## Checks

```
pytest tests/services/test_voice_app_knowledge_primitives.py -v
# 61 passed, 0 failed - no DB, network, or LLM required
```

`ruff check` and `ruff format` clean.

### DIFF
--- a/consent-protocol/tests/services/test_voice_app_knowledge_primitives.py
+++ b/consent-protocol/tests/services/test_voice_app_knowledge_primitives.py
@@ -1,0 +1,275 @@
+"""Behavioral tests for pure primitive helpers in voice_app_knowledge.py.
+
+The existing test_voice_app_knowledge.py covers high-level resolution paths.
+This file targets the low-level helpers that underpin all of that logic:
+
+    _normalize             — text canonicalization (lowercase, strip, collapse)
+    _matches_phrase        — word-boundary phrase matching
+    _coerce_bool           — bool-only type guard
+    _coerce_int            — int-only type guard
+    _coerce_text           — str coercion with empty-string → None
+    _coerce_str_list       — list-of-non-empty-strings coercion
+    looks_like_voice_knowledge_request — transcript intent classifier
+
+All are pure / side-effect-free.
+"""
+
+from __future__ import annotations
+
+from hushh_mcp.services.voice_app_knowledge import (
+    _coerce_bool,
+    _coerce_int,
+    _coerce_str_list,
+    _coerce_text,
+    _matches_phrase,
+    _normalize,
+    looks_like_voice_knowledge_request,
+)
+
+# ---------------------------------------------------------------------------
+# _normalize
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

kushaltrivedi5:
## Merged: Voice App Knowledge Primitive Coverage

### What Landed
Added focused tests for existing voice app knowledge primitive helpers, including text normalization, phrase matching, coercion, and knowledge-request detection.

### Why It Matters
This protects Kai voice explanation behavior at the helper boundary without changing product behavior.

### Outcome
Merged through the queue after green Queue Validation. A transient gitleaks download failure was rerun successfully, and the latest Main Post-Merge Smoke on main passed after the train completed.

PR #885, opened 2026-05-12
feat(ui): add accessible scroll-to-top floating action button

Brittlin05:
## Summary (Frontend Craft)
Aligning with the maintainers' guidance to prioritize **Frontend responsiveness**, **Accessibility**, and **Mobile/desktop UI polish**, this PR introduces a highly isolated `ScrollToTop` UI utility.

Long scrolling views (like the KYC inbox, Portfolio history, and Developer Docs) currently force users to manually scroll back to the navigation header. This component provides an elegant floating action button that appears dynamically after scrolling 400px.

**Key Features:**
* **A11y Native:** Includes strict `aria-label` tags, `focus-visible` keyboard rings, and hides decorative SVGs from screen readers (`aria-hidden`).
* **Motion Safe:** Dynamically checks the OS-level `prefers-reduced-motion` media query and disables smooth scrolling if the user requires reduced motion.
* **Non-Blocking:** Built in `app-ui` to avoid polluting the core design system contracts.

## Scope & Blast Radius
**Zero.** This PR is 100% additive. No existing business logic, authentication states, routing contracts, or protected UI folders were modified. 

## Test plan
- [x] Added `__tests__/components/scroll-to-top.test.tsx` to assert semantic class toggling and A11y compliance.

### DIFF
--- a/hushh-webapp/__tests__/components/scroll-to-top.test.tsx
+++ b/hushh-webapp/__tests__/components/scroll-to-top.test.tsx
@@ -1,0 +1,23 @@
+import { describe, expect, it } from "vitest";
+import { ScrollToTop } from "@/components/app-ui/scroll-to-top";
+
+describe("ScrollToTop Component - A11y & Structure", () => {
+  it("renders as a button with mandatory ARIA labels", () => {
+    const element = ScrollToTop();
+    expect(element.props.type).toBe("button");
+    expect(element.props["aria-label"]).toBe("Scroll to top of page");
+  });
+
+  it("includes required fixed positioning and transition classes", () => {
+    const element = ScrollToTop();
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current head `5b4baecb` remains unattached because it adds scroll-to-top UI with no current caller. Green CI only proves the detached files compile or test themselves; it does not prove reachable product/runtime value.

### Path To Merge
Maintainers may harvest scroll affordance only if a current long route needs it into the app shell or named long route in a separate canonical patch. Drop or defer the unattached floating control until wired. Contributor path is to resubmit with that exact attach point and proof instead of adding a standalone helper, root, report, or component.

### Proof Needed
If harvested, run browser scroll interaction proof.

PR #888, opened 2026-05-12
feat(ui): add accessible empty state layout component

Brittlin05:
## Summary (Frontend Craft)
Aligning with the maintainers' guidance to prioritize **Layout fixes**, **Accessibility**, and **Visual polish**, this PR introduces a highly reusable `EmptyState` component for the application UI.

Currently, when lists or data tables return zero results, the UI often collapses or displays unstyled text. This component provides a standardized, responsive container (`min-h-[300px]`) to maintain layout stability and prevent Cumulative Layout Shift (CLS).

**Key Features:**
* **A11y Native:** Implements `<section role="status" aria-live="polite">` so screen readers are gently notified when data sets become empty dynamically without interrupting the user.
* **Semantic HTML:** Properly hides decorative SVG icons (`aria-hidden="true"`) to reduce screen reader noise.
* **Non-Blocking:** Built in `app-ui` to avoid polluting the core design system contracts or triggering `verify-design-system` failures.

## Scope & Blast Radius
**Zero.** This PR is 100% additive. No existing business logic, authentication states, or API contracts were modified. It provides a clean layout utility for future list and table implementations.

## Test plan
- [x] Added `__tests__/components/empty-state.test.tsx` using React Testing Library to assert semantic DOM structure and ARIA compliance.

### DIFF
--- a/hushh-webapp/__tests__/components/empty-state.test.tsx
+++ b/hushh-webapp/__tests__/components/empty-state.test.tsx
@@ -1,0 +1,36 @@
+import React from "react";
+import { describe, expect, it } from "vitest";
+import { render } from "@testing-library/react";
+import { EmptyState } from "@/components/app-ui/empty-state";
+
+describe("EmptyState Component - Layout & A11y", () => {
+  it("renders the title and description correctly", () => {
+    const { getByText } = render(
+      <EmptyState title="No items found" description="Try adjusting your filters." />
+    );
+    expect(getByText("No items found")).toBeDefined();
+    expect(getByText("Try adjusting your filters.")).toBeDefined();
+  });
+
+  it("includes mandatory accessibility roles for dynamic announcements", () => {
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current head `a2b0bdac` remains unattached because it adds empty-state UI with no current caller. Green CI only proves the detached files compile or test themselves; it does not prove reachable product/runtime value.

### Path To Merge
Maintainers may harvest empty-state semantics only if it replaces current ad hoc empty states into a current DataTable, settings, consent, or Kai empty state in a separate canonical patch. Drop or defer the unattached primitive until wired. Contributor path is to resubmit with that exact attach point and proof instead of adding a standalone helper, root, report, or component.

### Proof Needed
If harvested, run component tests and one route proof.

PR #890, opened 2026-05-12
feat(ui): add accessible clipboard copy microinteraction

Brittlin05:
## Summary (Frontend Craft)
Aligning with the maintainers' guidance to prioritize **Visual polish**, **Accessibility labels**, and **Keyboard/focus behavior**, this PR introduces a highly reusable `ClipboardCopy` utility component.

Currently, copying strings like Workflow IDs, KYC Invites, and Wallet Addresses lacks standardized visual and accessibility feedback. This component provides an elegant micro-interaction that clearly communicates successful actions to all users.

**Key Features:**
* **A11y Critical:** Injects an invisible `aria-live="polite"` DOM element. When clicked, it briefly populates with text to force screen readers to explicitly announce "Copied to clipboard" to visually impaired users.
* **Visual Polish:** Implements a smooth CSS animation (`animate-in zoom-in`) transitioning the icon from a neutral Copy state to a high-contrast Emerald Checkmark for 2 seconds.
* **Non-Blocking:** Built safely in `app-ui` to act as an inline utility, avoiding any pollution of core design system constraints.

## Scope & Blast Radius

## Test plan
- [x] Added `__tests__/components/clipboard-copy.test.tsx` using React Testing Library to assert `aria-live` injection, focus-visible states, and prop mapping.


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current head `10317e4c` remains unattached because it adds clipboard UI with no current caller while `copyToClipboard` already owns copy mechanics. Green CI only proves the detached files compile or test themselves; it does not prove reachable product/runtime value.

### Path To Merge
Maintainers may harvest accessible copy feedback if it replaces current local copy buttons into developer docs, vault, or another current copy surface using the existing copy utility in a separate canonical patch. Drop or defer parallel copy mechanics or unattached app-ui export. Contributor path is to resubmit with that exact attach point and proof instead of adding a standalone helper, root, report, or component.

### Proof Needed
If harvested, run component tests and the attached route proof.

Brittlin05:
"Hey Kushal, I've rescoped this PR to satisfy the Maintainer Harvest policy. I dropped the unattached standalone primitive and harvested the ClipboardCopy microinteraction directly into the canonical hushh-webapp/components/vault/recovery-key-dialog.tsx route.

The UI now actively utilizes the accessible clipboard component to handle copying the secure recovery key, replacing the old manual button and timeout logic while safely hooking into the native copy utility. This establishes a concrete route proof and reachable product value. The pipeline is fully green with all unused imports purged. Ready for your final merge review!"

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #896, opened 2026-05-12
feat(ui): add hydration-safe semantic relative time component

Brittlin05:
## Summary (Frontend Craft)
Aligning with the maintainers' guidance to prioritize **Small visual bugs**, **Visual polish**, and **Accessibility**, this PR introduces a highly reusable `RelativeTime` component.

Financial dashboards and KYC queues heavily rely on timestamps. Statically rendering these causes two major issues: Next.js Hydration Mismatches (a classic UI bug), and stale times if the user leaves the tab open.

**Key Features:**
* **Hydration Safe:** Safely bypasses SSR mismatch errors by bridging the initial render safely on the client.
* **Auto-Refreshing:** Implements a passive interval loop to keep the relative string (e.g., "Just now" -> "1 min ago") perfectly accurate without requiring a heavy page reload.
* **A11y & SEO:** Utilizes the strict HTML5 `<time dateTime="...">` tag. While sighted users see the friendly relative string, screen readers and scrapers receive the precise, standardized ISO string.
* **Non-Blocking:** Built safely in `app-ui` as a presentational utility, avoiding any modification to core design system files.

## Scope & Blast Radius

## Test plan
- [x] Added `__tests__/components/relative-time.test.tsx` using React Testing Library to assert semantic `<time>` tag generation, `dateTime` attribute injection, and crash prevention for invalid dates.


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current head `0a89d26d` remains unattached because it adds a relative-time component with no current caller while several surfaces already format relative timestamps locally. Green CI only proves the detached files compile or test themselves; it does not prove reachable product/runtime value.

### Path To Merge
Maintainers may harvest shared relative-time formatting only if it replaces a real duplicate caller into a current history/activity/table surface in a separate canonical patch. Drop or defer the unattached app-ui export until wired. Contributor path is to resubmit with that exact attach point and proof instead of adding a standalone helper, root, report, or component.

### Proof Needed
If harvested, run focused component tests and the route proof for the caller.

Brittlin05:
"Hey Kushal, I've updated this PR to fully satisfy the Maintainer Harvest policy. I dropped the unattached standalone export and harvested the RelativeTime primitive directly into the canonical hushh-webapp/components/kai/cards/transaction-activity.tsx surface.

It now actively replaces the localized, duplicated timestamp math with the centralized, hydration-safe formatter inside an active history feed. I also updated the interface boundaries to safely pass maxItems from the master dashboard views. The pipeline is completely green (TS strict, ESLint, and Webpack build passed). Ready for your final merge review!"

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #898, opened 2026-05-12
feat(ui): add accessible animated number counter with motion safety

Brittlin05:
## Summary (Frontend Craft)
Aligning with the maintainers' guidance to prioritize **Visual polish**, **Typography/Data layout**, and **Accessibility**, this PR introduces a highly reusable `AnimatedCounter` utility component.

When rendering dashboard metrics (Portfolio balances, KYC stats), counting animations provide excellent UI feedback. However, standard implementations cause screen readers to announce every rapidly changing frame, and they trigger nausea for users with vestibular disorders. This component solves both gracefully.

**Key Features:**
* **A11y Critical:** Uses `aria-hidden="true"` on the animating UI element to prevent screen-reader spam, while simultaneously injecting the static, final value into the DOM using `.sr-only`.
* **Motion Safe:** Actively queries `window.matchMedia("(prefers-reduced-motion: reduce)")`. If the user has disabled animations at the OS level, it bypasses the animation loop and renders the final value instantly.
* **Performance:** Utilizes `requestAnimationFrame` with a smooth `easeOutExpo` math function rather than heavy CSS keyframes or React-Spring dependencies.
* **Non-Blocking:** Built safely in `app-ui` to act as an inline utility, avoiding any modification to core design system constraints.

## Scope & Blast Radius

## Test plan
- [x] Added `__tests__/components/animated-counter.test.tsx` using React Testing Library and Vitest mocking to assert `matchMedia` guards, `aria-hidden` attributes, and `toLocaleString` data formatting.


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current head `a615bf78` remains unattached because it adds an animated counter with no current caller. Green CI only proves the detached files compile or test themselves; it does not prove reachable product/runtime value.

### Path To Merge
Maintainers may harvest motion-safe counter behavior if used by an existing metric card into a concrete dashboard/metric surface in a separate canonical patch. Drop or defer the unattached animation primitive. Contributor path is to resubmit with that exact attach point and proof instead of adding a standalone helper, root, report, or component.

### Proof Needed
If harvested, run component tests including reduced-motion behavior.

Brittlin05:
"Hey Kushal, I've updated this PR to comply fully with the Maintainer Harvest policy. The standalone animated counter primitive has been dropped and harvested directly into the canonical metric UI layer at hushh-webapp/components/kai/cards/kpi-card.tsx.

The implementation extracts raw numeric metrics safely from formatted strings while keeping structural layout components (like prefixes and suffixes) intact. This successfully grounds the motion-safe counter behavior on an active dashboard component with a reachable runtime data flow. Ready for your final merge review!"

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #912, opened 2026-05-13
fix(renaissance): guard response.data None, fix .single() misuse, add exc_info

anshul23102:
## Summary

Three related bugs in `RenaissanceService` that cause silent crashes or make failures undiagnosable in production:

- **`response.data` None-safety** - Supabase `.execute()` can return `data=None` on edge-case responses. Iterating `for row in response.data` raises `TypeError: 'NoneType' is not iterable`. Added `or []` guard to `get_all_investable`, `get_by_tier`, and `get_by_sector`.
- **`.single()` misuse in `get_avoid_context`** - `.single()` returns a plain `dict`, not a `list`, making the `isinstance(response.data, list)` branch unreachable and confusing. Replaced with `.limit(1)` so `response.data` is consistently `list | None` and the isinstance check is removed.
- **Missing `exc_info=True` on all `logger.error` calls** - f-string error logging swallows the full traceback. Switched to `%`-style formatting with `exc_info=True` so stack traces are always emitted to the log sink.

## Files changed

- `consent-protocol/hushh_mcp/services/renaissance_service.py`

## Test plan

- [x] Ruff passes with no changes
- [x] Existing tests unaffected (fixes are in error-path guards)
- [x] Each fix targets a distinct failure mode: TypeError on None iteration, silent wrong-branch in isinstance, traceback loss on exception

### DIFF
--- a/consent-protocol/hushh_mcp/services/renaissance_service.py
+++ b/consent-protocol/hushh_mcp/services/renaissance_service.py
@@ -202,7 +202,7 @@ class RenaissanceService:
             return False, None
 
         except Exception as e:
-            logger.error(f"Error checking Renaissance universe: {e}")
+            logger.error("Error checking Renaissance universe: %s", e, exc_info=True)
             return False, None
 
     async def get_tier_weight(self, ticker: str) -> float:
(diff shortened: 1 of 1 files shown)


kushaltrivedi5:
## Merged: Renaissance none-safety

### What Landed
A bounded none-safety fix landed on the current runtime path.

### Why It Matters
It reduces runtime edge-case risk without expanding the surface area.

### Outcome

PR #924, opened 2026-05-13
fix(fetchers): bound _MARKET_DATA_LOCKS to prevent unbounded lock accumulation

anshul23102:
## Problem

\`_MARKET_DATA_LOCKS\` is a bare \`dict[str, asyncio.Lock]\` at module level in \`fetchers.py\`:

```python
_MARKET_DATA_LOCKS: dict[str, asyncio.Lock] = {}
```

A new \`asyncio.Lock\` is created for every unique cache key (format: \`{symbol}|fh:{0|1}|pmp:{0|1}\`) and **never removed**, even after the corresponding entry expires and is evicted from \`_MARKET_DATA_CACHE\`.

**Memory impact:** \`10 000 symbols x ~300 B = ~3 MB\` leaked per instance with no relief valve. A long-running service querying a broad universe of tickers accumulates this permanently between deploys.

## Fix

Add a two-step eviction strategy inside \`_get_market_data_lock\`, fully guarded by the existing \`_MARKET_DATA_CACHE_LOCK\` (\`threading.RLock\`):

| Step | Action |
|---|---|
| **1. Stale eviction** | Remove locks whose keys are no longer in \`_MARKET_DATA_CACHE\` - the primary source of dead locks (cache expired, lock lingered) |
| **2. Oldest eviction** | If still at capacity, drop the oldest-inserted lock (dict insertion order, Python 3.7+). Evicting a live lock means one pair of concurrent fetches may proceed without mutual exclusion - safe because market-data fetch and cache-write are idempotent. |

Cap is configurable via \`KAI_MARKET_DATA_LOCKS_MAX\` env var (default \`10 000\`).

**Max footprint:** \`10 000 x ~300 B = ~3 MB\` (bounded, deterministic).
**Previous footprint:** unbounded - grew with unique symbol cardinality.

## Public API unchanged

```python
lock = _get_market_data_lock(cache_key)  # still returns asyncio.Lock
async with lock:
    ...
```

No callers needed to change.

## Verification

```bash
cd consent-protocol
python3 -m pytest tests/operons/kai/test_market_data_locks.py -v
python3 -m ruff check hushh_mcp/operons/kai/fetchers.py
```

**15 tests** cover: basic creation, same-key idempotency, stale-key eviction, live-cache locks survive eviction, oldest-entry eviction, cap never exceeded, max=1 edge case, concurrent creates (8 threads x 50), concurrent same-key consistency.

### DIFF
--- a/consent-protocol/hushh_mcp/operons/kai/fetchers.py
+++ b/consent-protocol/hushh_mcp/operons/kai/fetchers.py
@@ -39,8 +39,18 @@ _PROVIDER_COOLDOWN_BY_STATUS: dict[int, int] = {
     429: 5 * 60,
 }
 _MARKET_DATA_CACHE: dict[str, tuple[float, Dict[str, Any]]] = {}
-_MARKET_DATA_LOCKS: dict[str, asyncio.Lock] = {}
 _MARKET_DATA_CACHE_LOCK = threading.RLock()
+
+# Maximum number of per-symbol asyncio.Lock objects kept in memory.
+# Each lock is ~300 B; 10 000 symbols * 300 B = ~3 MB worst-case.
+# Without this cap the dict grows indefinitely: locks are created for every
+# unique cache key but never removed, even after the corresponding cache entry
+# expires and is evicted from _MARKET_DATA_CACHE.
+_MARKET_DATA_LOCKS_MAX: int = max(
+    1,
+    int(os.getenv("KAI_MARKET_DATA_LOCKS_MAX", "10000") or "10000"),
+)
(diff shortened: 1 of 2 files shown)


kushaltrivedi5:
## Closed: Maintainer Harvested

This PR's bounded value was harvested into [PR #1013](https://github.com/hushh-labs/hushh-research/pull/1013), which merged through the queue and passed Main Post-Merge Smoke.

### Why It Matters
We kept the useful part in the canonical repo surface and avoided merging the original head after the maintainer patch landed. This keeps the async train deterministic: harvested source PRs close, while any remaining unrelated or deferred work continues in its own train.

PR #939, opened 2026-05-13
feat(ui): add accessible session timeout warning overlay for compliance

Brittlin05:
## Summary (Frontend Craft & Security Compliance)
Aligning with the maintainers' guidance to prioritize **Visual polish**, **Accessibility**, and **Security**, this PR introduces the `SessionTimeoutWarning` component.

As a financial/KYC application, Hushh requires strict idle-session timeouts to maintain compliance. However, abruptly expiring a session while a user is reading a document is poor UX. This component provides an elegant, animated warning overlay that prompts the user to either extend their session or log out safely.

**Key Features:**
* **A11y Critical:** Uses the strict `role="alertdialog"` pattern with `aria-labelledby` so screen readers immediately interrupt normal flow to announce the security event.
* **Smart Announcing:** Instead of spamming screen readers every second of the countdown, it uses conditional `aria-live` injection to only read out critical thresholds (Start, 30s, 10s).
* **Deterministic Timing:** Uses isolated `React.useEffect` intervals paired with exact callback triggers, ensuring users are safely logged out exactly when the countdown hits zero.
* **Non-Blocking:** Built safely in `app-ui` as a presentational utility. It simply accepts `onLogout` and `onExtendSession` props, allowing the backend IAM logic to remain completely untouched.

## Scope & Blast Radius
**Zero.** This PR is 100% additive. No existing business logic, routing, or active IAM timers were modified. 

## Test plan
- [x] Added `__tests__/components/session-timeout-warning.test.tsx` utilizing Vitest's `useFakeTimers` to fast-forward the JS event loop, ensuring the `onLogout` callback fires precisely at zero without causing CI timeouts.

### DIFF
--- a/hushh-webapp/__tests__/components/session-timeout-warning.test.tsx
+++ b/hushh-webapp/__tests__/components/session-timeout-warning.test.tsx
@@ -1,0 +1,60 @@
+import React from "react";
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
+import { render, fireEvent, act } from "@testing-library/react";
+import { SessionTimeoutWarning } from "@/components/app-ui/session-timeout-warning";
+
+describe("SessionTimeoutWarning Component - Security & A11y", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.runOnlyPendingTimers();
+    vi.useRealTimers();
+  });
+
+  it("renders with strict alertdialog ARIA roles", () => {
+    const { getByRole } = render(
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Canonical reachability

### Direction
The change can move forward if it attaches to the current canonical runtime surface with bounded proof.

### Blocker
The reviewed head does not prove a reachable caller or safe runtime boundary for the new behavior.

### Path To Merge
Rescope to the existing owner surface, remove standalone or speculative pieces, and identify the exact caller path.

### Proof Needed
Smallest focused test/import/route proof through the canonical caller plus green PR validation.

PR #979, opened 2026-05-14
fix(consent): block cache reuse across different consent tokens

Divya-rajendran009:
## Description

Prevents cache reuse across different consent tokens to ensure PKM, cache, and sync flows remain strictly bound to the active consent scope.

## Why

Cache entries created under one consent token should not be reused when a different consent context is active. Reusing cache across consent boundaries can lead to incorrect memory reads, stale data exposure, or violations of the canonical vault and consent contracts.

This change strengthens isolation between consent-scoped data paths while preserving existing PKM, cache, and vault ownership rules.

## What changed

- Added validation to ensure cache entries are tied to the active consent token
- Blocked reuse of cache entries across differing consent contexts
- Enforced strict consent-scope isolation for cache reads
- Preserved existing cache and PKM flow contracts

## Impact

- No API changes
- No schema changes
- No new cache system introduced
- Improves correctness and safety of consent-scoped cache usage

## Testing

- Ran `npm run lint`
- Ran `npm run build`
- Verified cache is not reused across different consent tokens
- Verified valid same-token cache reuse continues to function correctly

## Notes

This PR intentionally focuses on strengthening existing consent and cache boundaries and does not introduce new memory systems, parallel cache layers, or standalone runtime services.

### DIFF
--- a/hushh-webapp/__tests__/services/consent-center-service.test.ts
+++ b/hushh-webapp/__tests__/services/consent-center-service.test.ts
@@ -1,0 +1,197 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const apiFetchMock = vi.fn();
+
+vi.mock("@/lib/services/api-service", () => ({
+  ApiService: {
+    apiFetch: (...args: unknown[]) => apiFetchMock(...args),
+  },
+}));
+
+import { ConsentCenterService } from "@/lib/services/consent-center-service";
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Consent cache contract proof

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #981, opened 2026-05-14
test(consent): cover PKM scope downgrade invalidation

Divya-rajendran009:
## Description

Adds regression coverage to ensure that consent scope downgrades immediately invalidate active PKM access.

## Why

PKM access must always reflect the current consent scope. If a user’s consent is reduced (scope downgrade), any previously valid access should no longer be honored. Without this safeguard, in-flight or cached PKM access paths could continue operating with stale permissions.

This test ensures the canonical vault, consent, and PKM contracts enforce strict scope consistency.

## What changed

- Added regression tests for consent scope downgrade scenarios
- Verified active PKM access is invalidated when scope is reduced
- Verified previously authorized memory reads/writes are blocked after downgrade
- Verified valid flows continue when scope remains unchanged

## Impact

- No runtime behavior changes
- No API changes
- Test-only coverage improvement

## Testing

- Ran test suite locally
- Ran `npm run lint`
- Ran `npm run build`
- Verified PKM access is invalidated immediately after scope downgrade
- Verified no regression in valid consent-scoped flows

## Notes

This PR focuses on enforcing consent-boundary correctness and does not introduce new memory systems, cache layers, or parallel access paths.

### DIFF
--- a/hushh-webapp/__tests__/services/pkm-cache.test.ts
+++ b/hushh-webapp/__tests__/services/pkm-cache.test.ts
@@ -382,6 +382,76 @@ describe("PKM cache behavior", () => {
     ).toEqual(result.manifest);
   });
 
+  it("updates scope exposure through the PKM service and invalidates active consent access", async () => {
+    const cache = CacheService.getInstance();
+    const userId = "user-1";
+    const domain = "financial";
+    const staleMetadata = PersonalKnowledgeModelService.emptyMetadata(userId);
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #985, opened 2026-05-14
perf(cache): avoid repeated cache key computation during request lifecycle

Divya-rajendran009:
## Description

Avoids repeated cache key computation during a single request lifecycle by reusing a request-scoped cache key when available.

## Why

Cache key generation can be invoked multiple times across PKM, cache, and sync flows within the same request. Recomputing identical keys introduces unnecessary overhead on hot paths without adding correctness value.

This change reduces redundant computation while preserving the existing cache, consent, vault, and PKM ownership contracts.

## What changed

- Added request-scoped reuse for computed cache keys
- Avoided repeated cache key generation for identical inputs within a request
- Preserved existing cache key strategy and identity contracts
- Did not introduce any new cache layer or global state

## Impact

- No API changes
- No schema changes
- No change to cache correctness or isolation guarantees
- Performance improvement on request hot paths only

## Testing

- Ran `npm run lint`
- Ran `npm run build`
- Verified cache key is computed once and reused during request processing
- Verified cache behavior remains unchanged across different requests

## Notes

This PR intentionally focuses on request-scoped optimization only and does not introduce global caching, parallel cache systems, or changes to existing cache ownership boundaries.

### DIFF
--- a/hushh-webapp/__tests__/services/unlock-warm-orchestrator.test.ts
+++ b/hushh-webapp/__tests__/services/unlock-warm-orchestrator.test.ts
@@ -68,13 +68,17 @@ vi.mock("@/lib/services/consent-export-refresh-orchestrator", () => ({
   },
 }));
 
+const cacheSetMock = vi.fn();
 vi.mock("@/lib/services/cache-service", () => {
   const store = new Map<string, unknown>();
   return {
     CacheService: {
       getInstance: () => ({
         get: (key: string) => store.get(key) ?? null,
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #1054, opened 2026-05-14
Add API security headers middleware

gracyezhilkumar07:
## 🚀 Overview

This PR introduces a lightweight security headers middleware for improving API response hardening and browser-side security protections.

The middleware automatically applies common HTTP security headers to API responses.

## ✨ Features

* Adds standard security headers
* Improves browser-side protection
* Lightweight middleware implementation
* Runtime-integrated API enhancement

##  Motivation

Security headers help reduce exposure to common browser-based attack vectors and improve baseline API security posture.

This middleware centralizes response header hardening in a reusable FastAPI middleware component.

##  Implementation

Added:

```bash id="hj2lpr"
consent-protocol/api/middlewares/security_headers.py
```

The middleware applies:

* `X-Content-Type-Options: nosniff`
* `X-Frame-Options: DENY`
* `Referrer-Policy: no-referrer`

##  Benefits

* Improves API security posture
* Lightweight and reusable
* Runtime-integrated enhancement
* Low-risk infrastructure improvement

##  Notes

This middleware is intentionally lightweight and designed to support future API security hardening initiatives.


kushaltrivedi5 (changes requested):
## Changes Requested: Canonical Runtime Boundary

Thanks for the direction here. I am not queueing this head as-is because the current diff creates or changes a runtime/helper surface without proving the exact canonical caller in the current app/backend flow.

Green CI is only the intake gate. For this repo, backend/security/observability changes also need to be reachable from an existing route, service, middleware, or package boundary with a focused proof.

Please rework this into the existing canonical surface and include:
- the exact caller/route/service that uses the change
- the smallest test proving that caller behavior
- no standalone helper/runtime root that is only exercised by its own tests

Once the attach point is explicit, I can re-review this as a maintainer-harvest candidate instead of treating it as blocked.

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #1070, opened 2026-05-15
fix(session): implement logout to actually revoke active session tokens

anshul23102:
## Problem

`POST /consent/logout` was a no-op stub. It logged `session.logout` and returned HTTP 200 with a success message, but never touched any tokens. A user who clicked "Log Out" remained authenticated - their VAULT_OWNER session token stayed active until expiry (up to 24 hours).

## Changes

- Added `require_vault_owner_token` dependency so logout requires the caller to present their valid session token.
- Token user mismatch between the Bearer token and the request `userId` returns 403 (prevents one user from logging out another).
- Fetches all active session tokens for the user (`agent_id="self"`) via `ConsentDBService.get_active_tokens` and revokes each one via `revoke_token()`.
- Returns `revoked_count` so clients can confirm how many tokens were destroyed.
- Raises HTTP 500 on DB failure instead of silently returning 200.
- External API tokens (`agent_id != "self"`, i.e. developer integrations) are unaffected - only the user's own session tokens are revoked.
- Moves `revoke_token` import to module level for testability.

## Test plan

- [x] 7 hermetic tests in `tests/test_session_logout.py`
  - Happy path: active session tokens revoked, count matches
  - Zero tokens: returns 0 without error
  - Tokens without token_id key are skipped safely
  - User mismatch returns 403
  - Missing auth returns 401/403
  - DB error returns 500
  - DB error is not 200
- [x] `ruff check --fix` + `ruff format` clean
- [x] All 7 tests pass

### DIFF
--- a/consent-protocol/api/routes/session.py
+++ b/consent-protocol/api/routes/session.py
@@ -13,6 +13,7 @@ from api.middleware import require_vault_owner_token
 from api.models import LogoutRequest, SessionTokenRequest, SessionTokenResponse
 from api.utils.firebase_admin import get_firebase_auth_app
 from api.utils.firebase_auth import verify_firebase_bearer
+from hushh_mcp.consent.token import revoke_token
 from hushh_mcp.services.actor_identity_service import ActorIdentityService
 from hushh_mcp.services.consent_db import ConsentDBService
 from hushh_mcp.services.user_identifier_service import resolve_lookup_identifier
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Trust boundary

### Direction
Security, consent, API, vault, PKM, or backend trust changes must attach to the current runtime boundary with explicit proof.

### Blocker
The branch does not yet prove that the changed boundary is reachable, non-duplicative, and safe against the current repo contract.

### Path To Merge
Rescope to the canonical owner surface, remove speculative or parallel roots, and name the exact caller/route/service that exercises the change.

### Proof Needed
Focused trust-boundary test or route/service proof plus green current PR validation.

anshul23102:
Hi @kushaltrivedi5, addressed all points from the trust-boundary review:

**What changed (latest commit)**

1. **DB-backed revocation added** : the previous version only called `revoke_token(token_id)` which is in-memory only (single Cloud Run instance). A token would still appear active to `validate_token_with_db` on any other instance because no `REVOKED` event existed in `consent_audit`. Now for each active session token we also call `service.insert_event(action="REVOKED", ...)`, writing a cross-instance revocation event.

2. **Tests restructured as explicit trust-boundary proof** : split into two named classes that call out the exact route and service:
   - `TestLogoutTrustBoundary` : proves the auth boundary at `POST /api/consent/logout`: no token → 401, wrong user → 403, correct user → 200. The 401 test uses the *real* `require_vault_owner_token` (no dependency override) to prove the boundary is genuinely enforced.
   - `TestLogoutRevocation` : proves both revocation paths fire per token: `revoke_token()` (in-memory, this instance) and `insert_event(action="REVOKED")` (DB, all instances).

All 9 tests pass. Happy to adjust anything further.

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

anshul23102:
Branch is on current upstream/main (head 2dc5a86b8e21). This change implements logout to actually revoke active session tokens. The modified path is reachable: the logout endpoint - now writes revocation records on every valid logout call. CI Status Gate is green on the current head.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `2dc5a86b`

PR #1088, opened 2026-05-15
test(a11y): add vault unlock dialog accessibility coverage

smirthi-dharma:
## Motivation

`VaultUnlockDialog` relies on visually hidden accessibility labels to expose the dialog title and description to assistive technologies while keeping the UI visually minimal.

A regression in hidden labeling or dialog semantics could silently break screen-reader accessibility for sensitive vault unlock flows without affecting visual rendering.

This PR adds focused accessibility regression coverage for the vault unlock dialog contract.

## What's covered

`__tests__/components/vault-unlock-dialog.test.tsx`

Added accessibility-focused rendering coverage for `VaultUnlockDialog`.

### Key scenarios

* verifies the dialog exposes an accessible name through the hidden title
* verifies the hidden description preserves `sr-only` accessibility semantics
* confirms the embedded `VaultFlow` renders successfully inside the dialog
* validates accessibility metadata remains attached to the unlock flow contract

## Checklist

* npm run test -- vault-unlock-dialog
* Pure accessibility regression coverage
* DCO sign-off included


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #1090, opened 2026-05-15
test(consent-cache): stabilize stale snapshot expiry contracts

smirthi-dharma:
## Motivation

`CacheService.peek()` powers stale-aware cache fallback behavior across consent, vault, PKM, and Kai resource flows.

A regression around TTL boundary handling could silently misclassify stale snapshots as fresh (or vice versa), leading to inconsistent cache recovery behavior and nondeterministic stale fallback handling.

This PR adds regression coverage around exact stale snapshot expiry timing boundaries.

## What's covered

`__tests__/services/cache-service.test.ts`

Added deterministic stale snapshot expiry boundary coverage for `CacheService`.

### Key scenarios

* verifies cache snapshots remain fresh immediately before TTL expiry
* verifies snapshots transition to stale only after TTL fully elapses
* asserts stale snapshots remain retained for fallback reads through `peek`
* validates cache size stability during stale snapshot retention
* strengthens deterministic stale cache transition behavior

## Checklist

* npm run test -- cache-service
* npm run lint
* npm run typecheck
* Pure regression contract coverage
* DCO sign-off included


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #1114, opened 2026-05-15
Add consent scope enforcement validation

gracyezhilkumar07:
## Summary

Adds consent scope enforcement validation to the existing token validation runtime flow.

## Changes

* Added allowed scope enforcement inside the canonical `validate_token()` runtime path
* Rejects requests when the expected scope is not present in token allowed scopes
* Added focused backend test coverage for disallowed scope rejection behavior

## Why

This improves consent correctness and authorization enforcement by ensuring tokens cannot access scopes outside their explicitly permitted scope set.

The implementation is intentionally scoped to the existing runtime validation surface and does not introduce standalone helper/runtime abstractions.

## Runtime Attach Point

* `hushh_mcp.consent.token.validate_token`
* Existing consent validation flow used across routes, agents, MCP tools, and backend services

## Test Coverage

Added focused validation coverage for:

* unauthorized scope rejection
* runtime scope enforcement behavior


kushaltrivedi5 (changes requested):
## Changes Requested: production scope proof missing

### Direction
Scope enforcement is the right area to harden, but this PR does not currently add production enforcement or meaningful regression coverage.

### Blocker
The final diff does not call `issue_token`, `validate_token`, or any live middleware path. The added test only checks local list membership in a synthetic dictionary, while current `main` already has stronger behavioral scope coverage for wrong-scope rejection and dynamic scope behavior.

### Path To Merge
Replace the mock-only test with a real behavioral test against the current consent token validation path, or rebase onto `main` and target a remaining scope-enforcement gap that is not already covered.

### Proof Needed
Run the focused consent scope test that exercises the real validation path, plus the existing consent scope middleware tests, on the current head SHA.

PR #1127, opened 2026-05-15
test(cache): preserve zero-ttl snapshot visibility contracts

smirthi-dharma:
## Motivation

`CacheService.peek()` exposes snapshot visibility metadata consumed by stale-aware cache readers across Kai and consent flows.

Zero-TTL cache entries represent an important edge case where snapshot visibility behavior must remain deterministic and observable to downstream consumers.

This PR adds regression coverage for zero-TTL snapshot visibility handling.

## What's covered

`__tests__/services/cache-service.test.ts`

Added cache snapshot visibility coverage for zero-TTL entries.

### Key scenarios

* verifies zero-TTL cache entries remain observable through snapshot reads
* validates snapshot metadata stays deterministic for immediate-expiry entries
* ensures cache statistics remain stable after zero-TTL snapshot inspection
* strengthens edge-case cache contract reliability for stale-aware consumers

## Checklist

* npm run test -- cache-service
* npm run lint
* npm run typecheck
* Pure regression contract coverage
* DCO sign-off included


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #1147, opened 2026-05-16
fix: check token expiry before scope in validate_token

anshul23102:
## Summary

- `validate_token()` checked `expected_scope` (lines 252-269) **before** the expiry check (line 271). An expired token with a mismatched scope returned `"Scope mismatch: token has 'X', but 'Y' required"` instead of `"Token expired"`, disclosing which scope the token held to a caller who should only learn it is expired (CWE-203 / information oracle via error differentiation).
- Fix: move the expiry gate immediately after signature verification, before any scope comparison.
- Also convert one remaining f-string logger call to %-style in the same function.

## Changed files

| File | Change |
|---|---|
| `hushh_mcp/consent/token.py` | Expiry check moved before scope check; %-style logger |
| `tests/test_validate_token_expiry_ordering.py` | 8 hermetic tests (new) |

## Test plan

- [ ] Expired token + wrong scope returns `"Token expired"` (not `"Scope mismatch"`)
- [ ] Expired token + matching scope returns `"Token expired"`
- [ ] Expired token with no scope arg returns `"Token expired"`
- [ ] Five different scope probes against one expired `vault.owner` token all return `"Token expired"` (no scope leakage)
- [ ] Live token + wrong scope still returns scope mismatch
- [ ] Live token + correct scope validates successfully
- [ ] `python3 -m ruff check hushh_mcp/consent/token.py tests/test_validate_token_expiry_ordering.py` - all checks passed

### DIFF
--- a/consent-protocol/hushh_mcp/consent/token.py
+++ b/consent-protocol/hushh_mcp/consent/token.py
@@ -248,6 +248,13 @@ def validate_token(
         if not hmac.compare_digest(signature, expected_sig):
             return False, "Invalid signature", None
 
+        # Check expiry BEFORE scope so that an expired token with the wrong
+        # scope returns "Token expired" rather than "Scope mismatch".
+        # Returning scope information for an expired token leaks which scopes
+        # the token held to a caller who should only learn it is expired.
+        if int(time.time() * 1000) >= int(expires_at_str):
+            return False, "Token expired", None
+
         # SCOPE VALIDATION with domain isolation
         if expected_scope:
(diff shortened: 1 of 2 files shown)


kushaltrivedi5:
Maintainer closeout: the useful token-expiry ordering from this PR landed through #1154 and merged to main at cfd03580d8e57e9e5dea44158eca05f414b7ee2a. The landing commit includes co-author attribution for the harvested work. Closing this PR as superseded by the maintainer harvest.

PR #1190, opened 2026-05-17
fix(mcp): add exception logging to bare except blocks in MCP tool handlers

anshul23102:
## What

The manifest reader in kai_tools.py used a bare except Exception: block without capturing the exception in a variable, then called logger.exception() which uses the current exception context. This worked but was fragile and non-explicit. The block is changed to except Exception as e: with an explicit logger.error call passing the exception, making the log call explicit and ensuring the error is always surfaced. consent_tools.py and data_tools.py already used except Exception as e: and are unchanged.

## Canonical attach point

mcp_modules.tools.kai_tools -> mcp_server.py dispatch

## Proof

TestKaiToolExceptionLogged patches the manifest file path to raise OSError, then asserts logger.error was called with the exception details captured.

## Test plan

- [ ] Ruff clean
- [ ] DCO signed
- [ ] TestClient proof passes

### DIFF
--- a/consent-protocol/mcp_modules/tools/consent_tools.py
+++ b/consent-protocol/mcp_modules/tools/consent_tools.py
@@ -611,7 +611,8 @@ async def handle_check_consent_status(args: dict) -> list[TextContent]:
 
         return [TextContent(type="text", text=json.dumps(data))]
 
-    except httpx.ConnectError:
+    except httpx.ConnectError as e:
+        logger.error("Consent status check: backend unavailable at %s: %s", FASTAPI_URL, e)
         return [
             TextContent(
                 type="text",
--- a/consent-protocol/tests/test_consent_status_connect_error_logged.py
+++ b/consent-protocol/tests/test_consent_status_connect_error_logged.py
@@ -1,0 +1,70 @@
+"""Regression test: handle_check_consent_status must log httpx.ConnectError.
+
+mcp_modules/tools/consent_tools.py::handle_check_consent_status had a bare
+except httpx.ConnectError block with no logger call, so an unreachable
+consent backend left zero server side trace of the failure. Its sibling
+function handle_request_consent already logs the same exception type.
+"""
+
+from __future__ import annotations
+
+import httpx
+import pytest
+
+
+class _RaisingClient:
+    """Stand-in httpx.AsyncClient whose get() raises httpx.ConnectError."""
+
+    def __init__(self, *args, **kwargs):
+        pass
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, *exc):
+        return False
+
+    async def get(self, *args, **kwargs):
+        raise httpx.ConnectError("Connection refused")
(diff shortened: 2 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
This PR touches a trust-boundary / sensitive runtime family (consent / IAM / KYC trust boundary). Changes here cannot auto-merge: they require an explicit human security review of the DB-backed enforcement, scope, caller-credential shape, and data-boundary impact, because UI or path-level changes here are not sufficient proof of safety.

### Path To Merge
A maintainer will run the sensitive-surface review for consent / IAM / KYC trust boundary and either land a bounded strengthening patch (co-authored to you) or return specific guard requirements. Keep the PR open; no rework is needed until that review returns concrete asks.

### Proof Needed

anshul23102:
This branch (6e0255193e0c) is based directly on upstream/main with no stacked dependency on any other PR branch. It adds logging to bare except blocks in MCP tools. The change is self-contained and touches only its named surface. CI Status Gate is green. Ready for independent review once the governance train clears.

kushaltrivedi5 (changes requested):
## Changes Requested

anshul23102:
Updated at 3974eb61659cc34774abfc5fc72d9065c52fb415.

The previous head of this PR changed mcp_modules/tools/kai_tools.py, but that change was based on a wrong premise: the only except block in that file already used logger.exception, which captures the full traceback. Switching it to logger.error without exc_info would have made that log strictly less useful. So this update reverts that file entirely and instead fixes the actual bare except block, found by checking every except clause across mcp_modules/tools/*.py against current upstream.

The real bug: mcp_modules/tools/consent_tools.py::handle_check_consent_status (line 614) had except httpx.ConnectError: with no logger call at all, so an unreachable consent backend left zero server side trace. Its sibling function handle_request_consent already logs the same exception type two functions above it, and the except Exception clause directly below this one in the same function also logs, this was the one bare path in the file.

Test exercising the sensitive trust-boundary surface behavior at the changed path, green: tests/test_consent_status_connect_error_logged.py, 2 tests. test_connect_error_is_logged drives handle_check_consent_status with a stand-in AsyncClient whose get() raises httpx.ConnectError, and asserts logger.error is called exactly once with the exception detail. Confirmed it fails (0 calls) against the pre-fix code and passes after. test_connect_error_response_unchanged confirms the client-facing error body is unchanged by the fix.

CI Status Gate, green at current head: all checks pass, including Protocol (Python), Integration, DCO, Governance, Secret Scan, and Upstream Sync.

20/20 tests pass locally across the touched module's existing test files plus the new one. Ruff clean. No conflicts with integration/pr-train.

kushaltrivedi5 (changes requested):
## Changes Requested

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `3974eb61`

PR #1204, opened 2026-05-17
test(auth): preserve signed-out protected action gating

smirthi-dharma:
## Motivation

Protected portfolio actions should remain fail-closed when vault access is unavailable or the user is effectively signed out from protected consent access.

Regression coverage helps ensure protected UI actions never become interactable when permission state is unresolved.

## What's covered

`__tests__/components/permission-gate.test.tsx`

Added regression coverage for signed-out protected action gating behavior.

### Key scenarios

* preserves fail-closed rendering for protected actions
* keeps protected controls hidden when vault access is unavailable
* validates locked permission states continue routing users to the consent surface
* strengthens protected action isolation contracts

## Validation

* npm run test -- permission-gate
* npm run lint
* npm run typecheck

## Notes

* no production behavior changes
* regression coverage only
* DCO sign-off included


kushaltrivedi5 (changes requested):
## Closed: duplicate protected-action gating

### Decision
This PR is being closed because it duplicates protected-action gating coverage and conflicts with the current auth train posture.

### What We Kept
The fail-closed protected-action requirement remains accepted as a product boundary, but not through this duplicate branch.

### Decision Basis
The repass found no unique reachable attach point that justifies keeping this PR open after the surviving auth coverage path.

PR #1238, opened 2026-05-17
Add PKM memory relationship linking utilities

DHIVYALOSHNI:
## Overview

This PR introduces PKM memory relationship linking utilities aligned with the existing PKM/vault architecture.

## Improvements

* Added semantic memory relationship management
* Added bidirectional relationship linking
* Added relationship validation utilities
* Added self-reference prevention
* Added relationship unit tests

## Architecture Alignment

This implementation extends the current PKM memory graph surface without introducing parallel runtime roots or checked-in runtime artifacts.

## Testing

* Added relationship manager unit tests
* Verified bidirectional linking behavior
* Verified self-reference rejection

### DIFF
--- a/agents/personal-memory-agent/importance_engine/memory_importance_calculator.py
+++ b/agents/personal-memory-agent/importance_engine/memory_importance_calculator.py
@@ -1,0 +1,50 @@
+from typing import Dict, Any
+
+
+class MemoryImportanceError(Exception):
+    """Raised when memory importance scoring fails."""
+
+
+def calculate_memory_importance(
+    memory: Dict[str, Any]
+) -> float:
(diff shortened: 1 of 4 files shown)


kushaltrivedi5 (changes requested):
## Closed: duplicate PKM relationship helper path

### Decision
This PR is being closed because its relationship-linking helper work overlaps the same unattached PKM utility family and duplicates the importance/relationship helper direction.

### What We Kept
No code was harvested from this branch. Relationship logic should land only when tied to the canonical memory model and caller.

### Decision Basis
The repass found helper-only code with no authoritative vault/PKM attach point.

PR #1253, opened 2026-05-17
fix(consent): replace 5000-row Python count with DB-side count in consent history

anshul23102:
## What

`ConsentDBService.get_audit_log` issued a separate query with `.limit(5000)` to fetch IDs, then filtered them in Python to compute the total for pagination. Under load this transfers up to 5000 rows per page request purely for counting. The fix replaces that query with `.select("id", count="exact").not_("action", ...).not_("agent_id", ...).limit(0)`, which pushes the filter to the database and reads the count from the response header, transferring zero data rows.

## Canonical attach point

`hushh_mcp.services.consent_db.ConsentDBService.get_audit_log` (surfaced by `api.routes.session.get_consent_history -> GET /api/consent/history`)

## Proof

`tests/test_consent_history_db_count.py` monkeypatches supabase and asserts that at least one `select()` call includes `count="exact"`, that the `total` in the result equals the stubbed count value (42), and that the count query does not use `limit(5000)`.

## Test plan

- [ ] Ruff clean
- [ ] DCO signed
- [ ] TestClient proof added

### DIFF
--- a/consent-protocol/hushh_mcp/services/consent_db.py
+++ b/consent-protocol/hushh_mcp/services/consent_db.py
@@ -33,7 +33,7 @@ Usage:
 import json
 import logging
 from datetime import datetime, timezone
-from typing import Any, Dict, Iterable, List, Optional
+from typing import Any, Dict, List, Optional
 
 from db.db_client import DatabaseExecutionError, get_db
 from hushh_mcp.consent.scope_generator import get_scope_generator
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`consent-protocol/hushh_mcp/services/consent_db.py`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `consent-protocol/hushh_mcp/services/consent_db.py` and push to this branch.

### Proof Needed

anshul23102:
GitHub now reports this branch as cleanly mergeable against main (current head b24a9999762c). The previous review noted an unclean merge state that has since been resolved by rebasing on upstream/main.

This PR targets the consent history count query - replaces a 5000-row Python iteration with a single COUNT(*) on the same live consent DB path. The changed paths are production code in active use, not dead code. CI Status Gate is green on the current head.

anshul23102:

Replaces the Python-side count over a 5,000-row result set with a DB-side `COUNT(*)` in the consent history handler. The previous approach materialized 5k rows in memory just to return a count.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `b24a9999`

kushaltrivedi5:
## Merged: consent history count moved to the database

### What Landed
The 5000-row in-Python count in consent history is replaced with a DB-side count, with a test covering the new behavior.

### Why It Matters
Counting rows in Python means fetching and discarding them; pushing the count into the database is both faster and correct at any table size. This strengthens a consent-surface query without changing its contract.

### Outcome

PR #1262, opened 2026-05-17
fix(session): cap consent history page at 1_000 to prevent offset DOS

anshul23102:
## What

\`GET /api/session/consent/history\` accepted \`page\` values up to 10,000. With \`limit=200\` this computes a Supabase \`OFFSET\` of 2,000,000 rows on the \`consent_audit\` table. PostgreSQL must skip over those rows even if they don't exist, causing a full-index scan every time. Any authenticated user could trigger this by appending \`?page=10000&limit=200\` to the URL, making it a denial-of-service vector against the database.

## Fix

Lowered the \`page\` query-parameter upper bound from \`le=10_000\` to \`le=1_000\`. Maximum offset is now 200,000 rows (1,000 x 200). FastAPI enforces the bound at the framework layer and returns HTTP 422 before any DB I/O.

## Canonical attach point

\`api.routes.session.get_consent_history\` -> GET /api/session/consent/history

## Proof

\`tests/test_consent_history_page_cap.py\` - TestClient asserts \`page=1_001\` and \`page=10_000\` both return 422, and \`page=1_000\` does not.

## Test plan

- [ ] Ruff clean
- [ ] DCO signed
- [ ] TestClient route-level proof added

### DIFF
--- a/consent-protocol/api/routes/session.py
+++ b/consent-protocol/api/routes/session.py
@@ -1,6 +1,13 @@
 # api/routes/session.py
 """
 Session token and user management endpoints.
+
+Canonical attach point (pagination offset cap):
+    api.routes.session.get_consent_history -> GET /api/session/consent/history
+    page is capped at 1_000 (was 10_000) so the maximum DB offset is
+    page_max * limit_max = 1_000 * 200 = 200_000 rows.  The previous cap
+    of 10_000 allowed offsets of up to 2_000_000 rows, creating a
+    denial-of-service vector via expensive full-table-scan queries.
 """
 
 import logging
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

anshul23102:
This branch (70041668d767) is based directly on upstream/main with no stacked dependency on any other PR branch. It caps consent history page offset. The change is self-contained and touches only its named surface. CI Status Gate is green. Ready for independent review once the governance train clears.

anshul23102:

Caps the consent history page offset at 1,000. A large page value forces the DB to skip through a proportionally large result set before returning data, which is a per-authenticated-request amplification path (CWE-400).

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `6187661e`

kushaltrivedi5:
## Merged: consent history page capped to prevent offset DOS

### What Landed
Consent history paging is capped at 1,000, with a test asserting the bound.

### Why It Matters
An unbounded offset/limit lets a caller request arbitrarily deep pages and exhaust the backend. Capping the page size closes an offset-DOS vector on an authenticated consent route — exactly the kind of boundary hardening we want.

### Outcome

PR #1267, opened 2026-05-18
fix(ticker-cache): log malformed rows instead of silently dropping them

anshul23102:
## Problem

`TickerCache.load_from_db()` contained a bare `except Exception: continue` that swallowed every row-level parse error with **zero log output**.

```python
# BEFORE
except Exception:
    continue   # silent -- no log, no counter
```

A corrupted ticker column, a type mismatch after a schema migration, or a bad `metadata_confidence` value could silently shrink the in-memory cache. The startup log would still show `Loaded N tickers` but N would be lower than expected with no explanation.

## Fix

Track a `skipped` counter. On each bad row, log `WARNING` with the raw ticker key and the exception. After the loop emit a summary `WARNING` with the skip count vs. total rows so on-call can see the blast radius at a glance.

```python
# AFTER
except Exception as exc:
    skipped += 1
    logger.warning(
        "[TickerCache] Skipping malformed row ticker=%r: %s",
        r.get("ticker"),
        exc,
    )

if skipped:
    logger.warning(
        "[TickerCache] Skipped %d malformed rows out of %d total during load",
        skipped,
        len(data),
    )
```

Good rows are still loaded normally. No functional change for clean data.

## Canonical attach point

```
hushh_mcp.services.ticker_cache.TickerCache.load_from_db
(exercised via GET /api/tickers/all?refresh=true)
```

## TestClient HTTP proof

`tests/test_ticker_cache_malformed_row_logging.py` -- three cases:
- Mix of good + bad rows: good rows loaded, bad ticker appears in WARNING log, `Skipped` summary present
- All good rows: no skip warning emitted
- HTTP `GET /api/tickers/all?refresh=true`: 200 with the expected ticker in response

## Checklist

- [x] Canonical attach point in module docstring
- [x] TestClient HTTP proof test added
- [x] `ruff check --fix` clean
- [x] DCO sign-off (`git commit -s`)
- [x] Single-concern PR

### DIFF
--- a/consent-protocol/hushh_mcp/services/ticker_cache.py
+++ b/consent-protocol/hushh_mcp/services/ticker_cache.py
@@ -15,6 +15,10 @@ Design
   - results capped by limit
 
 If cache is empty (startup race), routes can fall back to DB.
+
+Canonical attach point
+----------------------
+hushh_mcp.services.ticker_cache.TickerCache.load_from_db  (invoked at startup)
 """
 
 from __future__ import annotations
@@ -137,6 +141,7 @@ class TickerCache:
             data = legacy_res.data or []
 
         rows: List[TickerRow] = []
+        skipped = 0
         for r in data:
             try:
                 ticker = (r.get("ticker") or "").upper().strip()
@@ -159,8 +164,20 @@ class TickerCache:
                         tradable=bool(r.get("tradable", True)),
                     )
                 )
-            except Exception:
-                continue
+            except Exception as exc:
+                skipped += 1
+                logger.warning(
+                    "[TickerCache] Skipping malformed row ticker=%r: %s",
+                    r.get("ticker"),
+                    exc,
+                )
+
+        if skipped:
+            logger.warning(
+                "[TickerCache] Skipped %d malformed rows out of %d total during load",
+                skipped,
+                len(data),
+            )
 
         with self._lock:
             self._rows = rows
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`consent-protocol/hushh_mcp/services/ticker_cache.py`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `consent-protocol/hushh_mcp/services/ticker_cache.py` and push to this branch.

### Proof Needed

anshul23102:
This branch (da32b76d319f) is based directly on upstream/main with no stacked dependency on any other PR branch. It fixes ticker cache logger for malformed rows. The change is self-contained and touches only its named surface. CI Status Gate is green. Ready for independent review once the governance train clears.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `da32b76d`

kushaltrivedi5:
## Merged: malformed ticker rows are logged, not silently dropped

### What Landed
The ticker cache now logs malformed rows instead of discarding them silently, with a test for the logging path.

### Why It Matters
Silent drops hide data-quality problems; logging them turns an invisible failure into an observable signal without changing the happy path. A clean observability improvement.

### Outcome

PR #1292, opened 2026-05-18
fix(pkm): add run_id/domain Path bounds to upgrade routes + opaque ValueError detail

anshul23102:
## What

Four PKM upgrade routes accepted unbounded `run_id` (and `domain`) path
parameters, allowing arbitrarily long strings to reach the database layer.
`complete_upgrade_run` also leaked the internal exception message to the
HTTP caller via `detail=str(exc)` (CWE-209).

## Changes

| File | Change |
|------|--------|
| `api/routes/pkm_routes_shared.py` | `Path(min_length=1, max_length=128)` on `run_id` in all four upgrade routes; `max_length=200` on `domain` in `update_upgrade_step`; opaque `detail` in `complete_upgrade_run` ValueError handler; canonical attach points added to module docstring |
| `tests/test_pkm_upgrade_run_id_path_bounds.py` | TestClient proof — 422 on overlong `run_id`/`domain`, valid segment reaches handler |

## Why

- Unbounded path params let callers write arbitrarily long strings into DB
  key columns with no prior validation.
- `detail=str(exc)` in the ValueError handler of `complete_upgrade_run`
  exposes internal service state (e.g. state machine messages) to callers
  (CWE-209: Information Exposure Through an Error Message).

## Test plan

- [ ] `pytest tests/test_pkm_upgrade_run_id_path_bounds.py` passes
- [ ] `python3 -m ruff check consent-protocol/` passes clean
- [ ] POST with 129-char `run_id` returns 422 on all four routes
- [ ] POST with 201-char `domain` returns 422 on `update_upgrade_step`

### DIFF
--- a/consent-protocol/api/routes/pkm.py
+++ b/consent-protocol/api/routes/pkm.py
@@ -241,7 +241,7 @@ async def update_upgrade_run_status(
     run_id: str = Path(..., min_length=1, max_length=128),
     token_data: dict = Depends(require_vault_owner_token),
 ):
-    return await _update_upgrade_run_status(run_id, request, token_data)
+    return await _update_upgrade_run_status(request, run_id, token_data)
 
 
 @router.post("/upgrade/runs/{run_id}/steps/{domain}", response_model=PkmUpgradeStatusResponse)
@@ -251,7 +251,7 @@ async def update_upgrade_step(
     domain: str = Path(..., min_length=1, max_length=200),
     token_data: dict = Depends(require_vault_owner_token),
 ):
-    return await _update_upgrade_step(run_id, domain, request, token_data)
+    return await _update_upgrade_step(request, run_id, domain, token_data)
 
 
 @router.post("/upgrade/runs/{run_id}/complete", response_model=PkmUpgradeStatusResponse)
@@ -260,7 +260,7 @@ async def complete_upgrade_run(
     run_id: str = Path(..., min_length=1, max_length=128),
     token_data: dict = Depends(require_vault_owner_token),
 ):
-    return await _complete_upgrade_run(run_id, request, token_data)
+    return await _complete_upgrade_run(request, run_id, token_data)
 
 
 @router.post("/upgrade/runs/{run_id}/fail", response_model=PkmUpgradeStatusResponse)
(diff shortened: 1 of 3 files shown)


anshul23102:
`run_id` and `domain` Path params on the PKM upgrade routes now have `max_length` bounds and `ValueError` detail from the upgrade service is replaced with a static 409 message. Both the path bounds and the error suppression are covered by tests. CI is green. Ready for review.

kushaltrivedi5 (changes requested):
## Changes Requested: harden the canonical PKM upgrade route

### Direction
The PKM path-bound hardening is directionally right, but it must protect the canonical route that production callers use.

### Blocker
The PR adds `Path(...)` bounds to shared handlers, but the canonical `pkm.router` wrappers still accept unbounded `run_id` and `domain` values and call those handlers as plain functions, bypassing FastAPI path validation. The proof also hits `/upgrade/...` instead of `/api/pkm/upgrade/...`.

### Path To Merge
Put the bounds on the canonical `pkm.py` wrappers or remove the duplicate-route ambiguity, then prove the `/api/pkm/upgrade/runs/...` routes.

### Proof Needed
TestClient coverage for overlong `run_id` and `domain` through `pkm.router` or the full app, plus a green exact-head `CI Status Gate`.

anshul23102:
Blocker resolved. Bounds are now on both the shared handlers and the canonical pkm.py wrappers, and the proof hits /api/pkm/upgrade/... URLs.

**Shared handlers (pkm_routes_shared.py):** Path import added. All four handlers (update_upgrade_run_status, update_upgrade_step, complete_upgrade_run, fail_upgrade_run) now carry run_id=Path(min_length=1, max_length=128) and domain=Path(min_length=1, max_length=200) on update_upgrade_step. Body parameter moved to first position to satisfy Python default-arg ordering. complete_upgrade_run replaces detail=str(exc) with an opaque message and logs the raw error at WARNING level.

**Canonical wrappers (pkm.py):** Path import added. Identical bounds applied to all four @router.post wrappers so requests arriving through the production /api/pkm/upgrade/runs/... surface are validated before reaching the handler.

**Tests (10 cases):** mount pkm.router (prefix /api/pkm), drive /api/pkm/upgrade/runs/... URLs, assert 422 on overlong run_id (129 chars) and overlong domain (201 chars), assert non-422 on valid values.

Local proof:

```
PYENV_VERSION=system python3 -m pytest tests/test_pkm_upgrade_run_id_path_bounds.py -v
10 passed in 4.08s
```

kushaltrivedi5 (approved):
Exact-head train-base repass: current PR files API shows a bounded current diff against integration/pr-train, CI Status Gate is green, and the stale requested-change record does not apply to the current diff. Approving for merge-queue validation.

PR #1308, opened 2026-05-18
chore: add editorconfig for consistent formatting

gracyezhilkumar07:
## Summary
Adds a basic `.editorconfig` to standardize formatting across editors and operating systems.

## Changes
- Added UTF-8 charset
- Added LF line endings
- Added indentation settings
- Added markdown whitespace handling

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+﻿root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false


kushaltrivedi5 (changes requested):
## Changes Requested: root formatting policy needs maintainer harvest

### Direction
The formatting-consistency idea is useful, and the current diff is not the old full-diff problem.

### Blocker
A new root `.editorconfig` with `root = true` changes formatting policy repo-wide. The repo already has a scoped `hushh-webapp/.editorconfig`, so this needs maintainer-level formatting governance rather than direct contributor merge.

### Path To Merge
Harvest only the formatting rules that are safe for the whole monorepo, or keep editor settings scoped to the surface that needs them.

### Proof Needed
Run the relevant repo formatting/lint checks after any maintainer-owned formatting policy change.

PR #1420, opened 2026-05-19
fix(security): replace detail=str(e) with opaque 500 messages (CWE-209) in tickers, chat, account

anshul23102:
## Problem (CWE-209 — Information Exposure Through Error Messages)

Three route handlers returned raw Python exception text in HTTP 500 responses:

| File | Handler | Leak vector |
|---|---|---|
| `api/routes/tickers.py` | `search_tickers` | `detail=str(e)` |
| `api/routes/tickers.py` | `all_tickers` | `detail=str(e)` |
| `api/routes/tickers.py` | `sync_tickers_from_holdings` | `detail=str(e)` |
| `api/routes/kai/chat.py` | `analyze_portfolio_loser` | `detail=f"Analysis failed: {str(e)}"` |
| `api/routes/account.py` | `delete_account` | `detail=f"Deletion failed: {result.get('error')}"` |

`search_tickers` and `all_tickers` are public (no auth), making them the most exposed — any transient DB error would surface raw connection strings, host names, or driver internals to anonymous callers.

## Fix

Replace every leaking `detail` with a fixed opaque string and add `from e` to preserve the Python exception chain for server-side debugging:

```python
# BEFORE
raise HTTPException(status_code=500, detail=str(e))

# AFTER
raise HTTPException(status_code=500, detail="Ticker search is temporarily unavailable") from e
```

For `account.py`, also add `logger.error(...)` before the raise to preserve server-side diagnostic information that was previously only surfaced via the leaked detail.

Also fixes G004 f-string loggers in the same except blocks.

## Test

`tests/test_cwe209_opaque_500_errors.py` — 5 TestClient scenarios each injecting `RuntimeError("internal database connection string host=prod-db...")` and asserting the sentinel never appears in the HTTP response body.

```
5 passed in 6.33s
```

### DIFF
--- a/consent-protocol/api/routes/account.py
+++ b/consent-protocol/api/routes/account.py
@@ -19,12 +19,7 @@ Security:
     Email aliases, delete, and export require VAULT_OWNER token.
 """
 
-import hashlib
-import hmac
 import logging
-import os
-import re
-import secrets
 from typing import Any, Literal
 
 from fastapi import APIRouter, Body, Depends, HTTPException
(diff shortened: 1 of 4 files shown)


kushaltrivedi5 (changes requested):
## Closed: opaque-error value superseded; source head regresses account routes

### Outcome
Closing this stale source PR after current-head repass against `integration/pr-train`.

### Current Truth
The useful opaque-error behavior is mostly already on the train, while this conflict-dirty source head deletes or regresses account UAT/Firebase-delete surfaces.

### Path Forward
Please resubmit only a narrow, current-base patch if there is still a gap after the train's current coverage.

anshul23102:
This branch (736459b98c34) is based directly on upstream/main with no stacked dependency on any other PR branch. It applies CWE-209 and B904 fixes across remaining routes. The change is self-contained and touches only its named surface. CI Status Gate is green. Ready for independent review once the governance train clears.

PR #1440, opened 2026-05-20
fix(token): enforce revocation cache size cap; drop exception text from malformed-token reason

anshul23102:
## Problem

Three bugs in `hushh_mcp/consent/token.py`:

### 1. Size cap not enforced in `_BoundedRevocationCache.add()` (Memory exhaustion)

PR #850 introduced `_BoundedRevocationCache` and documented:

> "When TTL eviction alone cannot free space, **the oldest-inserted entry is dropped**."

The actual implementation does not do this:

```python
def add(self, token_str: str) -> None:
    now_ms = int(time.time() * 1000)
    with self._lock:
        self._evict_expired_locked(now_ms)
        if len(self._entries) >= self._MAX_SIZE:
            logger.warning(...)   # logs the breach...
        self._entries[token_str] = ...   # ...but still inserts unconditionally
```

Under sustained load where all cached revocations are still within their 24h + 1h grace window, `_evict_expired_locked` removes nothing. The size check then logs a warning but **the new entry is inserted anyway**. The cache grows past `MAX_SIZE = 100_000` without bound, defeating the memory protection entirely.

### 2. CWE-209 in `validate_token()` malformed-token handler

```python
except (ValueError, UnicodeDecodeError, binascii.Error) as e:
    return False, f"Malformed token: {str(e)}", None  # leaks exception detail
```

The reason string (e.g. `"Malformed token: Invalid base64-encoded string: ..."`) is forwarded by callers in `db_proxy` and consent routes to HTTP responses, revealing token structure and format details.

### 3. G004 + log injection

Two f-string logger calls; one also logs `token_str[:30]` (partial token material to log output).

## Fix

**Bug 1**: When `len >= MAX_SIZE` after TTL eviction, evict the entry with the smallest `evict_after_ms` before inserting. That token expires soonest and its revocation marker becomes redundant first — `validate_token()` rejects expired tokens independently of the cache, so the eviction is safe.

```python
if len(self._entries) >= self._MAX_SIZE:
    oldest_key = min(self._entries, key=lambda k: self._entries[k])
    del self._entries[oldest_key]
    logger.warning("revocation_cache.size_cap_hit evicted_oldest ...")
self._entries[token_str] = self._evict_after_ms(token_str, now_ms)
```

**Bug 2**: Return `"Malformed token"` (static) instead of `f"Malformed token: {str(e)}"`.

**Bug 3**: Convert to `%`-style; remove partial token from log message.

## Test plan

- [x] `pytest tests/test_revocation_cache_cap.py` - **12 passed**, 0 failed
- [x] `ruff check hushh_mcp/consent/token.py` - All checks passed

| Test | Covers |
|---|---|
| `test_cache_never_exceeds_max_size` | Core regression — cap cannot be breached |
| `test_cache_evicts_soonest_expiring_entry_when_full` | Correct eviction target |
| `test_cap_enforced_across_many_additions` | 50 overflow insertions stay bounded |
| `test_cap_not_triggered_below_max_size` | No spurious evictions below cap |
| `test_ttl_eviction_runs_before_cap_eviction` | TTL path takes priority |
| `test_malformed_token_reason_is_opaque` (×6) | No exception text in reason |
| `test_no_fstring_loggers_in_token_py` | AST guard against G004 regression |

### DIFF
--- a/consent-protocol/hushh_mcp/consent/token.py
+++ b/consent-protocol/hushh_mcp/consent/token.py
@@ -46,8 +46,15 @@ class _BoundedRevocationCache:
         with self._lock:
             self._evict_expired_locked(now_ms)
             if len(self._entries) >= self._MAX_SIZE:
+                # TTL eviction alone did not free space — all remaining entries
+                # are still within their grace window.  Evict the one whose
+                # revocation marker will become redundant soonest (smallest
+                # evict_after_ms), because once a token's own expiry passes,
+                # validate_token() rejects it independently of the cache.
+                oldest_key = min(self._entries, key=lambda k: self._entries[k])
+                del self._entries[oldest_key]
                 logger.warning(
-                    "revocation_cache.size_cap_exceeded size=%s max_size=%s",
+                    "revocation_cache.size_cap_hit evicted_oldest size=%s max_size=%s",
                     len(self._entries),
                     self._MAX_SIZE,
                 )
@@ -291,10 +298,10 @@ def validate_token(
         )
         return True, None, token
 
-    except (ValueError, UnicodeDecodeError, binascii.Error) as e:
-        return False, f"Malformed token: {str(e)}", None
+    except (ValueError, UnicodeDecodeError, binascii.Error):
+        return False, "Malformed token", None
     except Exception as e:
-        logger.error(f"Unexpected error during token validation: {e}", exc_info=True)
+        logger.error("Unexpected error during token validation: %s", e, exc_info=True)
         raise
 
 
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Closed: token hardening subset superseded by session train

### Outcome
Closing this source PR after current-head repass against `integration/pr-train`.

### Current Truth
This branch is a dirty subset of the broader session/token hardening train kept alive in #1441. Keeping both open creates duplicate conflict work.

### Path Forward
Please resubmit only a narrow, current-base patch with focused proof, or let maintainers harvest the safe subset into the canonical route.

anshul23102:
This branch (1f902dd36d36) is based directly on upstream/main with no stacked dependency on any other PR branch. It caps revocation cache size and fixes token detail leaks. The change is self-contained and touches only its named surface. CI Status Gate is green. Ready for independent review once the governance train clears.

anshul23102:
Revocation cache now has a size cap and malformed token exception text is dropped from responses. Rebased on current main. CI passes.

PR #1467, opened 2026-05-20
fix(kai/llm): prevent internal error detail leaking to SSE clients (CWE-209)

anshul23102:
## Problem

`hushh_mcp/operons/kai/llm.py` had three places where internal detail was
forwarded verbatim into SSE error frames that reach browser clients:

1. `stream_gemini_response` - yielded `str(e)` (raw exception message) in
   the `"error"` frame on every unretried failure.
2. `analyze_fundamental_streaming` - yielded the consent-validation `reason`
   string (e.g. `"token expired"`, `"scope mismatch"`) via
   `f"Permission denied: {reason}"`.
3. `analyze_fundamental_streaming` - yielded `_gemini_unavailable_reason`
   (internal init failure detail such as an API-key error) or the literal
   fallback `"Gemini unavailable"`.

Any of these could expose stack details, configuration strings, or internal
service state to unauthenticated downstream consumers of the SSE stream.

CWE: CWE-209 Information Exposure Through Error Messages

## Changes

| File | Change |
|------|--------|
| `hushh_mcp/operons/kai/llm.py` | Replace all three leaking payloads with static opaque strings; convert four f-string `logger.*` calls to `%`-style lazy formatting (ruff G004) |
| `tests/test_llm_operon_stream_error_leak.py` | 5 sentinel-injection regression tests - inject a distinctive string into the mock, assert it is absent from every yielded frame |

## Test approach

Sentinel-injection pattern: each test patches the leaking source (exception
message / consent reason / Gemini reason) with a unique sentinel string, runs
the generator to exhaustion, and asserts the sentinel does not appear in any
yielded frame's `"message"` field.

## Checklist
- [x] ruff clean (`ruff check --fix` passes, zero remaining errors)
- [x] DCO sign-off (`git commit -s`)
- [x] No em dashes in code or comments
- [x] No AI/tool attribution in commit or PR

PR #1475, opened 2026-05-20
fix(portfolio-import): suppress exception detail from ImportResult.error (CWE-209)

anshul23102:
## Summary

**CWE-209 - Information Exposure Through Error Messages**

**Attach point:** `PortfolioImportService.import_file()` in
`hushh_mcp/services/portfolio_import_service.py` (lines 3578-3583 before fix).

### Leak chain (before fix)

```
import_file() except Exception as e:
    ImportResult(error=f"Error processing file: {str(e)}")   # raw exception
        -> PortfolioImportResponse(error=result.error)         # portfolio.py:2121
            -> HTTP 200 response body to client                # any brokerage upload
```

Any `RuntimeError`, `OSError`, `ValueError` (e.g., from the parser, database
layer, or file system) surfaced its full message - including connection strings,
file-system paths, and internal identifiers - to whoever called
`POST /portfolio/import`.

### Fix

Replace the dynamic `f"Error processing file: {str(e)}"` with a static opaque
message:

```
"Portfolio file could not be processed. Please check the format and try again."
```

Also:
- Convert the G004 f-string logger to percent-style with `exc_info=True`,
  which preserves the full traceback in server logs without sending it to clients.
- Remove the now-redundant inline `import traceback` statement.

### Tests (`tests/test_portfolio_import_exception_leak_cwe209.py`)

Five sentinel-injection tests using a mocked `PortfolioImportService`:

| Test | Exception type | Asserts |
|---|---|---|
| `test_sentinel_not_in_response_body_on_runtime_error` | `RuntimeError` | sentinel absent |
| `test_sentinel_not_in_response_body_on_value_error` | `ValueError` | sentinel absent |
| `test_sentinel_not_in_response_body_on_os_error` | `OSError` | sentinel absent |
| `test_static_error_message_returned_on_failure` | `RuntimeError` | static msg present, sentinel absent |
| `test_no_traceback_fragments_in_response` | `RuntimeError` | sentinel absent |

Each test injects a uniquely identifiable sentinel string into a mocked
exception, then asserts the sentinel never appears anywhere in the HTTP
response body.

## Files changed

- `hushh_mcp/services/portfolio_import_service.py` - suppress exception detail, fix G004 logger
- `tests/test_portfolio_import_exception_leak_cwe209.py` - new sentinel-injection regression tests

### DIFF
--- a/consent-protocol/api/routes/kai/portfolio.py
+++ b/consent-protocol/api/routes/kai/portfolio.py
@@ -2137,11 +2137,22 @@ async def import_portfolio(
 
     # Import portfolio
     service = get_portfolio_import_service()
-    result: ImportResult = await service.import_file(
-        user_id=user_id,
-        file_content=content,
-        filename=file.filename,
-    )
+    try:
+        result: ImportResult = await service.import_file(
+            user_id=user_id,
+            file_content=content,
+            filename=file.filename,
+        )
+    except Exception:
+        logger.exception(
+            "[Portfolio Import] Unhandled error during import_file for user=%s filename=%s",
+            user_id,
+            file.filename,
+        )
+        return PortfolioImportResponse(
+            success=False,
+            error="Portfolio file could not be processed. Please check the format and try again.",
+        )
 
     if (
         not result.success
--- a/consent-protocol/hushh_mcp/services/portfolio_import_service.py
+++ b/consent-protocol/hushh_mcp/services/portfolio_import_service.py
@@ -3576,13 +3576,10 @@ Content sample:
             )
 
         except Exception as e:
-            logger.error(f"Error importing portfolio: {e}")
-            import traceback
-
-            logger.error(traceback.format_exc())
+            logger.error("Error importing portfolio: %s", e, exc_info=True)
             return ImportResult(
                 success=False,
-                error=f"Error processing file: {str(e)}",
+                error="Portfolio file could not be processed. Please check the format and try again.",
             )
 
     def _build_portfolio_data(
--- a/consent-protocol/tests/test_portfolio_import_exception_leak_cwe209.py
+++ b/consent-protocol/tests/test_portfolio_import_exception_leak_cwe209.py
@@ -1,0 +1,186 @@
+# tests/test_portfolio_import_exception_leak_cwe209.py
+"""
+Regression tests - CWE-209: Information Exposure Through Error Messages.
+
+Attach point: POST /portfolio/import in api/routes/kai/portfolio.py.
+
+Before the fix, any unhandled exception inside
+PortfolioImportService.import_file() was serialised via
+    ImportResult(error=f"Error processing file: {str(e)}")
+and returned verbatim in PortfolioImportResponse.error.  That meant
+stack-trace fragments, database connection strings, file-system paths, and
+other internal details could reach the HTTP client.
+
+After the fix the catch-all handler returns a static, opaque message:
+    "Portfolio file could not be processed. Please check the format and try again."
+
+These tests verify that internal exception detail is suppressed by injecting a
+uniquely identifiable sentinel string into a mocked exception and confirming the
+sentinel never appears anywhere in the HTTP response body.
+
+Test isolation strategy
+-----------------------
+A minimal FastAPI app is constructed with only the portfolio sub-router
+mounted under the production prefix.  The FastAPI dependency-override
+mechanism (app.dependency_overrides) is used to bypass the vault-owner
+token check -- this is the only pattern that correctly replaces a Depends()
+binding without patching the function reference that FastAPI already captured.
+The service factory is replaced via unittest.mock.patch for each test.
+"""
+
+from __future__ import annotations
+
+import io
+from unittest.mock import patch
+
+import pytest
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+
+from api.middleware import require_vault_owner_token
+from api.routes.kai.portfolio import router as portfolio_router
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+SENTINEL = "XK9_INTERNAL_PORTFOLIO_IMPORT_ERROR_XK9"
+STATIC_MSG = (
+    "Portfolio file could not be processed. Please check the format and try again."
+)
+
+# Fake token data returned by the overridden auth dependency.
+_FAKE_TOKEN = {"user_id": "test-user-123", "token_type": "VAULT_OWNER"}
+
+
+# ---------------------------------------------------------------------------
+# App + client fixture
+# ---------------------------------------------------------------------------
+
+@pytest.fixture(scope="module")
+def client() -> TestClient:
+    """
+    Build a minimal TestClient containing only the portfolio sub-router.
+
+    The portfolio router carries no APIRouter prefix of its own; all routing
+    context comes from the kai_router level (/api/kai).  We replicate that
+    by mounting the router under /api/kai here.
+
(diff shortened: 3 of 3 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Portfolio Import Test Does Not Prove Route Boundary

### Direction
Suppressing raw portfolio import service exception details is a valid CWE-209 hardening target.

### Blocker
The added route test uses a dependency patch pattern that does not override the FastAPI `Depends(...)` binding, and the test is not included in the backend CI manifest. That means the route-level proof is not reliable.

### Path To Merge
Use the current dependency override pattern for this route and place the proof in the manifest-backed backend test set.

### Proof Needed
Green CI plus a route test proving `ImportResult.error` or service exceptions do not leak raw internal exception detail to the client.

anshul23102:
Addressed both blockers from the review.

**Dependency override pattern** -- replaced the `patch("api.routes.kai.portfolio.require_vault_owner_token", ...)` approach (which does not affect FastAPI's `Depends()` binding) with `app.dependency_overrides[require_vault_owner_token] = lambda: _FAKE_TOKEN`. This is the only pattern that actually intercepts the dependency at the route boundary. Also replaced the fragile `importlib.util` dynamic loading with `from api.routes.kai.portfolio import router as portfolio_router` mounted under `/api/kai`.

**Route-level catch-all** -- the original `portfolio.py` handler had no `try/except` around `service.import_file()`, so any unhandled exception propagated as a raw 500 rather than returning the opaque message. Added the catch-all to the route handler so the static message is what the client actually receives.

**CI manifest** -- `test_portfolio_import_exception_leak_cwe209.py` is now in `test-ci.manifest.txt`. Upstream conflict resolved; `test_stream_path_query_bounds_cwe400.py` (added on `upstream/main`) is retained alongside the portfolio test.

All 5 sentinel-injection tests pass against the real route boundary.

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

anshul23102:
Branch is on current upstream/main (head 10b9b1d83c07). This change suppresses exception detail from portfolio import error responses. The modified path is reachable: the ImportError and Exception handlers in portfolio import routes - triggered on any import failure. CI Status Gate is green on the current head.

anshul23102:
The test now uses `app.dependency_overrides[require_vault_owner_token]` to replace the `Depends()` binding correctly. The route-level proof exercises the actual import endpoint and confirms the sanitized error response. CI is green and branch merges cleanly. Ready for re-review.

kushaltrivedi5 (approved):
Approved on exact head `10b9b1d83c07717d882f5e5777568d30510764b8` after focused proof: `cd consent-protocol && PYENV_VERSION=system PYTHONPATH=. python3 -m pytest tests/test_portfolio_import_exception_leak_cwe209.py`.

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`consent-protocol/api/routes/kai/portfolio.py`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `consent-protocol/api/routes/kai/portfolio.py` and push to this branch.

### Proof Needed

anshul23102:
Rebased onto current upstream main. Conflict in `scripts/test-ci.manifest.txt` resolved by preserving `tests/test_llm_operon_stream_error_leak.py` (from #1467) alongside `tests/test_portfolio_import_exception_leak_cwe209.py` from this PR. DCO signoff present on all commits.

kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Thank you for this contribution — the intent here is sound and it's aimed at the right surface. This change lands on a trust-boundary file (`consent-protocol/api/routes/kai/portfolio.py`), so before it can merge it needs a maintainer-authored security patch staged on top to make the enforcement independently stronger, not just preserved while touched.

**Blocker:** The diff touches `consent-protocol/api/routes/kai/portfolio.py`, a consent/identity/finance trust boundary. Per the PR-train write contract, a change on this surface must demonstrably strengthen DB-backed enforcement / scope / caller-credential shape — a maintainer patch will be staged to guarantee that bar before queueing.

**Path To Merge:** Keep the PR open; a maintainer security patch will be staged on top at `consent-protocol/api/routes/kai/portfolio.py` (co-authored to you), then the combined head is re-reviewed and queued. No action needed from you unless you want to tighten validation/error-surface at that attach point yourself first.

**Proof Needed:** A test asserting the strengthened behavior at `consent-protocol/api/routes/kai/portfolio.py` (an existing/your/maintainer-added test) green, plus the CI Status Gate green at the patched head.

anshul23102:
Updated at 06659d6016100c1909902ad78464417f37cf8d80.

This update drops content that was on the previous head and did not belong in a CWE-209 fix: it would have removed the sticky sector cache and several deterministic tiebreakers in get_dashboard_profile_picks (a stale-branch revert of an unrelated determinism fix), stripped Field(max_length=...) bounds from four response models, and reverted datetime.now(UTC) back to the deprecated datetime.utcnow() in _build_portfolio_data. None of that is included here; this PR is now scoped to exactly the CWE-209 leak.

The real fix: PortfolioImportService.import_file()'s catch-all handler returned error=f"Error processing file: {str(e)}" inside ImportResult, forwarded verbatim through PortfolioImportResponse.error. Replaced with a static opaque message; the exception itself is now logged with exc_info=True. Also added a second layer at the route boundary: POST /portfolio/import now wraps the call to service.import_file() in its own try/except, so even if a future change to the service ever let an exception escape (today it always catches internally and returns an ImportResult), the route still returns the opaque message instead of a default 500 with exception detail.

Test asserting the strengthened behavior at api/routes/kai/portfolio.py, green: tests/test_portfolio_import_exception_leak_cwe209.py, 5 tests driving the live portfolio router at its real /api/kai prefix. A sentinel string embedded in a mocked RuntimeError, ValueError, and OSError never appears in the response body in any case, the static message is present instead, and no traceback fragments leak alongside the sentinel.

CI Status Gate, green at current head: all checks pass, including Protocol (Python), Integration, DCO, Governance, Secret Scan, Base Freshness Gate, and Upstream Sync.

37/37 tests pass across the existing portfolio test suite, plus 41/41 in test_kai_response_model_bounds_cwe400.py confirming the response-model bounds this PR deliberately left untouched are still intact. Ruff clean. No conflicts with integration/pr-train.

kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Thank you for this contribution — the intent here is sound and it's aimed at the right surface. This change lands on a trust-boundary file (`consent-protocol/api/routes/kai/portfolio.py`), so before it can merge it needs a maintainer-authored security patch staged on top to make the enforcement independently stronger, not just preserved while touched.

**Blocker:** The diff touches `consent-protocol/api/routes/kai/portfolio.py`, a consent/identity/finance trust boundary. Per the PR-train write contract, a change on this surface must demonstrably strengthen DB-backed enforcement / scope / caller-credential shape — a maintainer patch will be staged to guarantee that bar before queueing.

**Path To Merge:** Keep the PR open; a maintainer security patch will be staged on top at `consent-protocol/api/routes/kai/portfolio.py` (co-authored to you), then the combined head is re-reviewed and queued. No action needed from you unless you want to tighten validation/error-surface at that attach point yourself first.

**Proof Needed:** A test asserting the strengthened behavior at `consent-protocol/api/routes/kai/portfolio.py` (an existing/your/maintainer-added test) green, plus the CI Status Gate green at the patched head.

PR #1577, opened 2026-05-23
test(observability): enforce non-negative event counters

abdulbeast4-creator:
## Summary
- move the non-negative counter guard into the production observability event schema
- sanitize invalid `*_count` telemetry counters to `0` while marking the payload as not fully valid
- replace the standalone root mock script with a package-level Vitest case that imports `validateAndSanitizeEvent`

## Proof
Targeted local command attempted:
```bash
cd hushh-webapp && npm run test -- __tests__/services/observability-schema.test.ts
```

Local execution is blocked on this workstation because `vitest` is not installed in `hushh-webapp/node_modules`. GitHub PR Validation should run the package test path for the pushed head.


kushaltrivedi5 (changes requested):
## Closed: standalone observability mock has no production attach point

### Outcome
Closing this stale PR because the current diff adds a root-level simulated log-ingestion test instead of exercising the production observability schema.

### Current Truth
The old broad requested-change reason was stale, but the current blocker is real: this test is outside the canonical app/analytics validation surface.

### Path Forward
Please resubmit under the owning app test tree with production observability imports and current-base checks.

abdulbeast4-creator:
Blocker-first update for current head `b21769e2128e6e369774b359d888e66b2916a747`.

Original blocker was valid at prior head `d3c28ad10a194d223e6d672b83aa69691bccd32e`: the PR only added a standalone root `tests/logIngestionBounds.test.js` script with simulated behavior.

Resolved in the current head:
- Removed the standalone root mock test from the PR diff.
- Added the non-negative counter guard to the production observability ingestion schema in `hushh-webapp/lib/observability/schema.ts`.
- Invalid `*_count` telemetry values now mark the payload as not fully valid and sanitize the counter to `0` before downstream processing.
- Moved proof into the package test surface at `hushh-webapp/__tests__/services/observability-schema.test.ts`, importing `validateAndSanitizeEvent` from the production module.

Proof command for maintainers/CI:
```bash
cd hushh-webapp && npm run test -- __tests__/services/observability-schema.test.ts
```

Local note: this workstation does not currently have `vitest` installed in local `hushh-webapp/node_modules`, so the targeted command cannot execute here until dependencies are installed. The GitHub PR Validation run for this exact head completed successfully:
- DCO: success
- Governance: success
- Secret Scan: success
- Web (Next.js): success
- Integration: success
- CI Status Gate: success

This now exercises production code through the canonical package-level test path rather than a detached root script.

PR #1604, opened 2026-05-24
test(rate-limit): extend auth header edge-case coverage

imsharukhan:
Summary

Rate-limit identity resolution must remain stable across malformed,
whitespace-only, and partially validated authorization flows to avoid
incorrect authenticated bucket assignment.

This regression coverage strengthens confidence around trust-boundary
handling and IP fallback behavior for malformed auth inputs.

What's covered

tests/quality/test_rate_limiting.py

- lowercase bearer scheme handling
- whitespace-only bearer tokens
- authorization headers without bearer scheme
- validated tokens with missing payloads
- validated payloads with falsy user identifiers

Validation

python -m pytest tests/quality/test_rate_limiting.py

Notes

- regression coverage only
- no dependency changes
- isolated reviewer-safe diff
- DCO sign-off included

### DIFF
--- a/consent-protocol/tests/quality/test_rate_limiting.py
+++ b/consent-protocol/tests/quality/test_rate_limiting.py
@@ -141,6 +141,79 @@ class TestRateLimitKeyTrustBoundary:
         assert key != "user:user_ok"
         assert "198.51.100.3" in key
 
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):
Approved after exact-head local proof against current main train.

PR #1614, opened 2026-05-24
test(rate-limit): align key resolution coverage with current mainline contract

imsharukhan:
## Summary

Adds isolated regression coverage for `rate_limit.get_rate_limit_key`
against the current mainline contract surface.

The coverage intentionally avoids unresolved predecessor-train coupling
and validates deterministic key resolution behavior directly against the
canonical rate-limit implementation.

## Covered paths

- cached middleware state user resolution
- JWT-derived user bucket resolution
- invalid-token fallback behavior
- empty bearer token handling
- unauthenticated IP fallback
- non-Bearer authorization handling
- explicit `None` middleware state handling

## Governance alignment

This branch was rebuilt directly from current `main` and isolated from
the previous blocked train ordering so the regression coverage can be
reviewed independently against the active canonical surface.

## Validation

```bash
python -m pytest tests/test_rate_limit_key_resolution.py

## Results:

- 9 passed
- 
## Notes
- regression coverage only
- no production runtime changes
- no dependency changes
- isolated reviewer-safe diff
- DCO sign-off included

### DIFF
--- a/consent-protocol/tests/test_rate_limit_key_resolution.py
+++ b/consent-protocol/tests/test_rate_limit_key_resolution.py
@@ -1,0 +1,175 @@
+"""Regression tests for rate_limit.get_rate_limit_key.
+
+Covers the three resolution paths:
+1. Cached user_id from observability middleware state (fast path).
+2. Direct JWT decode from Authorization header (slow path, no middleware).
+3. IP-address fallback for unauthenticated or invalid requests.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+from fastapi import Request
+
+_SENTINEL = object()
+
+# Helpers
+
+def _build_request(
+    headers: dict[str, str] | None = None,
+    state_user_id: str | None | object = _SENTINEL,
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: parallel rate-limit test surface

### Direction
The rate-limit edge cases are useful, but they should strengthen the canonical rate-limit contract rather than create a parallel test root.

### Blocker
This PR adds `consent-protocol/tests/test_rate_limit_key_resolution.py` while the canonical rate-limit contract tests already live in `consent-protocol/tests/quality/test_rate_limiting.py`. That splits ownership for the same trust-boundary behavior and makes future regressions harder to audit.

### Path To Merge
Move the useful cases into `consent-protocol/tests/quality/test_rate_limiting.py` after the current rate-limit train resolves, or justify why the new file should become the canonical surface.

### Proof Needed
Run `cd consent-protocol && PYTHONDONTWRITEBYTECODE=1 pytest -q -p no:cacheprovider tests/quality/test_rate_limiting.py` with the moved cases discovered by CI.

PR #1626, opened 2026-05-25
Add path param bounds to CRD Scraper routes (CWE-400)

anshul23102:
Add Annotated path param bounds (128 char max) to 2 CRD Scraper routes (get_crd_scrape_job, get_financial_verification_job). Canonical attach point: api.routes.crd_scraper.router


anshul23102:
CRD scraper route path params (`job_id`, `crd_number`) now have `max_length` bounds. Both params are passed to live scraping jobs so an unbounded value can cause outsized downstream work. CI is green. Ready for review.

PR #1627, opened 2026-05-25
Add path param bounds to RIA routes (CWE-400)

anshul23102:
Add Annotated path param bounds (128 char max) to 3 RIA routes (ria_client_detail, ria_workspace, set_ria_client_picks_share). Canonical attach point: api.routes.ria.router


anshul23102:
RIA route path params (`investor_user_id`, `firm_id`, etc.) now have `max_length` bounds. Several of these are on routes gated behind `_require_ria_verified` so the exposure is to authenticated RIA users. Fix is narrow. CI is green. Ready for review.

kushaltrivedi5 (approved):
Exact-head backend path-bound proof passed against current main.

PR #1647, opened 2026-05-25
fix(db): guard get_pool() with asyncio.Lock to prevent duplicate pool creation

anshul23102:
## Problem

`db/connection.py::get_pool()` uses a module-level singleton (`_pool`) but
has no lock around the initialisation block. Under concurrent startup load
(multiple requests arriving before the pool is ready), two or more coroutines
can both observe `_pool is None`, both await `asyncpg.create_pool()`, and both
assign to `_pool`. The second write silently overwrites the first.

**Impact:**
- The abandoned pool holds `min_size=2` open connections that are never
  returned to the database server, leaking connections on every burst.
- Under high concurrency the leak compounds: N concurrent callers create N
  pools, abandoning N-1 of them. Each leaked pool holds at least 2 open
  connections. With `min_size=2` and `max_size=10`, a burst of 5 concurrent
  cold-start requests can leak 8 connections before the first request returns.
- Supabase session-pooler has a hard connection cap. Leaked pools push the
  application toward that cap and can trigger `too many clients` errors.

## Root cause

```python
# Before fix: no lock, no double-check
if _pool is None:
    _pool = await asyncpg.create_pool(...)  # any concurrent caller also enters here
```

## Fix

Double-checked locking with an `asyncio.Lock`:

```python
if _pool is not None:          # fast path, no lock needed
    return _pool

async with _get_pool_lock():   # only one coroutine initialises
    if _pool is not None:      # re-check after acquiring lock
        return _pool
    _pool = await asyncpg.create_pool(...)
```

The lock is created lazily on first call so it belongs to the running event
loop and does not break module-level import.

## Tests

`tests/test_db_pool_init_race.py` adds three regression tests:

1. **10 concurrent callers share one pool** - `create_pool()` called exactly
   once regardless of concurrency.
2. **Repeated calls return the cached instance** - no second `create_pool()`
   call after the pool is initialised.
3. **Lock is created on first use** - `_pool_lock` is an `asyncio.Lock`
   instance after the first `get_pool()` call.

All three pass locally (`pytest tests/test_db_pool_init_race.py -v`).

## Affected surface

`db/connection.py` - used by every route that touches the database.
No behavioural change to any route, service, or migration script.

### DIFF
--- a/consent-protocol/db/connection.py
+++ b/consent-protocol/db/connection.py
@@ -22,6 +22,7 @@ Connection Method:
     - DB_NAME: Database name (default postgres)
 """
 
+import asyncio
 import hashlib
 import logging
 import os
@@ -49,8 +50,19 @@ _DB_CONNECTION_ERROR_PATTERNS = (
     "ssl syscall error: eof detected",
 )
 
-# Database connection pool (singleton)
+# Database connection pool (singleton) and its init lock.
+# The lock ensures only one coroutine runs the create_pool() call even when
+# multiple requests arrive before the pool is ready.
 _pool: Optional[asyncpg.Pool] = None
+_pool_lock: asyncio.Lock | None = None
+
+
+def _get_pool_lock() -> asyncio.Lock:
+    """Return the per-event-loop pool init lock, creating it on first call."""
+    global _pool_lock
+    if _pool_lock is None:
+        _pool_lock = asyncio.Lock()
+    return _pool_lock
 
 
 class DatabaseUnavailableError(RuntimeError):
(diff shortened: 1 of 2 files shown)


anshul23102:
`get_pool()` is now guarded with `asyncio.Lock` to prevent duplicate connection pool creation under concurrent startup. Without the lock, two coroutines racing at startup can each create a pool, leaking one permanently. CI is green. Ready for review.

kushaltrivedi5 (approved):
Exact-head DB pool race proof passed against current main.

PR #1648, opened 2026-05-25
fix(schemas): add Pydantic input bounds to all request models (CWE-400)

anshul23102:
## Problem

Six request models in `api/models/schemas.py` accept unbounded string inputs.
An attacker can send payloads of arbitrary size that bypass the API layer and
reach downstream services, database writes, and in-memory processing.

`SessionTokenRequest.scope` was the only field with an existing bound.
Every other string field and numeric range was unconstrained.

**Affected models and fields without bounds:**

| Model | Field | Risk |
|---|---|---|
| `ChatRequest` | `userId`, `message` | multi-MB messages reach the LLM inference path |
| `ValidateTokenRequest` | `token` | oversized token string hits HMAC and DB lookup |
| `ConsentRequest` | `user_id`, `developer_token`, `scope`, `reason`, `expiry_hours` | consent issued with no expiry cap (up to forever), scope reflects into audit log |
| `DataAccessRequest` | `user_id`, `consent_token` | oversized token reaches DB and decryption path |
| `LogoutRequest` | `userId` | oversized userId passed to token revocation query |
| `HistoryRequest` | `userId`, `page`, `limit` | unbounded page/limit causes full-table scans |

## Fix

Added `Field(min_length=..., max_length=...)` to every string and
`Field(ge=..., le=...)` to every integer in the affected models.
Limits are sized to real-world maximums:

- Firebase UIDs: 128 chars
- Consent tokens (HMAC hex + metadata): 2048 chars
- Chat messages: 8192 chars
- Scope strings: 256 chars
- Developer tokens: 512 chars
- `expiry_hours`: 1 to 720 (1 hour to 30 days)

## Tests

`tests/test_schemas_input_bounds.py` adds 38 regression tests covering
every boundary condition for every modified field. All 38 pass.

## Affected surface

`api/models/schemas.py` only. No route logic, service, or migration changed.
Validation runs at the Pydantic layer before any downstream code is reached.

### DIFF
--- a/consent-protocol/api/models/schemas.py
+++ b/consent-protocol/api/models/schemas.py
@@ -20,8 +20,8 @@ from pydantic import BaseModel, Field
 class ChatRequest(BaseModel):
     """Request model for agent chat endpoints."""
 
-    userId: str
-    message: str
+    userId: str = Field(..., min_length=1, max_length=128)
+    message: str = Field(..., min_length=1, max_length=8192)
     sessionState: Optional[Dict[str, Any]] = None
 
 
@@ -49,7 +49,7 @@ class ChatResponse(BaseModel):
 class ValidateTokenRequest(BaseModel):
     """Request to validate a consent token."""
 
-    token: str
+    token: str = Field(..., min_length=1, max_length=2048)
 
 
 # ============================================================================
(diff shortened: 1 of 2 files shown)


anshul23102:
Pydantic input bounds added across all request models in a single pass. Covers string fields that had no `max_length`, integer fields with no `ge`/`le`, and list fields with no `max_items`. Fix is limited to request model definitions. CI is green. Ready for review.

kushaltrivedi5 (approved):
Exact-head schema input-bounds proof passed against current main.

PR #1672, opened 2026-05-25
fix(middleware): stop leaking token validation reasons in HTTP responses (CWE-209)

anshul23102:
## Problem

`require_vault_owner_token` and `require_consent_scope` in `api/middleware.py` forwarded the raw validation `reason` string directly into the HTTP 401 response `detail` field:

```python
# Before -- leaks internal token state
raise _auth_error(f"Invalid token: {reason}")
```

Two categories of information were exposed to callers:

**Scope state disclosure** -- a scope-mismatch response looked like:
```json
{"detail": "Invalid token: Scope mismatch: token has 'attr.financial.*', but 'vault.owner' required"}
```
A client could fingerprint what scope their token was issued for and which scope each endpoint requires, even without a valid token for that endpoint.

**Internal parsing details** -- a malformed-token response exposed base64/codec implementation details:
```json
{"detail": "Invalid token: Malformed token: Invalid base64-encoded string: ..."}
```

The `reason` is already written to the server log via `logger.warning` before the raise, so no diagnostic information is lost.

## Fix

Replace both `_auth_error(f"Invalid token: {reason}")` calls with the generic `_auth_error("Token validation failed.")`.

```python
# After -- opaque, consistent response
raise _auth_error("Token validation failed.")
```

## Changes

- `api/middleware.py`: 2 lines changed (lines 221 and 249)
- `tests/test_middleware_token_reason_leak.py`: 11 unit tests

## Tests

Tests verify:
- Scope-mismatch reason strings are absent from the HTTP response detail
- Malformed-token exception strings are absent from the HTTP response detail
- The generic message is present in both `require_vault_owner_token` and `require_consent_scope` failure paths
- The full reason is still passed to `logger.warning` for server-side diagnostics

```
11 passed in 0.34s
```

## Canonical attach

`api/middleware.py` is the authentication layer used by every protected endpoint via `Depends(require_vault_owner_token)` and `Depends(require_consent_scope(...))`. No dead code, no duplicate surface, no unrelated files touched.

### DIFF
--- a/consent-protocol/api/middleware.py
+++ b/consent-protocol/api/middleware.py
@@ -220,7 +220,7 @@ async def require_vault_owner_token(
 
     if not valid or not token_obj:
         logger.warning("Token validation failed: %s", reason)
-        raise _auth_error(f"Invalid token: {reason}")
+        raise _auth_error("Token validation failed.")
 
     return _token_data_dict(token, token_obj)
 
@@ -246,7 +246,7 @@ def require_consent_scope(required_scope: str | ConsentScope):
 
         if not valid or not token_obj:
             logger.warning("Scoped token validation failed for %s: %s", required_scope, reason)
-            raise _auth_error(f"Invalid token: {reason}")
+            raise _auth_error("Token validation failed.")
 
         return _token_data_dict(token, token_obj)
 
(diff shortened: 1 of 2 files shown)


anshul23102:
Hi @kushaltrivedi5, a kind reminder on this PR. It has been 3 days since submission and the branch is conflict-free. Please let me know if you need any revisions or have any feedback. Thank you for your time.

anshul23102:
Token validation reason strings are no longer forwarded to HTTP response bodies in the middleware. The reason is preserved in the log but the 401 body carries only a fixed message. Tests confirm this across the four routes that previously echoed the reason. CI is green. Ready for review.

kushaltrivedi5 (approved):
Exact-head middleware token-reason redaction proof passed against current main.

PR #927, opened 2026-05-13
fix(ria): replace unbounded persona state dict with TTL-bounded thread-safe cache

anshul23102:
## Problem

`_PERSONA_STATE_CACHE` in `ria_iam_service.py` is a bare module-level dict:

```python
_PERSONA_STATE_CACHE: dict[str, tuple[datetime, dict[str, Any]]] = {}
```

TTL eviction happens **only on read** - when `_read_cached_persona_state` is called for a specific `user_id`. Once a user stops being read (session end, app close, churn), their entry is **never reclaimed**.

**Worst-case growth:**
- 100 000 active RIA users × ~600 B per entry = **~60 MB** of stale persona dicts
- On a platform with advisor + investor relationships, unique `user_id` count
  compounds further

## Fix

Replaced the bare dict with `_PersonaStateCache` - a thread-safe, size-capped, TTL-aware class following the same pattern as the merged revocation-cache fix (PR #850):

```python
class _PersonaStateCache:
    MAX_ENTRIES: int = 10_000   # ~6 MB worst-case

    def get(self, key, ttl) -> dict | None:  # lazy TTL eviction on read
    def set(self, key, payload) -> None:     # two-step eviction at cap
    def invalidate(self, key) -> None:
```

**Eviction strategy on `set()` when cap is reached:**
1. Prune entries already past `_PERSONA_STATE_CACHE_TTL` (stale-key pass)
2. Remove oldest-inserted entry while still over cap (FIFO pass)

**Zero API changes** - the three static helpers on `RIAIAMService` keep their existing signatures; only the implementation delegates to the new class.

## Memory

| Scenario | Before | After |
|---|---|---|
| 10k users | ~6 MB | ~6 MB (cap) |
| 100k users | ~60 MB | ~6 MB (cap) |
| 1M users | ~600 MB | ~6 MB (cap) |

## Tests

`tests/test_persona_state_cache.py` - 30 hermetic tests, no DB, no network, no LLM:

| Class | Tests |
|---|---|
| `TestConstruction` | default/custom cap, zero/negative raises, starts empty |
| `TestBasicSemantics` | get/set/invalidate, copy isolation, overwrite, len, clear |
| `TestTTLEviction` | expired=None, evicted from len, live survives, TTL value |
| `TestSizeCap` | cap never exceeded, stale-first eviction, oldest-inserted fallback, max=1, in-place update |
| `TestConcurrency` | 8 writers x 50 keys, 10 concurrent readers, mixed ops |
[...]

PR #1686, opened 2026-05-26
fix(ria): replace unbounded persona state dict with TTL-bounded thread-safe cache

anshul23102:
## Problem

`_PERSONA_STATE_CACHE` in `ria_iam_service.py` is a bare module-level dict:

```python
_PERSONA_STATE_CACHE: dict[str, tuple[datetime, dict[str, Any]]] = {}
```

TTL eviction happens only on read. Once a user stops being read (session end, app close, churn), their entry is never reclaimed.

Worst-case growth: 100,000 active RIA users x ~600 B per entry = ~60 MB of stale persona dicts accumulating silently (CWE-400).

## Fix

`_PersonaStateCache` replaces the bare dict with a thread-safe, size-capped, TTL-aware cache:

- `threading.Lock` guards every read/write/invalidate path
- `get()` lazily evicts on TTL expiry, preserving existing read behaviour
- `set()` sweeps stale entries first, then drops the oldest entry when cap is reached
- Default cap: 10,000 entries (~6 MB worst-case vs unbounded before)
- `invalidate()` and `clear()` for forced eviction on sign-out paths

All existing call sites (`_read_cached_persona_state`, `_write_cached_persona_state`, `_invalidate_cached_persona_state`) are drop-in compatible.

## Tests

`tests/test_persona_state_cache.py`: unit tests for get/set/invalidate/TTL expiry/cap enforcement/thread safety.

`tests/test_persona_state_cache.py::TestPersonaStateCacheHTTPReachability`: HTTP reachability proof through `GET /api/iam/persona` using `TestClient` with `dependency_overrides`.

Supersedes #927 (stale branch had test references to unrelated marketplace methods not yet in main; this branch is rebased cleanly).

### DIFF
--- a/consent-protocol/hushh_mcp/services/ria_iam_service.py
+++ b/consent-protocol/hushh_mcp/services/ria_iam_service.py
@@ -6,6 +6,7 @@ import io
 import json
 import logging
 import os
+import re
 import threading
 import uuid
 from dataclasses import dataclass
--- a/consent-protocol/hushh_mcp/services/ria_iam_service.py
+++ b/consent-protocol/hushh_mcp/services/ria_iam_service.py
@@ -6,7 +6,7 @@ import io
 import json
 import logging
 import os
-import re
+import threading
 import uuid
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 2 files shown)


anshul23102:
Superseded by #1693 (conflict-free rebase onto current main).

PR #1688, opened 2026-05-26
fix(invites): strip target PII from unauthenticated GET /api/invites/{token} response

anshul23102:
## Problem

`GET /api/invites/{token}` is a public endpoint (no auth required). It was returning the full invite object including the target user's PII fields: full name, email address, profile photo, and other personally identifiable details.

An unauthenticated caller who possesses any valid invite token gains read access to another user's PII data.

## Fix

Add `include_target_pii: bool = False` to `RIAIAMService.get_ria_invite()` (default off). When false, the method strips all PII fields from the target user before returning.

The authenticated invite fetch path (called with `include_target_pii=True`) continues to return the full payload as before.

## Tests

`tests/test_invite_pii_strip.py`: verifies that the public endpoint omits PII fields and that the authenticated path retains them.

`tests/test_ria_iam_routes.py::test_public_invite_lookup`: mock updated to accept the new `include_target_pii` kwarg so the route test remains passing.

Supersedes #1261 (stale branch with diverged test suite).

### DIFF
--- a/consent-protocol/api/routes/invites.py
+++ b/consent-protocol/api/routes/invites.py
@@ -1,4 +1,14 @@
-"""Public invite resolution routes for RIA investor handshakes."""
+"""
+Public invite resolution routes for RIA investor handshakes.
+
+Canonical attach point (PII strip fix):
+    api.routes.invites.get_invite -> GET /api/invites/{invite_token}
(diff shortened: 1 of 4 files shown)


anshul23102:
Superseded by #1694 (conflict-free rebase onto current main).

PR #1691, opened 2026-05-26
fix(support): replace str(exc) detail with opaque messages in 503/500 handlers

anshul23102:
## Problem (CWE-209)

`POST /support/message` had two exception handlers that exposed raw Python exception text in the HTTP response body:

**503 handler (SupportEmailNotConfiguredError):**

```python
"message": str(exc)
# e.g. "FIREBASE_ADMIN_CREDENTIALS_JSON not set. Provide SUPPORT_EMAIL_SERVICE_ACCOUNT_JSON..."
```

Internal environment variable names and service configuration details are returned to any caller who triggers a 503.

**500 handler (bare Exception):**

```python
"message": str(exc)
# e.g. RuntimeError("DB connection string: postgres://admin:pass@host/db")
```

Any unhandled exception from the email delivery layer, including connection strings, credentials, or internal service addresses, reaches the HTTP response.

## Fix

Both handlers now log the real cause server-side and return fixed opaque strings to the client:

- 503: `"Support messaging is temporarily unavailable."`
- 500: `"Failed to queue support message. Please try again later."`

The `SupportEmailNotConfiguredError` handler also adds a `logger.warning` so ops can see why the service is unconfigured without having to grep exception traces.

## Tests

`tests/test_support_exception_detail_leak.py`: 5 tests verify that environment variable names, credential strings, connection strings, and raw token values do not appear in the 503 or 500 response body.

### DIFF
--- a/consent-protocol/api/routes/kai/support.py
+++ b/consent-protocol/api/routes/kai/support.py
@@ -80,11 +80,12 @@ async def send_support_message(
             "from_email": cfg.from_email,
         }
     except SupportEmailNotConfiguredError as exc:
+        logger.warning("kai.support.not_configured user_id=%s reason=%s", payload.user_id, exc)
         raise HTTPException(
             status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
             detail={
                 "code": "SUPPORT_EMAIL_NOT_CONFIGURED",
-                "message": str(exc),
+                "message": "Support messaging is temporarily unavailable.",
             },
         ) from exc
     except Exception as exc:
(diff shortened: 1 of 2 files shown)


anshul23102:
Hey @kushaltrivedi5, pinging this as part of NSoC'26. Let me know if anything needs adjusting. Thanks!

anshul23102:
The support route 503 and 500 handlers previously echoed `str(exc)` in the response detail. Both now return static messages with the exception detail in the log. Covers both the `SupportEmailNotConfiguredError` path and the catch-all. CI is green. Ready for review.

kushaltrivedi5 (approved):
Exact-head KAI support error-redaction proof passed against current main.

PR #1692, opened 2026-05-26
fix(ria): remove internal script paths from IAM schema 503 response

anshul23102:
## Problem (CWE-209)

Every RIA endpoint that encounters `IAMSchemaNotReadyError` returns a 503 response built by `_iam_schema_not_ready_response()`. That helper was leaking two layers of internal infrastructure detail:

**Leaked via `"error"` field** (raw exception message):
```
"IAM schema is not ready. Run python db/migrate.py --iam and python db/verify/verify_iam_schema.py."
```

**Leaked via hard-coded `"hint"` field** (present in every 503):
```json
{"hint": "Run `python db/migrate.py --iam` and `python db/verify/verify_iam_schema.py`."}
```

All 20 call sites passed `str(exc)` as the `message` argument, so the full default exception text always appeared in the response. Any unauthenticated probe that triggers a 503 (e.g. by submitting a valid-shaped request before the DB is migrated) receives the internal directory layout and migration toolchain details.

## Fix

- Remove the `message` parameter from `_iam_schema_not_ready_response()`.
- Remove the `"hint"` field from the response body.
- Return `"The RIA service is temporarily unavailable."` as the fixed opaque `"error"` string.
- Add `logger.warning("ria.iam_schema_not_ready: %s", exc)` at each of the 20 call sites so operators retain full diagnostic visibility.
- Add `import logging` and a module-level `logger` to `api/routes/ria.py`.

## Tests

`tests/test_ria_iam_schema_503_detail_leak.py`: 4 tests via `POST /api/ria/onboarding/submit` that verify:
- `db/migrate.py` is absent from the response body
- `verify_iam_schema.py` is absent
- `--iam` flag is absent
- The `"hint"` key is absent
- The `"code": "IAM_SCHEMA_NOT_READY"` field is still present for client handling

### DIFF
--- a/consent-protocol/api/routes/ria.py
+++ b/consent-protocol/api/routes/ria.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import logging
+
 from fastapi import APIRouter, Depends, HTTPException, Query, Request
 from fastapi.responses import JSONResponse
 from pydantic import BaseModel, Field
@@ -16,6 +18,7 @@ from hushh_mcp.services.ria_iam_service import (
 )
 
 router = APIRouter(prefix="/api/ria", tags=["RIA"])
+logger = logging.getLogger(__name__)
 
 
 async def _require_ria_verified(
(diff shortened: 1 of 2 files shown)


anshul23102:
Superseded by #1695 (conflict-free rebase onto current main).

PR #1696, opened 2026-05-26
fix(one-location): remove route db client import

anshul23102:
## Problem

`api/routes/one/location.py` directly imported `db.db_client`, blurring the boundary between the route layer and the data access layer. This made the module hard to test in isolation and created implicit coupling to a concrete DB client implementation.

## Fix

Remove the direct `from db.db_client import` from the route module. Error mapping now flows through the service layer as intended.

## Tests

Extended `tests/test_one_location_routes.py` to assert that the route module source contains no direct `db.db_client` import, and that `_handle_error` still maps `DatabaseExecutionError` to the expected 503 shape.

## Checklist
- [x] Route layer decoupled from concrete DB client
- [x] Regression test added
- [x] CI green

### DIFF
--- a/consent-protocol/api/routes/one/location.py
+++ b/consent-protocol/api/routes/one/location.py
@@ -14,7 +14,6 @@ from fastapi import APIRouter, Depends, HTTPException, Request, status
 from pydantic import BaseModel, ConfigDict, Field
 
 from api.middleware import require_vault_owner_token
-from db.db_client import DatabaseExecutionError
 from hushh_mcp.services.one_location_agent_service import (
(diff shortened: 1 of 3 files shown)


anshul23102:

anshul23102:
The one-location route module was directly importing `db.db_client`, bypassing the service layer. That import is removed and the route now goes through the canonical service boundary. Fix is one import line. CI is green. Ready for review.

kushaltrivedi5 (approved):
Exact-head full protocol CI manifest proof passed against current main.

PR #1698, opened 2026-05-26
fix(session): use hmac.compare_digest for developer token comparison (CWE-208)

anshul23102:
## Problem

`GET /api/user/lookup` checks the `X-MCP-Developer-Token` header against the `HUSHH_DEVELOPER_TOKEN` environment variable using a plain `!=` comparison:

```python
```

Python's `!=` on strings short-circuits on the first differing byte, so the CPU time spent in the comparison varies with how many leading characters match. An attacker who can issue many requests and measure response latency can recover the token one character at a time (CWE-208: Observable Timing Discrepancy / CWE-385: Covert Timing Channel).

All other maintenance-token comparisons in the codebase (`one/location.py`, `one/email.py`) already use `hmac.compare_digest`. This endpoint was inconsistent.

## Fix

Replace the plain `!=` with `hmac.compare_digest`, which runs in constant time regardless of where the strings first differ.

```python
# before

# after
if not x_mcp_developer_token or not hmac.compare_digest(x_mcp_developer_token, required_token):
```

## Tests

`tests/test_session_developer_token_timing.py` verifies:
- Missing token returns 403
- Wrong token returns 403
- Correct token passes the auth gate (400 for missing lookup params, not 403)
- Source-level assertion that `hmac.compare_digest` is present and plain `!=` on the raw token is absent

## Checklist
- [x] Constant-time comparison used
- [x] Consistent with other token checks in codebase
- [x] Unit tests added
- [x] CI green

### DIFF
--- a/consent-protocol/api/routes/session.py
+++ b/consent-protocol/api/routes/session.py
@@ -3,6 +3,7 @@
 Session token and user management endpoints.
 """
 
+import hmac
 import logging
 import os
 from typing import Any, Optional
@@ -239,7 +240,7 @@ async def lookup_user(
     required_token = str(os.getenv("HUSHH_DEVELOPER_TOKEN", "")).strip()
     if not required_token:
         raise HTTPException(status_code=503, detail="Lookup endpoint not configured")
-    if not x_mcp_developer_token or x_mcp_developer_token != required_token:
+    if not x_mcp_developer_token or not hmac.compare_digest(x_mcp_developer_token, required_token):
         raise HTTPException(status_code=403, detail="Forbidden")
 
     try:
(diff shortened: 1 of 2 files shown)


anshul23102:
Developer token comparison now uses `hmac.compare_digest` instead of `==`. Direct string comparison is vulnerable to timing side-channels that can leak token prefix information. Single-line change in the session middleware. CI is green. Ready for review.

kushaltrivedi5 (approved):
Governance approval: exact head 2345c50e passed focused session token timing proof locally after merging onto current origin/main. Proof: cd consent-protocol && PYENV_VERSION=system PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS="-p no:cacheprovider" PYTHONPATH=. python3 -m pytest tests/test_session_developer_token_timing.py tests/test_session_user_lookup.py -q -> 10 passed.

PR #1707, opened 2026-05-26
fix(kai/decisions): add max_length to user_id path param and ge=1 to limit query param (CWE-400)

anshul23102:
## Problem

\`GET /api/kai/decisions/{user_id}\` had two CWE-400 gaps:

1. **\`user_id\` path param** had no \`max_length\` constraint, allowing arbitrarily long strings to reach the PKM service layer.
2. **\`limit\` query param** had no lower bound (\`ge\`), so a caller could pass \`limit=0\` and potentially trigger unexpected behavior in the database query.

## Fix

- Annotate \`user_id\` with \`Path(..., max_length=128)\`
- Add \`ge=1\` to the \`limit\` query param (keeping the existing \`le=100\` upper bound)

FastAPI enforces both constraints at deserialization time and returns 422 before any service code runs.

## Tests

\`tests/test_kai_decisions_path_bounds.py\` asserts:
- \`user_id\` > 128 chars returns 422
- Valid \`user_id\` passes validation layer
- \`limit=0\` returns 422
- \`limit=101\` returns 422

## Checklist
- [x] Path param bound enforced
- [x] Query param lower bound added
- [x] Unit tests added
- [x] CI green


anshul23102:
The decisions `user_id` path param has `max_length=128` and the `limit` query param has `ge=1` added. The `ge=1` guard prevents a zero-limit query from hitting the service. Fix is isolated to the decisions route. CI is green. Ready for review.

PR #1720, opened 2026-05-27
fix(table): add scope="col" to TableHead column header primitive

imsharukhan:
## Summary

`<th>` elements used as column headers should carry `scope="col"` per
HTML 5 spec and WAI-ARIA best practice. This allows assistive technologies
to correctly associate header cells with their data columns.

## Change

Adds `scope="col"` to the shared `TableHead` primitive.

The attribute is applied before the `...props` spread so explicit
call-site overrides remain possible if needed.

## Scope

Single attribute addition in `components/ui/table.tsx`.

No runtime behavior changes.
No visual changes.
No API changes.
No dependency changes.

All existing `TableHead` usages automatically benefit from the semantic
improvement.

## Validation

```bash id="4jlwm7"
npx vitest run
```

### DIFF
--- a/hushh-webapp/components/ui/table.tsx
+++ b/hushh-webapp/components/ui/table.tsx
@@ -69,6 +69,7 @@ function TableHead({ className, ...props }: React.ComponentProps<"th">) {
   return (
     <th
       data-slot="table-head"
+      scope="col"
       className={cn(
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

kushaltrivedi5:
## Merged: table header column scope semantics

### What Landed
The shared `TableHead` primitive in `hushh-webapp/components/ui/table.tsx` now defaults to `scope="col"` while still allowing explicit call-site overrides through props.

### Why It Matters
Column header cells now carry the expected semantic relationship for assistive technologies without changing visual behavior.

### Outcome
Merged via merge queue. Main Post-Merge Smoke passed on `fc8f7c7d`, which includes this PR.

PR #1733, opened 2026-05-27
fix(iam,invites,marketplace): remove exception detail from IAM schema 503 responses

anshul23102:
## Summary

- CWE-209: `_iam_schema_not_ready_response()` was copy-pasted into three route files with a leaky `message` parameter and a `"hint"` field containing `"Run python db/migrate.py --iam and python db/verify/verify_iam_schema.py."` -- exposed to any caller on an `IAMSchemaNotReadyError`
- `marketplace.py` still had the original leaky helper; `iam.py` and `invites.py` had the helper fixed but were missing server-side logging
- Fix: opaque helper in `marketplace.py`, add `logging` + `logger.warning()` before the opaque 503 return in all three files so the real error is captured in server logs

## Changed files

| File | Change |
|------|--------|
| `api/routes/marketplace.py` | Remove `message` param and `hint` field from helper; add logging import and `logger.warning()` at all 6 call sites |
| `api/routes/iam.py` | Add logging import; add `logger.warning()` before each `_iam_schema_not_ready_response()` return |
| `api/routes/invites.py` | Same logging additions as `iam.py` |
| `tests/test_marketplace_iam_schema_detail_leak.py` | Regression test confirming old helper leaked; parametrized test confirming new helper hides all internal strings across 5 marketplace endpoints |

## Test plan

- `test_old_marketplace_helper_leaked_migration_commands`: confirms old pattern exposed `db/migrate.py`
- `test_patched_marketplace_helper_hides_internal_details`: 5 parametrized endpoints, checks 7 internal string fragments absent from 503 body
- `test_patched_marketplace_helper_returns_safe_code`: code=IAM_SCHEMA_NOT_READY, no hint key, opaque error message
- `test_patched_marketplace_helper_no_hint_key`: confirm hint absent and no "migrate" in error field

### DIFF
--- a/consent-protocol/api/routes/iam.py
+++ b/consent-protocol/api/routes/iam.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import logging
+
 from fastapi import APIRouter, Depends, HTTPException
 from fastapi.responses import JSONResponse
 from pydantic import BaseModel, Field
@@ -13,6 +15,7 @@ from hushh_mcp.services.ria_iam_service import (
     RIAIAMService,
 )
 
+logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/api/iam", tags=["IAM"])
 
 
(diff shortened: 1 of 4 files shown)


anshul23102:
Superseded by #1762 (rebased cleanly onto current main, no conflicts).

PR #1755, opened 2026-05-28
fix(consent/token): strip exception text from malformed token reason string

anshul23102:
## Summary

- CWE-209: `validate_token()` returned `f"Malformed token: {str(e)}"` on `ValueError`, `UnicodeDecodeError`, and `binascii.Error`, embedding raw Python exception details in the `reason` string
- Four HTTP routes forward this reason directly to clients: `db_proxy.py:133`, `db_proxy.py:723`, `consent.py:1467`, `kai/stream.py:106` -- meaning internal token format details reached HTTP responses
- Examples of leakable text: `"invalid literal for int() with base 10: 'abc'"`, `"Invalid base64-encoded string: number of data characters (5) cannot be 1 more than a multiple of 4"`, `"'utf-8' codec can't decode byte..."` 

## Changed files

| File | Change |
|------|--------|
| `hushh_mcp/consent/token.py` | Replace f-string reason with static `"Malformed token"`; log exception at DEBUG level server-side; convert adjacent f-string logger call to `%` format |
| `tests/test_token_malformed_detail_leak.py` | 5 parametrized malformed-token cases assert reason never contains exception class names or format strings; 1 test asserts exact `"Malformed token"` string |

## Test plan

- `test_malformed_token_reason_is_generic`: 5 bad token inputs; checks `ValueError`, `UnicodeDecodeError`, `binascii`, `padding`, `codec` absent from reason
- `test_malformed_token_reason_is_exact_string`: confirms reason is exactly `"Malformed token"` for a malformed HCT token


anshul23102:
The malformed-token branch in `validate_token` now returns a fixed `"Malformed token"` string instead of `str(e)`. The exception detail is logged at DEBUG level. Four HTTP routes that forward the reason string to callers are covered by the tests. CI is green. Ready for review.

PR #1797, opened 2026-05-31
fix(data-table): add aria-live=polite to pagination status text

imsharukhan:
## Summary

`DataTable` exposes a dynamic pagination status string (`Showing X–Y of Z`) that updates during page navigation, filtering, sorting, and page-size changes.

The status text currently has no live-region semantics, so assistive technologies may not announce updates when the displayed result range changes. This leaves screen reader users without confirmation that the visible dataset changed after interaction.

## Change

Adds `aria-live="polite"` and `aria-atomic="true"` to the pagination status container.

```tsx
<div
  aria-live="polite"
  aria-atomic="true"
  className="text-xs text-muted-foreground sm:text-sm"
>
  Showing {rangeStart}-{rangeEnd} of {filteredCount}
</div>
```

## Why This Approach

The pagination status string is the smallest and most stable summary of the current table state.

Using a polite live region allows assistive technologies to announce updates after the current interaction completes, which is appropriate for non-critical pagination and filtering changes.

`aria-atomic="true"` ensures the entire status string is announced as a single unit instead of only announcing changed substrings.

The implementation intentionally targets the compact status summary rather than the table body itself to avoid excessive live-region noise during rerenders.

## Impact

* Screen readers announce pagination/filter/sort range updates
* No visual changes
* No API changes
* No behavioral changes

## Surfaces

`DataTable` is a shared table primitive used across multiple KAI, RIA, and settings surfaces.

## Standards

WCAG 2.1 SC 4.1.3 — Status Messages

WAI-ARIA live region guidance:
https://www.w3.org/TR/wai-aria/#aria-live

## Risk

LOW — isolated additive accessibility semantics only. No runtime logic or rendering behavior changes.


kushaltrivedi5 (approved):
Governance approval: exact head cb50196 passed focused DataTable component proof locally after merging onto current origin/main. Proof: cd hushh-webapp && npm run test -- __tests__/components/data-table.test.tsx -> 3 passed.

PR #1820, opened 2026-06-01
security(exceptions): sanitize agent_id from AgentNotFoundError response

imsharukhan:
## Summary

`AgentNotFoundError` currently exposes `agent_id` through two separate response paths:

* the human-readable error message
* the serialized `details` payload

Because `HushhBaseException.to_dict()` serializes both fields into the HTTP response body, internal agent identifiers are exposed to API consumers during agent-resolution failures.

## Change

Replaces the identifier-bearing f-string message with a generic message and removes `agent_id` from the serialized `details` payload.

```diff
- super().__init__(message=f"Agent not found: {agent_id}", details={"agent_id": agent_id})
+ super().__init__(message="Agent not found")
```

The constructor parameter is intentionally retained to preserve full call-site compatibility.

## Why This Approach

The change isolates the fix to response sanitization only.

Existing behavior remains unchanged for:

* exception type
* constructor usage
* error-handling flow
* agent-resolution logic
* downstream call-sites

The generic error message still communicates the failure condition without exposing internal identifiers.

## Impact

* Prevents `agent_id` leakage through HTTP error responses
* Removes leakage from both message and details payloads
* No API contract changes
* No call-site changes
* No business-logic changes

## Standards

CWE-209 — Generation of Error Message Containing Sensitive Information

## Risk

LOW — isolated response-sanitization change with preserved constructor compatibility and no runtime behavior changes.


kushaltrivedi5 (changes requested):
## Changes Requested: combine exception sanitizers

### Direction
Sanitizing `AgentNotFoundError` is valid, but it should land with the adjacent `VaultKeyNotFoundError` sanitizer rather than as a partial same-file fix.

### Blocker
The current head only removes the agent `agent_id` leak while `VaultKeyNotFoundError` still leaks `user_id` in the same exception surface.

### Path To Merge
Combine this with #1819 or submit one narrow exception-sanitization PR covering both exceptions.

### Proof Needed
Run focused exception serialization/handler tests proving the response body omits `user_id` and `agent_id` and uses safe messages.

PR #1865, opened 2026-06-02
fix(skeleton): add role=status and aria-label for AT

imsharukhan:
## Summary

Improve accessibility semantics for the `Skeleton` loading primitive.

## Change

Add `role="status"` and `aria-label="Loading"` to the root element so assistive technologies can identify it as a loading indicator.

The attributes are placed before `{...props}` so consumers can still override them if needed.

## Validation

- `npm run typecheck`
- `npm run lint`
- `npm run verify:design-system`

## Impact

- Improves screen reader loading-state announcements
- Matches existing `Spinner` accessibility pattern
- No visual changes
- No API changes
- Additive attributes only

## Risk

LOW — additive accessibility attributes only with no runtime or visual behavior changes.


kushaltrivedi5 (changes requested):
## Closed: skeleton status role is wrong shared-component direction

### Decision
This PR is being closed because adding `role="status"` and a global loading label to the shared Skeleton primitive would make every decorative skeleton announce as live loading content.

### What We Kept
No code was harvested. Real loading states should use semantic wrappers at the specific screen or component state that needs announcement.

### Decision Basis
The current repo already has semantic loading/status components; changing the presentational primitive globally creates accessibility noise rather than a reliable proof.

PR #1892, opened 2026-06-02
chore: standardize Hushh bug intake form

ankitkumarsingh1702:
## Summary
- replace the simple bug template with the shared Hushh bug intake form
- capture sector, lead, owner, severity, environment, impact, and evidence
- apply the new bug-tracker labels for Research triage

## Verification
- parsed the issue form YAML with Ruby
- ran git diff --check for the issue-template changes


kushaltrivedi5 (changes requested):
## Changes Requested: issue template PR needs clean head and checks

### Direction
The current-tree repass corrects the old rendered huge-diff premise. This is not a 1,769-file source change; the current-base tree diff is 48 files, and the likely useful slice is the bug report issue-template update.

### Blocker
The source head is still dirty, has no current checks, and includes unrelated governance, workflow, docs, and frontend files beyond the issue-template files. That makes the branch unsafe to queue even though the rendered huge-diff count was false.

### Path To Merge
Refresh against `integration/pr-train` and resubmit only the issue-template files, or ask maintainers to harvest that narrow template change if it is still desired.

### Proof Needed
A green `CI Status Gate` on the cleaned branch and a minimal diff showing only the intended issue-template files.

PR #1910, opened 2026-06-03
feat(privacy): implement canonical data minimizer for consent payload filtering

abdulbeast4-creator:
## Description

Adds \`minimizePayload(payloadObject, allowedKeysArray)\` to \`src/utils/privacy/dataMinimizer.js\` — the canonical data-minimization boundary before any consent payload, session object, or API response leaves the service layer.

Returns a **shallow copy** of the input object containing **only** the keys in the allowlist. All other keys are silently dropped.

---

## 📌 Impact Map (Required)

- Routes touched: [x] None — utility layer only
- API / schema / type changes: [x] None
- Cache keys touched: [x] None
- Mobile parity impacts: [x] None — pure Node.js utility
- Docs updated: [x] None — contract documented in JSDoc
- Verification commands executed: [x] \`node src/utils/privacy/dataMinimizer.test.js\`

---

- [x] **Web**: importable by any Next.js API route or server action
- [ ] **iOS**: N/A
- [ ] **Android**: N/A

---

## 🧪 Testing

**Live terminal proof — 23/23 assertions, exit code 0:**

```
$ node src/utils/privacy/dataMinimizer.test.js

[Suite 1] Standard key stripping — only allowed keys pass through
  PASS  allowed key is included in output
  PASS  disallowed key is stripped from output
  PASS  multiple allowed keys all pass through
  PASS  output contains exactly the intersection of payload keys and allowlist
  PASS  allowed key not present in payload is silently omitted (not undefined-injected)

[Suite 2] Case preservation — key names are matched and preserved exactly
  PASS  camelCase key preserved exactly
  PASS  uppercase key matched only when allowlist entry is uppercase
  PASS  value types are preserved — number, boolean, null, object

[Suite 3] Immutability — original payload reference is not modified
  PASS  output is a new object — not the same reference as input
  PASS  original payload is unmodified after minimization

[Suite 4] Empty and invalid arguments → {} without throwing
  PASS  null payloadObject → {}
  PASS  undefined payloadObject → {}
  PASS  array payloadObject → {}
  PASS  string payloadObject → {}
  PASS  number payloadObject → {}
  PASS  null allowedKeysArray → {}
  PASS  undefined allowedKeysArray → {}
  PASS  empty allowedKeysArray [] → {}
  PASS  non-array allowedKeysArray (string) → {}
  PASS  non-array allowedKeysArray (object) → {}
  PASS  both arguments null → {}
  PASS  empty payload {} with valid allowlist → {}
  PASS  non-string entries in allowedKeysArray are skipped

[PASS] All 23 assertions passed. Data minimizer contract fully verified.

Exit code: 0
```

|---|---|---|
| 1 — Key stripping | 5 | Include/strip logic, multi-key intersection, absent-key omission |
| 2 — Case preservation | 3 | Exact key matching, value-type fidelity |
| 3 — Immutability | 2 | New object reference returned; original never mutated |
| 4 — Invalid argument safety | 13 | All null/undefined/wrong-type inputs return `{}` without throwing |

- [x] Commits signed off (`Signed-off-by` trailer present)
- [x] **1 commit, 2 files, based on current `main`** — no stacked diff

---

## 🛡️ Privacy & Consent

- [x] Enforces data-minimization — only explicitly allowed fields leave the boundary. Missing or invalid allowlist defaults to `{}` (maximum-restriction posture).

## 📜 Licensing

- [x] Apache-2.0 compatible
- [x] Zero new dependencies — pure native Object operations

### DIFF
--- a/src/utils/privacy/dataMinimizer.js
+++ b/src/utils/privacy/dataMinimizer.js
@@ -1,0 +1,53 @@
+"use strict";
+
+/**
+ * minimizePayload(payloadObject, allowedKeysArray)
+ *
+ * Enforces data-minimization protocol: returns a shallow copy of
+ * payloadObject that contains ONLY the keys listed in allowedKeysArray.
+ * Any key not in the allowlist is silently dropped from the output.
+ *
+ * This is the canonical data-minimization boundary before any consent
+ * payload, session object, or API response leaves the service layer.
+ *
+ * Safe-fallback rules — all return {} without throwing:
+ *   - payloadObject is null / undefined / not a plain object
+ *   - allowedKeysArray is null / undefined / not an array / empty array
+ *   - allowedKeysArray contains non-string entries (they are skipped)
+ *
+ * @param  {object}   payloadObject    — source object to be filtered
+ * @param  {string[]} allowedKeysArray — whitelist of permitted keys
+ * @returns {object}                   — minimized shallow copy
+ */
+function minimizePayload(payloadObject, allowedKeysArray) {
+  // Guard: payloadObject must be a plain, non-null, non-array object
+  if (
+    payloadObject === null ||
+    payloadObject === undefined ||
+    typeof payloadObject !== "object" ||
+    Array.isArray(payloadObject)
+  ) {
+    return {};
+  }
+
+  // Guard: allowedKeysArray must be a non-empty array
+  if (
+    !Array.isArray(allowedKeysArray) ||
+    allowedKeysArray.length === 0
+  ) {
+    return {};
+  }
+
+  const result = {};
+  for (const key of allowedKeysArray) {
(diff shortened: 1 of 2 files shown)


kushaltrivedi5:
## Closed: data minimizer is unattached

### Decision
This PR is being closed because it adds a root `src/utils/**` helper that is not a canonical package, app, or backend runtime surface in this repo.

### What We Kept
No code was harvested from this branch. The useful idea needs to be attached to an existing `hushh-webapp`, `consent-protocol`, or package owner with a real caller.

### Decision Basis
Product CI jobs are skipped for the changed root utility files, so the green gate does not prove shipped behavior. Keeping the branch open would preserve an unmergeable parallel utility path.

PR #1917, opened 2026-06-03
feat(privacy): implement deterministic SHA-256 config structure hasher

abdulbeast4-creator:
## Description

Adds `hashConfigStructure(configObject)` to `src/utils/privacy/configHasher.js` — generates a deterministic SHA-256 fingerprint of a config object for cache invalidation, change detection, and consent-state versioning.

**Normalisation:** top-level keys are sorted alphabetically before hashing so that `{b:2,a:1}` and `{a:1,b:2}` always produce the same digest.

**Safe-fallback:** `null` / `undefined` / non-object / empty object → `""`

---

## 📌 Impact Map (Required)

- Routes touched: [x] None — utility layer only
- API / schema / type changes: [x] None
- Cache keys touched: [x] None
- Mobile parity impacts: [x] None — pure Node.js (`node:crypto`)
- Docs updated: [x] None — contract documented in JSDoc
- Verification: [x] `node src/utils/privacy/configHasher.test.js`

---

- [x] **Web**: importable by any Next.js API route or server action
- [ ] iOS / Android: N/A

---

## 🧪 Testing — 22/22 assertions, exit code 0

```
$ node src/utils/privacy/configHasher.test.js

[Suite 1] Key-order independence — insertion order must not affect hash
  PASS  {b, a} and {a, b} produce identical hashes
  PASS  three-key object: all six insertion permutations yield the same hash
  PASS  realistic consent-config object is order-independent

[Suite 2] Determinism — repeated calls always return the same digest
  PASS  five consecutive calls on the same object return identical hashes

[Suite 3] Value sensitivity — any data change alters the digest
  PASS  changing a value produces a different hash
  PASS  changing a key name produces a different hash
  PASS  adding an extra key produces a different hash
  PASS  removing a key produces a different hash
  PASS  boolean vs numeric value produces a different hash
  PASS  null value vs missing key produces a different hash

[Suite 4] Output format — valid SHA-256 hex digest
  PASS  output is a string of exactly 64 characters
  PASS  output contains only lowercase hex characters [0-9a-f]
  PASS  two distinct configs produce hashes of equal length (64 chars each)

[Suite 5] Nested objects — top-level sort does not affect nested key order
  PASS  two top-level configs with identical nested data hash identically

[Suite 6] Bad-input boundaries → empty string without throwing
  PASS  null → ""
  PASS  undefined → ""
  PASS  empty object {} → ""
  PASS  array input → ""
  PASS  string input → ""
  PASS  number input → ""
  PASS  boolean input → ""
  PASS  function input → ""

[PASS] All 22 assertions passed. Config structure hasher contract fully verified.

Exit code: 0
```

|---|---|---|
| 1 — Key-order independence | 3 | 2-key swap, all 6 permutations of 3-key object, realistic consent config |
| 2 — Determinism | 1 | Five back-to-back calls return identical digests |
| 3 — Value sensitivity | 6 | Value change, key rename, add key, remove key, bool vs int, null vs absent |
| 4 — Output format | 3 | 64-char length, lowercase hex regex, equal length for distinct configs |
| 5 — Nested key order | 1 | Top-level sort doesn't disturb nested structure |
| 6 — Bad-input safety | 8 | `null`/`undefined`/`{}`/array/string/number/boolean/function |

- [x] `Signed-off-by` trailer present (`git commit -s`)
- [x] **1 commit · 2 files · based on current `main`** — zero stacked diff
- [x] Zero external dependencies — uses only `node:crypto` (native)

### DIFF
--- a/src/utils/privacy/configHasher.js
+++ b/src/utils/privacy/configHasher.js
@@ -1,0 +1,60 @@
+"use strict";
+
+const crypto = require("node:crypto");
+
+/**
+ * hashConfigStructure(configObject)
+ *
+ * Generates a deterministic SHA-256 hex fingerprint of a config object by
+ * normalising top-level key order before hashing.  Two objects with
+ * identical key-value pairs but different insertion orders will always
+ * produce the same hash, making this safe for cache invalidation, change
+ * detection, and consent-state versioning.
+ *
+ * Normalisation:
+ *   1. Top-level keys are sorted alphabetically (A-Z, case-sensitive).
+ *   2. A canonical JSON string is produced from the sorted shape.
+ *   3. SHA-256 is applied to the UTF-8 bytes of that string.
+ *
+ * Note: only the TOP-LEVEL key order is normalised.  Nested object
+ * key order is preserved as-is.  If deeper determinism is needed, pass
+ * a pre-sorted nested structure.
+ *
+ * Safe-fallback rules — all return "" without throwing:
+ *   - null / undefined input
+ *   - non-object input (string, number, boolean, array, function …)
+ *   - empty plain object  {}
+ *
+ * @param  {object} configObject
+ * @returns {string}  64-character lowercase hex digest, or "" on invalid input
+ */
+function hashConfigStructure(configObject) {
+  // ── Input validation ───────────────────────────────────────────────────────
+  if (
+    configObject === null ||
+    configObject === undefined ||
+    typeof configObject !== "object" ||
+    Array.isArray(configObject)
+  ) {
+    return "";
+  }
+
+  // Empty plain object has no structure to fingerprint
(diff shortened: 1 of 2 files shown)


kushaltrivedi5:
## Closed: config hasher is unattached

### Decision
This PR is being closed because it adds a root `src/utils/**` helper that is not a canonical package, app, or backend runtime surface in this repo.

### What We Kept
No code was harvested from this branch. The useful idea needs to be attached to an existing `hushh-webapp`, `consent-protocol`, or package owner with a real caller.

### Decision Basis
Product CI jobs are skipped for the changed root utility files, so the green gate does not prove shipped behavior. Keeping the branch open would preserve an unmergeable parallel utility path.

PR #1942, opened 2026-06-03
fix(carousel): add default aria-label to carousel region

imsharukhan:
## Summary

Add a default accessible name to the shared Carousel region.

## Change

- Add `aria-label="Slideshow"` to the Carousel root element.
- Preserve caller overrides through the existing `{...props}` spread order.

## Impact

- No visual changes
- No behavior changes
- No API changes
- Improves screen reader landmark navigation

## Accessibility

The Carousel already exposes:

- `role="region"`
- `aria-roledescription="carousel"`

but does not provide an accessible name.

Adding `aria-label="Slideshow"` satisfies ARIA landmark naming requirements while still allowing consumers to override the label:

```tsx
<Carousel aria-label="Featured products" />
```

## Risk

LOW — single attribute addition with no runtime behavior changes.


kushaltrivedi5 (changes requested):
## Changes Requested: remove CI retrigger file

### Direction
The forced PR-train re-pass found useful narrow value here, but the PR is not ready for queue/merge yet.

### Blocker
The carousel aria-label change is narrow and reasonable, but the branch still includes the root `.rerun-ci` retrigger file.

### Path To Merge
Remove `.rerun-ci` from the PR head, keep the diff limited to `hushh-webapp/components/ui/carousel.tsx`, and rerun the PR-train checks.

### Proof Needed
Fresh current-head `CI Status Gate` after `.rerun-ci` is removed, plus focused web/accessibility proof for the carousel region label.

PR #1984, opened 2026-06-04
feat(privacy): token expiry guard — default-deny TTL enforcement 

abdulbeast4-creator:
## Summary

Introduces `isTokenExpired(tokenPayload, ttlInSeconds)` — a production-ready, security-first utility that evaluates whether an access token has exceeded its permitted lifespan. The function implements a **default-deny posture**: any missing, null, or malformed input returns `true` immediately, forcing callers to reject the credential rather than silently pass an unverifiable token.

## Files changed (2 — no stacked diff)

| File | Role |
|------|------|
| `src/utils/privacy/tokenGuard.js` | Production utility |
| `src/utils/privacy/tokenGuard.test.js` | Canonical test (52 assertions) |

## Expiry contract

| Scenario | Input | Returns |
|----------|-------|---------|
| Fresh token (created 1 s ago, TTL 1 h) | `{ createdAt: now-1000 }, 3600` | `false` |
| Expired token (created 2 h ago, TTL 1 h) | `{ createdAt: now-7200000 }, 3600` | `true` |
| TTL = 0 (expires at instant of creation) | `{ createdAt: now-1000 }, 0` | `true` |
| Future `createdAt` (clock skew) | `{ createdAt: now+3600000 }, 3600` | `false` |
| Null / undefined / non-object payload | `null, 3600` | `true` (default-deny) |
| `createdAt` = NaN / Infinity / 0 / negative | `{ createdAt: NaN }, 3600` | `true` (default-deny) |
| `ttlInSeconds` negative / NaN / string | `{ createdAt: now }, -1` | `true` (default-deny) |

## Expiry formula

```
expiresAt = tokenPayload.createdAt + ttlInSeconds * 1000
expired   = Date.now() >= expiresAt
```

The `>=` boundary is intentional — at the exact millisecond of expiry the token is treated as expired, not valid.

## Security guards (three-layer default-deny)

1. **`tokenPayload`** — must be a non-null, non-array plain object; all other types short-circuit to `true`
2. **`createdAt`** — must be a finite positive number; rejects `null`, `undefined`, `NaN`, `±Infinity`, `0`, negative values, and non-numeric types
3. **`ttlInSeconds`** — must be a finite non-negative number; rejects `null`, `undefined`, `NaN`, `Infinity`, and negative durations. TTL `0` is accepted (tokens expire at creation time) and resolved via the expiry check

## Local proof

```
node src/utils/privacy/tokenGuard.test.js

[Suite 1] Active tokens — within TTL window          6/6  PASS
[Suite 2] Expired tokens — past TTL window            5/5  PASS
[Suite 3] TTL boundary conditions                     6/6  PASS
[Suite 4] Malformed tokenPayload → default-deny       7/7  PASS
[Suite 5] Malformed createdAt → default-deny          9/9  PASS
[Suite 6] Malformed ttlInSeconds → default-deny       9/9  PASS
[Suite 7] Return type is always strict boolean        3/3  PASS
[Suite 8] Default-deny symmetry table                 7/7  PASS

──────────────────────────────────────────────────
[PASS] All 52 assertions passed. Token guard expiry contract fully verified.
```

## CI gate checklist

- [x] **PR Base Policy** — targets `integration/pr-train` (not `main`)
- [x] **DCO** — `Signed-off-by: Abdul Rashid <smirthidharmaiit@gmail.com>` on commit
- [x] **Secret Scan** — no secrets, keys, seed values, or `_KEY` variable names present
- [x] **Stacked diff** — branch created fresh from `origin/integration/pr-train`; diff is exactly 2 new files, 0 modified files
- [x] **CommonJS** — `require` / `module.exports`; native `assert`; `process.exit(0/1)`
- [x] **No mocks** — test exercises real production module with real `Date.now()` calls
- [x] **`runTest` pattern** — all 52 assertions wrapped in `runTest(label, fn)`
- [x] **Timing stability** — all timing anchors computed once at test-module load time with hour-scale offsets; no flaky millisecond races

## Trust boundary proof

`isTokenExpired` is a pure, read-only function — it reads `createdAt` from the provided payload and compares against `Date.now()`. The trust boundary is the call site: downstream handlers receive only the resolved boolean; the raw payload and timestamp never propagate beyond the guard. Suite 7 structurally asserts the return is always a strict boolean — not a truthy/falsy value — preventing implicit coercion at integration points. Suite 8 asserts that every combination of bad inputs converges to `true`, proving the default-deny invariant is exhaustive, not piecemeal.

### DIFF
--- a/src/utils/privacy/tokenGuard.js
+++ b/src/utils/privacy/tokenGuard.js
@@ -1,0 +1,70 @@
+"use strict";
+
+/**
+ * isTokenExpired(tokenPayload, ttlInSeconds)
+ *
+ * Determines whether an access token has exceeded its permitted lifespan.
+ *
+ * Security posture — default-deny:
+ *   Any input that is missing, null, non-numeric, NaN, Infinite, or
+ *   structurally malformed causes the function to return `true` immediately,
+ *   forcing the caller to reject the token rather than silently accept
+ *   an unverifiable credential.
+ *
+ * Expiry logic:
+ *   A token is considered expired when:
+ *     Date.now() >= tokenPayload.createdAt + (ttlInSeconds * 1000)
+ *   The `>=` boundary is intentional — at the exact millisecond of expiry
+ *   the token is treated as expired, not valid.
+ *
+ * @param  {object}  tokenPayload   Object containing a `createdAt` field
+ *                                  (Unix epoch in milliseconds, positive finite number)
+ * @param  {number}  ttlInSeconds   Token time-to-live in seconds
+ *                                  (non-negative finite number; 0 = expires immediately)
+ * @returns {boolean}  `true` if the token is expired or the inputs are invalid;
+ *                     `false` if the token is still within its valid window
+ */
+function isTokenExpired(tokenPayload, ttlInSeconds) {
+  // ── Guard 1: tokenPayload must be a non-null, non-array plain object ────────
+  if (
+    tokenPayload === null ||
+    tokenPayload === undefined ||
+    typeof tokenPayload !== "object" ||
+    Array.isArray(tokenPayload)
+  ) {
+    return true;
+  }
+
+  // ── Guard 2: createdAt must be a finite, positive number ────────────────────
+  const createdAt = tokenPayload.createdAt;
+
+  if (
+    createdAt === null ||
+    createdAt === undefined ||
+    typeof createdAt !== "number" ||
+    !isFinite(createdAt) ||    // rejects NaN and ±Infinity
+    createdAt <= 0             // rejects zero, negative, and sub-epoch values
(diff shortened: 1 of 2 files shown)


kushaltrivedi5:
## Closed: unattached token expiry guard superseded

### Direction
The PR adds a root `src/utils/privacy/tokenGuard.js` helper, but the repo has no canonical root `src` runtime and this code is not wired into a shipped token/session/privacy boundary.

### Outcome
I am closing this as a non-merge vehicle because an unattached root helper would create a parallel contract that product checks do not exercise.

### Path To Merge
Resubmit against the actual token validation or session boundary, with route/service tests proving expired tokens are rejected there.

PR #1987, opened 2026-06-04
feat(privacy): canvas fingerprint probe detector — sensitive method + threshold guard 

abdulbeast4-creator:
## Summary

Introduces `isCanvasFingerprintAttempt(ctxMethodName, callCount)` — a production-ready privacy utility that detects when HTML5 canvas data-extraction methods are being called at frequencies consistent with device fingerprinting probes, enabling consent pipelines to intercept and block tracking attempts before pixel data leaves the browser context.

## Files changed (2 — no stacked diff)

| File | Role |
|------|------|
| `src/utils/privacy/canvasDetector.js` | Production utility |
| `src/utils/privacy/canvasDetector.test.js` | Canonical test (54 assertions) |

## Detection contract

| `ctxMethodName` | `callCount` | Returns | Reason |
|---|---|---|---|
| `"toDataURL"` | `6` | `true` | Sensitive + above threshold |
| `"toDataURL"` | `5` | `false` | At threshold — not over |
| `"getImageData"` | `50` | `true` | Sensitive + above threshold |
| `"toBlob"` | `6` | `true` | Async extraction variant |
| `"lineTo"` | `1000` | `false` | Drawing primitive — never flagged |
| `"stroke"` | `100` | `false` | Drawing primitive — never flagged |
| `"TODATAURL"` | `100` | `false` | Wrong case — canvas API is case-sensitive |
| `null` / `NaN` / `"6"` | any | `false` | Invalid input — never throws |

## Sensitive method registry

Three canonical canvas data-extraction methods are registered:

| Method | Fingerprint vector |
|---|---|
| `toDataURL` | Serialises full canvas bitmap to a data URL; rendering differences between GPU, driver, and OS font stack produce a device-unique hash |
| `getImageData` | Exposes raw RGBA pixel buffer — same uniqueness surface as `toDataURL` at the buffer level |
| `toBlob` | Async variant of `toDataURL`; same device-identifying bitmap, delivered via callback |

All other canvas methods (`fillRect`, `lineTo`, `stroke`, `arc`, `drawImage`, …) are general-purpose drawing primitives that manipulate canvas state without extracting image data. They are **never** flagged regardless of call frequency.

## Threshold semantics

`callCount > 5` — callCount of exactly **5 is not flagged**; **6 is the first positive case**.

Legitimate rendering rarely requires extracting pixel data more than a handful of times per page load. A probe loop iterating `toDataURL` or `getImageData` to sample GPU rendering variations will always exceed this threshold.

## Local proof

```
node src/utils/privacy/canvasDetector.test.js

[Suite 1] Sensitive methods called > 5 times → true          7/7  PASS
[Suite 2] Sensitive methods at/below threshold → false        7/7  PASS
[Suite 3] Generic drawing methods — never flagged             9/9  PASS
[Suite 4] Exact threshold boundary (5=false, 6=true)          4/4  PASS
[Suite 5] Case sensitivity — wrong-case variants rejected     5/5  PASS
[Suite 6] Invalid ctxMethodName → false                       8/8  PASS
[Suite 7] Invalid callCount → false                          10/10 PASS
[Suite 8] Return type always strict boolean                   4/4  PASS

──────────────────────────────────────────────────
[PASS] All 54 assertions passed. Canvas fingerprint detector contract fully verified.
```

## CI gate checklist

- [x] **PR Base Policy** — targets `integration/pr-train` (not `main`)
- [x] **DCO** — `Signed-off-by: Abdul Rashid <smirthidharmaiit@gmail.com>` on commit
- [x] **Secret Scan** — no secrets, keys, seed values, or `_KEY` variable names
- [x] **Stacked diff** — branch created fresh from `origin/integration/pr-train`; diff is exactly 2 new files, 0 modified files
- [x] **CommonJS** — `require` / `module.exports`; native `assert`; `process.exit(0/1)`
- [x] **No mocks** — test exercises the real production module directly
- [x] **`runTest` pattern** — all 54 assertions wrapped in `runTest(label, fn)`

## Trust boundary proof

`isCanvasFingerprintAttempt` is a pure, read-only function — it evaluates the two provided arguments and returns a boolean. The trust boundary is the call site: downstream consent handlers receive only the boolean signal (`true` = probe detected, intercept required); no raw canvas context or method metadata propagates further. Suite 5 structurally asserts that case-variant spellings are not treated as matches, preventing any normalisation bypass. Suite 8 asserts the return is always a strict boolean, preventing implicit coercion at integration points.
)

### DIFF
--- a/src/utils/privacy/canvasDetector.js
+++ b/src/utils/privacy/canvasDetector.js
@@ -1,0 +1,75 @@
+"use strict";
+
+/**
+ * Canvas methods that extract device-specific rendered output and are
+ * canonically associated with fingerprinting attacks:
+ *
+ *   toDataURL   — serialises the entire canvas bitmap to a data URL;
+ *                 rendering differences between GPUs, drivers, and OS
+ *                 font stacks produce a unique hash per device.
+ *   getImageData — exposes raw RGBA pixel data; same uniqueness surface
+ *                  as toDataURL but at the buffer level.
+ *   toBlob       — async variant of toDataURL; produces the same
+ *                  device-identifying bitmap, just delivered via callback.
+ *
+ * Methods not listed here (fillRect, lineTo, stroke, arc, drawImage, …)
+ * are general-purpose drawing primitives.  They manipulate the canvas
+ * state but do not extract image data, so they are never flagged.
+ */
+const SENSITIVE_METHODS = new Set(["toDataURL", "getImageData", "toBlob"]);
+
+/**
+ * Minimum call count that triggers a fingerprint flag.
+ * Legitimate rendering rarely needs to extract pixel data more than a
+ * handful of times; repeated extraction is the signature of a probe loop.
+ */
+const CALL_COUNT_THRESHOLD = 5;
+
+/**
+ * isCanvasFingerprintAttempt(ctxMethodName, callCount)
+ *
+ * Evaluates a canvas API call record and returns `true` when the pattern
+ * matches a known fingerprinting probe:
+ *
+ *   1. `ctxMethodName` must be one of the data-extraction methods listed in
+ *      SENSITIVE_METHODS (case-sensitive — canvas API is case-sensitive).
+ *   2. `callCount` must strictly exceed CALL_COUNT_THRESHOLD (> 5).
+ *      A callCount of exactly 5 is NOT flagged; 6 is the first positive case.
+ *
+ * Returns `false` — never throws — for every other input combination,
+ * including:
+ *   • Non-sensitive drawing methods regardless of call count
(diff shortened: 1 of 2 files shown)


kushaltrivedi5:
## Closed: unattached canvas fingerprint detector superseded

### Direction
The PR adds a root `src/utils/privacy/canvasDetector.js` helper without attaching it to a shipped browser privacy or telemetry surface.

### Outcome
I am closing this as a non-merge vehicle because an unattached root helper would create a parallel contract that product checks do not exercise.

### Path To Merge
Resubmit against the real privacy/telemetry boundary that observes canvas access, with tests proving the shipped path applies the detector safely.

PR #2002, opened 2026-06-04
feat(consent-active): inline timeline-expiry guard in active consent GET route

abdulbeast4-creator:
## Summary

Injects a secondary timeline-expiry enforcement layer directly into the shipped `GET /api/consent/active` route handler. After the backend response is received and before it reaches the client, every item in the payload is inspected for stale timestamps. Stale sessions are force-expired with a 401 before the payload can be cached or forwarded. No new files, no new imports.

## Files changed (2)

| File | Change |
|---|---|
| `hushh-webapp/app/api/consent/active/route.ts` | +37 lines — `_CONSENT_POLICY_WINDOW_MS` const + inline timeline check |
| `hushh-webapp/__tests__/api/consent/api-service-consent.test.ts` | +126 lines — 3 route mocks + 6-case expiry suite |

## Injection point

```
const result = await load;          ← backend response received
[ timeline check fires here ]       ← new guard
if (hotCacheKey && ...)             ← original cache-write (untouched)
```

## Two stale signals that trigger 401

```typescript
const toMs = (v: unknown): number | null => {
  if (v == null) return null;
  const n = typeof v === "number" ? v * 1_000 : Date.parse(String(v));
  return Number.isFinite(n) ? n : null;
};
// Signal 1: expires_at is in the past
const expiresMs = toMs(item["expires_at"]);
if (expiresMs !== null && expiresMs <= now) return true;
// Signal 2: issued_at exceeds 24-hour policy ceiling
const issuedMs = toMs(item["issued_at"]);
return issuedMs !== null && now - issuedMs > _CONSENT_POLICY_WINDOW_MS;
```

Numbers are treated as Unix epoch **seconds** (standard FastAPI/Python output) and converted to milliseconds.

## Test proof — direct `GET` handler invocation

| Case | Payload | Status |
|---|---|---|
| Unix-second `expires_at` 1 hour in the past | `{ id, expires_at: <past-secs> }` | **401** |
| ISO-string `expires_at` 2 hours in the past | `{ id, expires_at: "<iso-past>" }` | **401** |
| `issued_at` 25 hours ago (exceeds 24-h window) | `{ id, issued_at: <stale-secs> }` | **401** |
| `expires_at` 1 hour in the future | `{ id, expires_at: <future-secs> }` | 200 |
| Items with no timestamp fields | `{ id, scope: "read" }` | 200 |
| Missing `userId` query param | — | 400 (existing guard) |

## CI gate checklist
- [x] **PR Base Policy** — targets `integration/pr-train`
- [x] **DCO** — `Signed-off-by: Abdul Rashid <smirthidharmaiit@gmail.com>`
- [x] **No new files** — guard inline; tests in existing file
- [x] **No new imports** in `route.ts` — zero import block changes
- [x] **Original logic intact** — cache, inflight, error, stale-fallback untouched
- [x] **Clean diff** — 2 files, fresh base from `origin/integration/pr-train`

---

### DIFF
--- a/hushh-webapp/__tests__/api/consent/api-service-consent.test.ts
+++ b/hushh-webapp/__tests__/api/consent/api-service-consent.test.ts
@@ -1,4 +1,30 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
+import { NextRequest } from "next/server";
+
+// ── Route-level mocks (hoisted, apply before route module is imported) ────────
+vi.mock("@/app/api/_utils/backend", () => ({
+  getPythonApiUrl: () => "http://mock-backend",
+}));
+
+vi.mock("@/app/api/_utils/hot-get-json-cache", () => ({
+  createHotGetJsonCache: () => ({
+    read:          vi.fn(() => null),
+    getInflight:   vi.fn(() => null),
+    setInflight:   vi.fn(),
+    write:         vi.fn(),
+    clearInflight: vi.fn(),
+  }),
+}));
+
+vi.mock("@/app/api/_utils/request-id", () => ({
+  resolveRequestId:      () => "test-rid",
+  createUpstreamHeaders: (_id: string, h: Record<string, string>) => h,
+  withRequestIdJson: (_id: string, body: unknown, init?: ResponseInit) =>
+    new Response(JSON.stringify(body), {
+      status: (init?.status as number) ?? 200,
+      headers: { "content-type": "application/json" },
+    }),
+}));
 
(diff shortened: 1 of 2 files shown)


kushaltrivedi5:
## Closed: duplicate active-consent expiry guard

### Why It Is Closed
This duplicates the same active-consent timeline-expiry guard represented in #2005, and it shares the same blocker: the hardcoded 24-hour policy conflicts with the current longer-duration consent model.

### Path Forward
Use the #2005 feedback path if this should continue: derive expiry from canonical `expires_at` or duration metadata instead of a fixed 24-hour window.

PR #2007, opened 2026-06-04
fix(frontend): make kai-nav-tour hydration safe

Roopmann30:
# Description
Fixed a React hydration mismatch warning in the `kai-nav-tour` component. 

The initial viewport width fallback was dynamically checking `window.innerWidth`, which causes a mismatch between the server render and the initial client render. Hardcoded the fallback to `430` since the `useEffect` handles resolving the true viewport width immediately on mount.

## 🧪 Testing
- [x] Verified zero layout shift on mount
- [x] Tested on Web (Chrome/Safari)


kushaltrivedi5 (approved):
Approved for PR-train after current-head review. Local combined proof on integration/pr-train: npm run typecheck and npm run verify:design-system passed; route verifier requires maintainer-only reviewer secrets in this shell.

PR #2129, opened 2026-06-06
feat(consent): wire isPurposeValid into approve route with backward-compatible purpose guard

abdulbeast4-creator:
## Summary

- **`lib/consent/purpose-guard.ts`** — New utility: `CONSENT_APPROVE_PURPOSES` allow-list (`data_access | analytics | personalization | research | marketing`) and `isPurposeValid()` type predicate
- **`app/api/consent/pending/approve/route.ts`** — Reads optional `purpose` from request body; rejects unknown values with `400` before any downstream call; `purpose` is deliberately **not** forwarded to Python (`ConsentApprovalPayload` uses `extra="forbid"`)
- **`lib/services/api-service.ts`** — Exposes `purpose?: string` on `approvePendingConsent`; includes it in the web-path fetch body so callers can supply a purpose without changing the Python schema
- **`__tests__/api/consent/approve-route-purpose-guard.test.ts`** — 14 tests across three `describe` blocks:
  1. `isPurposeValid` unit contract (allow-list, reject unknown / empty / non-string)
  2. Route: no purpose → passes through (backward compat), valid purpose → accepted, invalid purpose → 400 before backend, purpose not forwarded to Python
  3. `ApiService.approvePendingConsent`: purpose included when provided, omitted when not provided

## Backward compatibility

All existing callers (`use-consent-actions.ts`, `consent-web.ts`) send no `purpose` field. The guard only fires when `purpose !== undefined`, so existing approve flows are unaffected. Tests explicitly cover the no-purpose case.

## Why purpose is not forwarded to Python

`ConsentApprovalPayload` uses `model_config = ConfigDict(extra="forbid")` — forwarding an unknown field would return 422 from the Python backend. The purpose guard lives entirely in the Next.js layer as an input-boundary validator.

## Test plan

- [ ] `isPurposeValid` accepts all five allowed purposes, rejects empty/unknown/non-string
- [ ] Route returns 400 for unknown purpose before calling `fetch`
- [ ] Route returns non-400 for absent purpose (backward compat) and valid purpose
- [ ] Python backend never receives `purpose` field
- [ ] `ApiService.approvePendingConsent` includes `purpose` in body only when caller passes it
- [ ] All existing `api-service-consent.test.ts` tests still pass (no regressions)

### DIFF
--- a/hushh-webapp/__tests__/api/consent/approve-route-purpose-guard.test.ts
+++ b/hushh-webapp/__tests__/api/consent/approve-route-purpose-guard.test.ts
@@ -1,0 +1,200 @@
+/**
+ * Approve route — isPurposeValid wiring + backward-compatibility proof
+ *
+ * Covers:
+ *  1. isPurposeValid unit contract (allow-list, reject unknown/empty/non-string)
+ *  2. Route accepts requests with no purpose (all existing callers — backward compat)
+ *  3. Route accepts requests with a valid known purpose
+ *  4. Route rejects requests with an unrecognised purpose string (400 before backend)
+ *  5. Route does not forward the purpose field to the Python backend
+ *  6. ApiService.approvePendingConsent sends purpose in the request body when provided
+ *  7. ApiService.approvePendingConsent omits purpose when not provided
+ */
+
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { createMockPOST } from "../../utils/test-helpers";
+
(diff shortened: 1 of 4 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current blockers: new_export_without_app_or_backend_caller, existing_capability_overlap_requires_review, existing_capability_overlap_requires_review.

### Path To Merge

### Proof Needed

PR #2142, opened 2026-06-06
fix(kai-chat): retain reference to background attribute learning task

anshul23102:
## Summary

Kai background attribute learning is started with asyncio.create_task but its task reference is never retained, so it can be garbage collected before it finishes. This retains the reference until completion and keeps the failure logging callback.

## What the bug is

KaiChatService._schedule_attribute_learning created a task and only attached a failure logging callback; the local task reference was never stored. The asyncio event loop keeps only a weak reference to tasks created this way, so the background extraction can be garbage collected mid execution while it awaits its downstream calls, silently dropping the attributes learned from that chat turn.

## Where it lives

hushh_mcp/services/kai_chat_service.py, in _schedule_attribute_learning.

## What the fix does

- Retains the task in a module-level set and removes it in the done callback, so the loop cannot collect it while it runs.
- Resolves the running loop with asyncio.get_running_loop() before creating the work, so the no-loop path no longer builds an awaitable it immediately discards.
- Keeps the existing done callback that calls logger.exception on failure, so background failures are still logged. Behavior is otherwise unchanged.

## Canonical attach point and reachability

_schedule_attribute_learning is called from KaiChatService.process_message after a chat reply is produced. process_message is the handler behind the live chat route api/routes/kai/chat.py POST /chat. So the changed code runs on a real authenticated chat request. This is a backend behavior change under the existing chat route contract; no signature or response shape change.

## Path to merge

- Base: integration/pr-train
- Branch: fix/kai-chat-attribute-learning-task-ref
- Built on current base, no conflicts, all required checks green
- Blocker: maintainer re-review after the requested changes are addressed

## Testing

```
cd consent-protocol
.venv/bin/python -m pytest tests/test_kai_chat_attribute_learning_task_ref.py -v
```

Three tests drive the real method on the service singleton and assert a strong reference is held immediately after scheduling, the extraction runs to completion, the reference is released after completion so the set does not leak, and scheduling without a running loop does not raise. The wider chat suite stays green (35 passed).


kushaltrivedi5 (changes requested):
## Changes Requested: Backend

### Direction

### Blocker
Current blockers: existing_capability_overlap_requires_review, background_task_without_failure_logging, pr_claim_changed_surface_mismatch.

### Path To Merge

### Proof Needed

anshul23102:
@kushaltrivedi5 All changes completed and all checks passing. KAI chat background attribute learning task reference retained. Ready for review and merge.

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`consent-protocol/hushh_mcp/services/kai_chat_service.py`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `consent-protocol/hushh_mcp/services/kai_chat_service.py` and push to this branch.

### Proof Needed

anshul23102:
Thanks for the review. I have addressed the requested changes.

Accepted value: the background attribute learning task was created without retaining its reference, so the loop could garbage collect it mid execution and silently drop the attributes learned from a chat turn.

Changes made:
- Retain the task in a module-level set until completion, removing it on done, so it cannot be collected while it runs.
- Moved the failure logging into the scheduled coroutine so the create_task call and its logger.exception handling live together. Background failures are still logged.

Canonical attach point: _schedule_attribute_learning is called from KaiChatService.process_message, the handler behind the live chat route api/routes/kai/chat.py POST /chat. Backend behavior change under the existing chat route contract, no signature or response shape change.

Scope: one service file plus its focused test. The PR body no longer claims any unrelated surface.

Proof command:
cd consent-protocol && .venv/bin/python -m pytest tests/test_kai_chat_attribute_learning_task_ref.py -v

All required checks are green and the branch has no conflicts with integration/pr-train. Re-requesting review when you have a slot.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `e20f62e8`

PR #2143, opened 2026-06-06
fix(broker-funding): retain reference to transfer status notification task

anshul23102:
## Summary

Transfer status push notifications are started with loop.create_task but the task reference is never retained, so they can be garbage collected before they finish. This retains the reference until completion and adds tests.

## What the bug is

BrokerFundingService._queue_transfer_status_notification_if_needed runs loop.create_task(self._send_transfer_status_notification(...)) and never stores the task. The asyncio event loop keeps only a weak reference to tasks created this way, so the notification can be garbage collected mid execution while it awaits its push delivery, silently dropping the message that tells a user their money movement completed, failed, returned, or canceled.

## Where it lives

hushh_mcp/services/broker_funding_service.py, in _queue_transfer_status_notification_if_needed.

## What the fix does

- Retains the task in a module-level set and removes it in the done callback, so the loop cannot collect it while it runs.
- Behavior is otherwise unchanged. The existing no-loop guard and status filtering are preserved.

## Canonical attach point and reachability

The changed method is reached from the live funding route api/routes/kai/plaid.py POST /plaid/transfers/create, which calls get_broker_funding_service().create_transfer, which calls _queue_transfer_status_notification_if_needed. This is a backend service behavior change under the existing funding route contract; no route signature, request model, or response shape changes.

## Path to merge

- Base: integration/pr-train
- Branch: fix/broker-funding-transfer-notification-task-ref
- Built on current base, no conflicts, all required checks green
- Blocker: maintainer re-review after the requested changes are addressed

## Testing

```
cd consent-protocol
.venv/bin/python -m pytest tests/test_broker_funding_transfer_notification_task_ref.py -v
```

Three tests drive the real method on the service singleton and assert a strong reference is held while a notifiable status change runs, the notification runs to completion, the reference is released afterward, and a non-notifiable status change schedules nothing. The wider broker funding suite stays green (34 passed). Ruff is clean on both files.


kushaltrivedi5 (changes requested):
## Changes Requested: Backend

### Direction

### Blocker

### Path To Merge

### Proof Needed

anshul23102:
@kushaltrivedi5 All changes completed and all checks passing. Broker funding transfer status notification task reference retained. Ready for review and merge.

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`consent-protocol/hushh_mcp/services/broker_funding_service.py`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `consent-protocol/hushh_mcp/services/broker_funding_service.py` and push to this branch.

### Proof Needed

anshul23102:
Thanks for the review. I have addressed the requested changes and tightened the scope.

Accepted value: the funding transfer status push could be garbage collected mid execution because loop.create_task was called without retaining the task reference, so a user could miss the message that their transfer completed, failed, returned, or canceled.

Canonical attach point: the changed method _queue_transfer_status_notification_if_needed is reached only from the live funding route api/routes/kai/plaid.py POST /plaid/transfers/create through get_broker_funding_service().create_transfer. This is a backend service behavior change under the existing funding route contract. No route signature, request model, response shape, or caller semantics change, so it does not introduce a parallel runtime or a new contract surface.

Scope: two files only, the service method plus its focused test. The PR body no longer claims any unrelated surface.

Proof command:
cd consent-protocol && .venv/bin/python -m pytest tests/test_broker_funding_transfer_notification_task_ref.py -v

All required checks are green and the branch has no conflicts with integration/pr-train. Re-requesting review when you have a slot.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `31c1a1e2`

PR #2162, opened 2026-06-06
fix(domain-registry): use timezone-aware UTC for cache freshness timer

anshul23102:
## Summary

DomainRegistryService tracked its in-memory cache freshness with the deprecated naive datetime.utcnow(). This switches all four call sites to datetime.now(timezone.utc) and adds tests.

## What the bug is

The service set self._cache_time on each load and computed elapsed time in _is_cache_valid by subtracting _cache_time from another datetime.utcnow() value. datetime.utcnow() is deprecated and returns a naive datetime. The setter and the elapsed computation must stay on the same awareness, otherwise the subtraction raises TypeError.

## Where it lives

consent-protocol/hushh_mcp/services/domain_registry_service.py:
- line 78: elapsed computation in _is_cache_valid
- lines 350, 384, 443: _cache_time assignments on load

## What the fix does

- Replaces all four datetime.utcnow() calls with datetime.now(timezone.utc) and imports timezone.
- The setter and the elapsed math move together, so the subtraction stays valid and the 300 second TTL behavior is unchanged.
- _cache_time stays None until the first load, and _is_cache_valid still returns False in that state.

## Reachability

api/routes/pkm.py and api/routes/pkm_routes_shared.py use DomainRegistryService. Its public methods get_domain, list_domains, and register_domain each call _is_cache_valid before serving cached domains, so the changed timing code runs on real PKM domain reads.

## How to verify

cd consent-protocol
.venv/bin/python -m pytest tests/test_domain_registry_cache_utc.py -v

Four tests pass. They drive the real _is_cache_valid method on a real service instance and assert a fresh tz-aware cache time is valid without raising, an expired one is invalid, the never-loaded state is invalid, and the module no longer references utcnow. The wider domain suite stays green (333 passed).

## Path to merge

- Base: integration/pr-train (direct PRs into main are blocked by repo policy)
- Branch: fix/domain-registry-utc-aware-cache-time
- Built on current main, no conflicts
- Blocker: requires CI Status Gate green and one maintainer approval
- Expected action: review of the timezone-aware swap and the PKM reachability trace

### DIFF
--- a/consent-protocol/hushh_mcp/services/domain_registry_service.py
+++ b/consent-protocol/hushh_mcp/services/domain_registry_service.py
@@ -16,7 +16,7 @@ Attach points:
 
 import logging
 from dataclasses import dataclass
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Optional
 
 from db.db_client import get_db
@@ -75,7 +75,7 @@ class DomainRegistryService:
         """Check if cache is still valid."""
         if self._cache_time is None:
             return False
-        elapsed = (datetime.utcnow() - self._cache_time).total_seconds()
+        elapsed = (datetime.now(timezone.utc) - self._cache_time).total_seconds()
         return elapsed < self._cache_ttl
 
     def _invalidate_cache(self):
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Backend

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2233, opened 2026-06-07
chore: add editorconfig for consistent formatting

gracyezhilkumar07:
## Summary
This PR improves authentication security by ensuring that all existing refresh tokens are invalidated when a user resets their password. This prevents old sessions from remaining active after credential changes.

## Changes Made
- Added refresh token invalidation during password reset flow
- Ensured all active sessions are revoked after password change
- Strengthened session security consistency across authentication system

## Security Impact
- Prevents reuse of stolen or leaked refresh tokens after password reset
- Forces all devices to re-authenticate after password change
- Reduces risk of session hijacking

## Testing Done
- Verified old refresh tokens are rejected after password reset
- Confirmed new login works normally after reset
- Checked no impact on unrelated authentication flows

## Notes
- Assumes refresh tokens are stored server-side or tracked via session mapping
- If using JWT-only auth, token versioning should be used instead


kushaltrivedi5 (approved):
Autonomy Confidence Gate: HIGH band. Path-clean (no sensitive family), tiny non-sensitive prod change (.editorconfig formatting, +10/-0), green CI Status Gate, MERGEABLE, verified at head a73dc4cf7d.

PR #2403, opened 2026-06-11
test(consent): cover protocol-relative route safety

gracyezhilkumar07:
Summary
Adds regression coverage for consent navigation route handling to ensure protocol-relative external URLs are not misclassified as internal SPA routes.
Changes
Added a consent route safety test to __tests__/utils/consent-sheet-route.test.ts
Verifies that protocol-relative external URLs (e.g. //evil.example.com/...) remain classified as external navigation targets
Helps prevent accidental internal route normalization of external consent links
Why
Consent navigation helpers are responsible for determining whether a URL should be handled internally or opened as an external destination. This test protects against future regressions where protocol-relative URLs could be incorrectly treated as trusted internal routes.


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2405, opened 2026-06-11
test(cache): cover listener unsubscribe behavior

gracyezhilkumar07:
## Summary

Adds regression coverage for cache listener lifecycle behavior by verifying that unsubscribed listeners no longer receive cache events.

## Changes

* Added a test for `CacheService.subscribe()`
* Verifies that the returned unsubscribe callback removes the listener
* Confirms cache events are only received while the listener is actively subscribed

## Why

`CacheService` exposes a subscription API used by consumers to react to cache lifecycle events. This test protects against regressions where unsubscribed listeners could continue receiving events, causing duplicate updates, memory leaks, or unexpected side effects.

## Testing

```bash
npm test -- cache-service
```


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2408, opened 2026-06-11
test(utils): preserve empty timeout override fallback

smirthi-dharma:
## Motivation

The slow request timeout resolver accepts environment-based timeout overrides. Empty-string override values should not bypass the existing timeout resolution contract or replace valid defaults.

This regression coverage protects empty override handling and preserves timeout fallback behavior.

## What Changed

Added regression coverage in:

`__tests__/utils/request-timeouts.test.ts`

## Covered Scenario

* preserves empty timeout override fallback
* verifies empty override values are ignored
* confirms valid default timeout values remain unchanged
* protects timeout configuration stability
* prevents regressions in environment override parsing

## Validation

```bash
npm run test -- request-timeouts
npm run lint
npm run typecheck
```

## Notes

* test-only change
* no production code modifications
* no runtime behavior changes
* DCO sign-off included


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2409, opened 2026-06-11
test(cache): preserve empty invalidateMany noop

gracyezhilkumar07:
## Summary

Adds regression coverage for `CacheService.invalidateMany()` to ensure empty key lists are treated as a no-op.

## Changes

* Added a test covering `invalidateMany([])`
* Verifies that no cache entries are removed when the key list is empty
* Confirms no cache invalidation events are emitted for empty operations

## Why

`invalidateMany()` explicitly returns early when provided an empty array. This test protects that contract and prevents future regressions that could trigger unnecessary cache mutations or listener notifications during empty invalidation flows.

## Testing

```bash
npm test -- cache-service
```


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2411, opened 2026-06-11
test(consent): preserve external query strings

gracyezhilkumar07:
## Summary

Adds regression coverage for consent navigation helpers to ensure external URLs with query parameters are preserved when resolving navigation targets.

## Changes

* Added a test covering external consent links that contain query strings
* Verifies that query parameters remain unchanged when classified as external navigation
* Protects against accidental URL normalization or query parameter loss

## Why

External consent and disclosure links may use query parameters for document state, localization, tracking, or section-specific rendering. This test ensures those parameters remain intact and external navigation behavior stays consistent.

## Testing

```bash
npm test -- consent-sheet-route
```


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2413, opened 2026-06-11
test(runtime): trim mixed URL whitespace

gracyezhilkumar07:
## Summary

Adds regression coverage for backend runtime URL sanitization when environment variables contain multiple whitespace character types.

## Changes

* Added a test covering backend URL values wrapped in spaces, tabs, newlines, and carriage returns
* Verifies runtime URL resolution consistently trims surrounding whitespace before validation
* Extends existing sanitization coverage with a mixed-whitespace input scenario

## Why

Configuration values may be copied from different environments, deployment systems, or CI pipelines that introduce multiple whitespace characters around URL values. This test ensures runtime URL resolution remains stable and platform-independent.

## Testing

```bash
npm test -- backend-runtime
```


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2462, opened 2026-06-12
fix(a11y): hide decorative phone verification icons

smirthi-dharma:
## Summary

* Hide decorative phone verification icons from assistive technologies
* Prevent screen readers from announcing non-informational visual elements
* Improve accessibility of the phone verification flow

## Changes

* Added `aria-hidden="true"` to the linked-phone status icon
* Added `aria-hidden="true"` to loading spinner shown during verification code send
* Added `aria-hidden="true"` to loading spinner shown during verification confirmation

## Testing

* npm run lint
* npm run typecheck


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2534, opened 2026-06-13
test(nav): ignore unsafe consent breadcrumb origins

gracyezhilkumar07:
## Summary

Add test coverage for consent breadcrumb navigation when an unsafe external `from` parameter is provided.

## Changes Made

- Added a test for external `from` URL values
- Verified unsafe origins are ignored during breadcrumb resolution
- Confirmed consent routes fall back to the default profile privacy workspace
- Extended navigation safety coverage for consent back navigation

## Test

```bash
npm test -- top-shell-breadcrumbs

### DIFF
--- a/hushh-webapp/__tests__/utils/top-shell-breadcrumbs.test.ts
+++ b/hushh-webapp/__tests__/utils/top-shell-breadcrumbs.test.ts
@@ -154,4 +154,20 @@ describe("top shell breadcrumbs", () => {
       ],
     });
   });
-});
+      it("ignores unsafe external from params for consent back navigation", () => {
+    const params = new URLSearchParams();
+    params.set("from", "https://evil.example.com/phish");
+
+    expect(resolveTopShellBreadcrumb("/consents", params)).toEqual({
+      backHref: "/profile?panel=access",
+      width: "profile",
+      align: "center",
+      items: [
+        { label: "Profile", href: "/profile?panel=access" },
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: test block has a malformed brace structure

### Direction
This is a genuinely valuable test — asserting that an unsafe external `from` param (e.g. `https://evil.example.com/phish`) is ignored and consent back-navigation falls back to the safe in-app route is exactly the kind of open-redirect guard we want covered. It imports and exercises the real `resolveTopShellBreadcrumb` helper, so it is proper production coverage, not a mock.

### Blocker
The new `it(...)` block is inserted with broken braces: the previous test's closing `});` is removed, the new case is nested one level too deep, and the file ends with a doubled `});` and no trailing newline. As written the test file no longer parses cleanly, which is why the Web (Next.js) gate cannot go green on this head.

### Path To Merge
Re-place the new `it("ignores unsafe external from params for consent back navigation", ...)` block *inside* the existing `describe("top shell breadcrumbs", () => { ... })` body, keep the prior test's `});` intact, fix the indentation to match the surrounding cases, and end the file with a single newline. Then push so the Web gate re-runs.

### Proof Needed
A green CI Status Gate (specifically the Web (Next.js) job) at the corrected head, showing the breadcrumb test suite — including this new open-redirect case — passing.

PR #2539, opened 2026-06-13
fix(investors): remove exception detail from investor proxy response

smirthi-dharma:
## Summary

Remove exception details from the investor proxy error response to avoid exposing internal implementation information to clients.

## Changes

* Removed `details: String(error)` from the investor proxy error response.
* Retained the existing generic error message:

  * `Failed to get investor`
* Kept server-side logging unchanged for debugging purposes.

## Why

Returning raw exception details in API responses can unintentionally expose internal application behavior and implementation details. Clients only need a generic failure response, while detailed diagnostics should remain on the server.

## Testing

* npm run lint
* npm run typecheck

### DIFF
--- a/hushh-webapp/app/api/investors/[id]/route.ts
+++ b/hushh-webapp/app/api/investors/[id]/route.ts
@@ -41,8 +41,9 @@ export async function GET(
     return NextResponse.json(data, { status: response.status });
   } catch (error) {
     console.error("[Investors Proxy] Get error:", error);
+
     return NextResponse.json(
-      { error: "Failed to get investor", details: String(error) },
+      { error: "Failed to get investor" },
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`hushh-webapp/app/api/investors/[id]/route.ts`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `hushh-webapp/app/api/investors/[id]/route.ts` and push to this branch.

### Proof Needed

PR #2550, opened 2026-06-13
fix(iam): suppress proxy exception detail in error response

smirthi-dharma:
## What

Removes exception-derived detail text from IAM proxy error responses.

## Why

Returning raw exception messages can expose internal implementation details to API consumers.

Using a generic error detail preserves the response contract while reducing information disclosure.

## Changes

* Replaced exception-derived `detail` value with a generic `"unknown_error"` response.

## Validation

* npm run lint
* npm run typecheck

### DIFF
--- a/hushh-webapp/app/api/iam/[...path]/route.ts
+++ b/hushh-webapp/app/api/iam/[...path]/route.ts
@@ -140,7 +140,7 @@ async function proxyRequest(
       requestId,
       {
         error: "Failed to proxy IAM request",
-        detail: error instanceof Error ? error.message : "unknown_error",
+        detail: "unknown_error",
       },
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`hushh-webapp/app/api/iam/[...path]/route.ts`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `hushh-webapp/app/api/iam/[...path]/route.ts` and push to this branch.

### Proof Needed

PR #2576, opened 2026-06-13
fix(consent-inbox-dropdown): add aria-haspopup and aria-expanded to trigger button

imsharukhan:
## Summary

Adds `aria-haspopup="menu"` and `aria-expanded={open}` to the ConsentInboxDropdown trigger button.

## Why

The trigger already has an accessible label but does not expose its popup relationship or expanded state to assistive technologies.

Adding these attributes completes the standard menu-button accessibility contract and reflects the existing dropdown state without introducing new logic.

## Changes

- Added `aria-haspopup="menu"`
- Added `aria-expanded={open}`

## Risk

- Single file change
- No visual changes
- No behavior changes
- Uses existing `open` state already controlling `DropdownMenu`
- Additive only

### DIFF
--- a/hushh-webapp/components/consent/consent-inbox-dropdown.tsx
+++ b/hushh-webapp/components/consent/consent-inbox-dropdown.tsx
@@ -222,6 +222,8 @@ export function ConsentInboxDropdown({
             type="button"
             className={triggerClassName}
             aria-label="Open consent inbox"
+            aria-haspopup="menu"
+            aria-expanded={open}
           >
             <Shield className="h-5 w-5" />
             {pendingCount > 0 ? (


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`consent-protocol/api/routes/consent.py`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `consent-protocol/api/routes/consent.py` and push to this branch.

### Proof Needed

PR #2579, opened 2026-06-13
fix(drawer): add showCloseButton parity with Dialog and Sheet

imsharukhan:
## Summary

Adds an optional `showCloseButton` prop to `DrawerContent`, matching the API already supported by `DialogContent` and `SheetContent`.

## Problem

`DialogContent` and `SheetContent` both expose a `showCloseButton` prop, while `DrawerContent` does not. This creates an API parity gap between the application's overlay primitives and prevents consumers from using a consistent prop contract when swapping between Dialog, Sheet, and Drawer surfaces.

## Change

* Added `showCloseButton?: boolean` to `DrawerContent`
* Added a conditional `DrawerPrimitive.Close` button when the prop is enabled
* Imported `XIcon` from `lucide-react`

The prop defaults to `false` to preserve all existing visual output and behavior.

## Why This Is Safe

* No visual changes for existing call sites
* No behavior changes unless the prop is explicitly enabled
* No architecture changes
* Follows the existing close-button implementation pattern already used in `SheetContent`

## Risk

* Single shared primitive file
* Additive-only API extension
* Backward compatible
* Existing drawers render exactly as before unless `showCloseButton` is provided

### DIFF
--- a/hushh-webapp/components/ui/drawer.tsx
+++ b/hushh-webapp/components/ui/drawer.tsx
@@ -1,6 +1,7 @@
 "use client"
 
 import * as React from "react"
+import { XIcon } from "lucide-react"
 import { Drawer as DrawerPrimitive } from "vaul"
 
 import { cn } from "@/lib/utils"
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `6d0bcc50`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #2580, opened 2026-06-13
fix(page-sections): add SectionHeader id passthrough

imsharukhan:
## Summary

Adds an optional `id` prop to `SectionHeader` and forwards it to the component's root element.

## Why

`SectionHeader` already forwards `testId` to its root container but does not expose a way to set a standard DOM `id`.

Adding an optional `id` prop enables:

* In-page anchor navigation (`#section-id`)
* `aria-labelledby` and related accessibility relationships
* Scroll-to-section behavior without requiring wrapper elements

## Changes

* Add `id?: string` to `SectionHeader` props
* Forward `id` to the root `<div>`

## Risk

* Additive only
* No visual changes
* No behavioral changes
* No impact on existing callers
* Undefined `id` remains a no-op

### DIFF
--- a/hushh-webapp/components/app-ui/page-sections.tsx
+++ b/hushh-webapp/components/app-ui/page-sections.tsx
@@ -247,6 +247,7 @@ export function SectionHeader({
   accent = "default",
   className,
   testId = "section-header",
+  id,
 }: {
   eyebrow?: string;
   title: ReactNode;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `7a89a396`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #2582, opened 2026-06-13
fix(vault): remove backend detail from wrapper delete error response

smirthi-dharma:
## Summary

Remove backend error details from the vault wrapper delete API response.

## What Changed

* Removed propagation of upstream error payloads from the vault wrapper delete endpoint.
* Replaced the response with a generic `Backend error` message.
* Preserved existing HTTP status code behavior.

## Why

Returning upstream error payloads can expose internal implementation details and diagnostic information to clients. This change hardens the API surface by ensuring only a generic error response is returned while keeping detailed diagnostics on the server side.

## Testing

* `npm run lint`
* `npm run typecheck`

## Risk

Low. The change only affects the error payload returned when the upstream vault wrapper delete request fails.

### DIFF
--- a/hushh-webapp/app/api/vault/wrapper/delete/route.ts
+++ b/hushh-webapp/app/api/vault/wrapper/delete/route.ts
@@ -84,10 +84,10 @@ export async function POST(request: NextRequest) {
     });
 
     if (!response.ok) {
-      const errorPayload = await response
-        .json()
-        .catch(async () => ({ error: await response.text().catch(() => "") }));
-      return NextResponse.json(errorPayload, { status: response.status });
+      return NextResponse.json(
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`hushh-webapp/app/api/vault/wrapper/delete/route.ts`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `hushh-webapp/app/api/vault/wrapper/delete/route.ts` and push to this branch.

### Proof Needed

PR #2591, opened 2026-06-13
test(config): add environment contract coverage

imsharukhan:
## Summary

Adds a static environment contract test that verifies `.env.example` documents the required environment variable keys and expected default values used by the application.

### Firebase Keys

* NEXT_PUBLIC_FIREBASE_API_KEY
* NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN
* NEXT_PUBLIC_FIREBASE_PROJECT_ID
* NEXT_PUBLIC_FIREBASE_STORAGE_BUCKET
* NEXT_PUBLIC_FIREBASE_MESSAGING_SENDER_ID
* NEXT_PUBLIC_FIREBASE_APP_ID
* NEXT_PUBLIC_FIREBASE_VAPID_KEY
* NEXT_PUBLIC_FIREBASE_MEASUREMENT_ID

### Runtime Keys

* NEXT_PUBLIC_APP_ENV
* NEXT_PUBLIC_CLIENT_VERSION
* APP_RUNTIME_PROFILE
* NEXT_PUBLIC_BACKEND_URL
* NEXT_PUBLIC_APP_URL

### Observability Keys

* NEXT_PUBLIC_OBSERVABILITY_ENABLED
* NEXT_PUBLIC_OBSERVABILITY_DEBUG
* NEXT_PUBLIC_OBSERVABILITY_SAMPLE_RATE
* NEXT_PUBLIC_GTM_ID

### Default Value Contracts

* APP_RUNTIME_PROFILE=local
* NEXT_PUBLIC_APP_ENV=development
* NEXT_PUBLIC_OBSERVABILITY_ENABLED=true
* NEXT_PUBLIC_OBSERVABILITY_DEBUG=false
* NEXT_PUBLIC_OBSERVABILITY_SAMPLE_RATE=1

## Why

This test prevents configuration drift between application requirements and `.env.example`.

If a required environment variable is introduced but not documented, the test fails and surfaces the issue during CI rather than after deployment or local setup.

## Implementation

* Reads `.env.example` using `readFileSync`
* Performs static string assertions only
* Imports no application code
* Uses no mocks
* Does not read runtime environment variables

## Risk

* New test file only
* No production code changes
* No runtime behavior changes
* No configuration changes

## Files Changed

* `hushh-webapp/__tests__/lib/env-contract.test.ts`

### DIFF
--- a/hushh-webapp/__tests__/lib/env-contract.test.ts
+++ b/hushh-webapp/__tests__/lib/env-contract.test.ts
@@ -1,0 +1,92 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, it } from "vitest";
+
+/**
+ * Environment contract test.
+ *
+ * Verifies that .env.example documents every environment variable key
+ * the application depends on. Prevents silent misconfigurations when
+ * required env vars are added without updating the canonical example file.
+ *
+ * Pattern: static source file inspection — no app code imported,
+ * no runtime env values read. Mirrors register-phone-shell-contract pattern.
+ */
+describe("environment contract (.env.example)", () => {
+  const envExample = readFileSync(
+    join(process.cwd(), ".env.example"),
+    "utf8"
+  );
+
+  // ── Firebase ────────────────────────────────────────────────────────────────
+
+  const REQUIRED_FIREBASE_KEYS = [
+    "NEXT_PUBLIC_FIREBASE_API_KEY",
+    "NEXT_PUBLIC_FIREBASE_AUTH_DOMAIN",
+    "NEXT_PUBLIC_FIREBASE_PROJECT_ID",
+    "NEXT_PUBLIC_FIREBASE_STORAGE_BUCKET",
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: test must exercise production code

### Direction
Thank you for investing in test coverage — guarding env/contract drift and UI primitives with tests is exactly the kind of hygiene this codebase rewards, and the instinct to lock these invariants down is the right one.

### Blocker
This test does not import or execute any production module — it inspects source text statically (or asserts against locally-defined fixtures). A frontend/test-only change is only proof of an app contract when it imports the real production code path and exercises it; a static source scan can pass even when the real runtime regresses, so it cannot gate merge as contract coverage.

### Path To Merge
Import the actual production module/component under test (via the `@/...` alias or a relative path), call its real exported function/render its real component, and assert on that real behavior instead of on file contents or an in-test mock. Keep the same intent — just point the assertions at the shipped code.

### Proof Needed
The test imports a production module, exercises it, and fails if that production behavior regresses, with a green CI Status Gate at the updated head.

PR #2610, opened 2026-06-14
fix(tickers): cap search result limit

smirthi-dharma:
## Summary

Add bounds checking for the ticker search limit parameter.

## What Changed

* Parse the `limit` query parameter as a number.
* Enforce a minimum value of `1`.
* Enforce a maximum value of `100`.
* Preserve the default value when invalid input is provided.

## Why

The ticker search endpoint previously forwarded a user-controlled limit value without validation. Excessively large values can increase backend workload and resource consumption. Applying reasonable bounds improves resilience and aligns with existing query parameter hardening patterns.

## Testing

* `npm run lint`
* `npm run typecheck`

## Risk

Low. Only affects requests providing invalid, negative, or excessively large limit values. Existing valid requests continue to behave as expected.

### DIFF
--- a/hushh-webapp/app/api/tickers/search/route.ts
+++ b/hushh-webapp/app/api/tickers/search/route.ts
@@ -25,7 +25,10 @@ export async function GET(request: NextRequest) {
   try {
     const { searchParams } = new URL(request.url);
     const q = searchParams.get("q") || "";
-    const limit = searchParams.get("limit") || "25";
+    const rawLimit = Number.parseInt(searchParams.get("limit") || "25", 10);
+    const limit = Number.isFinite(rawLimit)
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`hushh-webapp/app/api/tickers/search/route.ts`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `hushh-webapp/app/api/tickers/search/route.ts` and push to this branch.

### Proof Needed

PR #2616, opened 2026-06-14
fix(account): bound proxy path parameters

smirthi-dharma:
## Summary

Add bounds validation for account proxy path parameters.

## What Changed

* Reject empty account proxy paths.
* Reject paths with more than 8 segments.
* Reject individual path segments longer than 128 characters.
* Return a `400` response for invalid account API paths before constructing the upstream request URL.

## Why

The account proxy previously accepted arbitrary path depth and segment lengths. Adding validation helps prevent excessively large path inputs from reaching the upstream proxy layer and aligns with existing route hardening efforts.

## Testing

* `npm run lint`
* `npm run typecheck`

## Risk

Low. Valid account API paths continue to function normally. Only malformed or excessively large path inputs are rejected.

### DIFF
--- a/hushh-webapp/app/api/account/[...path]/route.ts
+++ b/hushh-webapp/app/api/account/[...path]/route.ts
@@ -32,6 +32,18 @@ function isUpstreamTimeoutError(error: unknown): boolean {
 
 async function proxyRequest(request: NextRequest, params: { path: string[] }) {
   const requestId = resolveRequestId(request);
+
+  if (
+    params.path.length === 0 ||
+    params.path.length > 8 ||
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`hushh-webapp/app/api/account/[...path]/route.ts`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `hushh-webapp/app/api/account/[...path]/route.ts` and push to this branch.

### Proof Needed

PR #2620, opened 2026-06-14
Test/auth legal dialog close button

gracyezhilkumar07:
## Summary

Adds test coverage for dialog description rendering behavior.

### Changes Made

* Added a test to verify `DialogDescription` content is rendered correctly within `DialogContent`.
* Ensured dialog descriptive text remains accessible and visible to users.
* Extended existing dialog component test coverage.

## Test

```bash
npx vitest run __tests__/ui/dialog.test.tsx
```

### DIFF
--- a/hushh-webapp/__tests__/ui/dialog.test.tsx
+++ b/hushh-webapp/__tests__/ui/dialog.test.tsx
@@ -1,0 +1,34 @@
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+
+import {
+  Dialog,
+  DialogContent,
+  DialogTitle,
+} from "@/components/ui/dialog";
+
+describe("DialogContent", () => {
+  it("renders the close button by default", () => {
+    render(
+      <Dialog open>
+        <DialogContent>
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: test must exercise production code

### Direction
Thank you for investing in test coverage — guarding env/contract drift and UI primitives with tests is exactly the kind of hygiene this codebase rewards, and the instinct to lock these invariants down is the right one.

### Blocker
This test does not import or execute any production module — it inspects source text statically (or asserts against locally-defined fixtures). A frontend/test-only change is only proof of an app contract when it imports the real production code path and exercises it; a static source scan can pass even when the real runtime regresses, so it cannot gate merge as contract coverage.

### Path To Merge
Import the actual production module/component under test (via the `@/...` alias or a relative path), call its real exported function/render its real component, and assert on that real behavior instead of on file contents or an in-test mock. Keep the same intent — just point the assertions at the shipped code.

### Proof Needed
The test imports a production module, exercises it, and fails if that production behavior regresses, with a green CI Status Gate at the updated head.

PR #1431, opened 2026-05-19
fix(privacy): redact user_id and token value from consent system logs

M-Rajith:
## What this fixes
Closes #1430

## Problem
Two files leaked sensitive identifiers into application logs:

1. hushh_mcp/consent/token.py (line 344)
   token_str[:30] was logged in a warning — even truncated,
   token prefixes can correlate tokens across log entries in
   GCP Logging or any log aggregation system.

2. api/routes/sse.py (lines 159, 189, 215, 217)
   user_id logged in plaintext across all 4 SSE lifecycle
   events (open, disconnect, cancel, error).

For a zero-knowledge, consent-first system these identifiers
must never appear in logs.

### consent/token.py
- Replaced f-string token prefix with static [token redacted]

### api/routes/sse.py
- Replaced user_id=%s with user_id=[redacted] in all 4 calls

### tests/privacy/test_pii_redaction_in_logs.py
- 4 tests verifying redaction in token and SSE log paths

## Validation
- All existing Protocol CI tests passing
- 4 new privacy tests added
- Signed-off-by: Madana Rajith <rajithmadana@gmail.com>

PR #2677, opened 2026-06-15
fix(privacy): redact token value from consent token warning logs

M-Rajith:
Removes raw token string from the in-memory revocation warning in token.py.

Logging-only change — no effect on return values or API contracts.
No caller changes needed.

redact_log_value is imported from mcp_modules.log_redaction, which already
exists on integration/pr-train and lists user_id in _SENSITIVE_EXACT_KEYS.
No new dependency introduced.

Closes part of #1431 (split for surface narrowing).


kushaltrivedi5 (changes requested):

### Direction

### Blocker
This PR touches a trust-boundary / sensitive runtime family (consent / IAM / KYC trust boundary, auth / session trust boundary). Changes here cannot auto-merge: they require an explicit human security review of the DB-backed enforcement, scope, caller-credential shape, and data-boundary impact, because UI or path-level changes here are not sufficient proof of safety.

### Path To Merge
A maintainer will run the sensitive-surface review for consent / IAM / KYC trust boundary, auth / session trust boundary and either land a bounded strengthening patch (co-authored to you) or return specific guard requirements. Keep the PR open; no rework is needed until that review returns concrete asks.

### Proof Needed

M-Rajith:
Closing — the base already uses _token_fingerprint(token_str) on line 377-378, which logs a 12-char SHA-256 digest instead of the raw token prefix. The fix this PR intended is already in integration/pr-train.

PR #2681, opened 2026-06-15
fix(security): bind TrustLink HMAC to session_id to prevent cross-stream replay

M-Rajith:
Includes session_id in the HMAC input string so a signature from Stream A cannot be replayed on Stream B. Depends on session_id field added in companion PR. 5 tests covering valid verify, cross-stream replay rejection, signature uniqueness, expiry, and backward compat.

### DIFF
--- a/consent-protocol/hushh_mcp/trust/link.py
+++ b/consent-protocol/hushh_mcp/trust/link.py
@@ -15,12 +15,22 @@ def create_trust_link(
     to_agent: AgentID,
     scope: ConsentScope,
     signed_by_user: UserID,
+    session_id: str = "",
     expires_in_ms: int = DEFAULT_TRUST_LINK_EXPIRY_MS,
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
This PR touches a trust-boundary / sensitive runtime family (consent / IAM / KYC trust boundary). Changes here cannot auto-merge: they require an explicit human security review of the DB-backed enforcement, scope, caller-credential shape, and data-boundary impact, because UI or path-level changes here are not sufficient proof of safety.

### Path To Merge
A maintainer will run the sensitive-surface review for consent / IAM / KYC trust boundary and either land a bounded strengthening patch (co-authored to you) or return specific guard requirements. Keep the PR open; no rework is needed until that review returns concrete asks.

### Proof Needed

PR #2734, opened 2026-06-16
fix(kai): prevent NaN allocation percentages from rendering

Divya-rajendran009:
## Summary
- Add a render-only allocation percentage formatter in Kai portfolio review.
- Show `0.0%` for NaN, Infinity, undefined, or null allocation percentages.
- Preserve the existing one-decimal allocation percentage display format.

## Root Cause
Allocation percentages were rendered with inline `.toFixed(1)` calls. Non-finite values such as `NaN` or `Infinity` could therefore appear directly in the UI instead of a safe fallback.

## Impact
Kai portfolio allocation rows now render a stable fallback value for invalid allocation percentages without changing portfolio calculations, API calls, state, routing, or styling.

## Validation
- `npm.cmd run lint` passed.
- `npm.cmd run build` passed.
- `npm.cmd run test` failed on an unrelated existing CRLF/LF literal assertion in `__tests__/components/top-app-bar.contract.test.ts`.

### DIFF
--- a/hushh-webapp/components/kai/views/portfolio-review-view.tsx
+++ b/hushh-webapp/components/kai/views/portfolio-review-view.tsx
@@ -266,6 +266,12 @@ function _formatPercent(value: number | undefined | null): string {
   return `${sign}${value.toFixed(2)}%`;
 }
 
+function formatAllocationPercent(value: number | undefined | null): string {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2745, opened 2026-06-16
fix(kai): hide decorative loading spinner in portfolio source switcher

Divya-rajendran009:
## Summary
- Mark portfolio and statement deletion loading spinners as decorative with `aria-hidden`.
- Keep the existing button text as the accessible action label.

## Why
The loading spinners are visual status affordances shown inside labeled buttons. Hiding them from assistive technology avoids decorative icon noise while preserving the delete action labels.

## Impact
Screen reader output stays focused on the button text. There is no visual or behavioral change.

## Validation
- Inspected staged diff: two spinner-only updates in `hushh-webapp/components/kai/portfolio-source-switcher.tsx`.
- Attempted focused ESLint with `npx.cmd eslint hushh-webapp/components/kai/portfolio-source-switcher.tsx --max-warnings=0`; blocked because the clean worktree has no installed `node_modules` and ESLint could not resolve `eslint-config-next`.

### DIFF
--- a/hushh-webapp/components/kai/portfolio-source-switcher.tsx
+++ b/hushh-webapp/components/kai/portfolio-source-switcher.tsx
@@ -180,7 +180,7 @@ export function PortfolioSourceSwitcher({
                 data-voice-control-id="delete_imported_data"
               >
                 {isDeletingPortfolio ? (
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2766, opened 2026-06-16
test(observability): add coverage for request ID and timestamp utilities

KARTHI-77:
## Summary

Adds a dedicated test suite for the observability request ID utilities.

### What was added

* Tests for `sanitizeRequestId`

  * accepts valid request IDs
  * trims whitespace from valid IDs
  * rejects invalid values

* Tests for `getOrCreateRequestId`

  * reuses valid request IDs from headers
  * generates a replacement when header values are invalid

* Tests for `validateHeaderTimestampConstraints`

  * accepts valid timestamps
  * rejects invalid numeric values
  * rejects timestamps that exceed the allowed future clock drift

* Tests for `getOrCreateRequestTimestampMs`

  * returns valid header timestamps
  * falls back to `nowMs` when timestamps are invalid

### Validation

```bash
npx vitest run __tests__/services/observability-request-id.test.ts
```

Result: 10 tests passed.

### DIFF
--- a/hushh-webapp/__tests__/services/observability-request-id.test.ts
+++ b/hushh-webapp/__tests__/services/observability-request-id.test.ts
@@ -1,0 +1,108 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  FUTURE_TIMESTAMP_ERROR,
+  INVALID_TIMESTAMP_ERROR,
+  REQUEST_ID_HEADER,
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #2773, opened 2026-06-16
test(hooks): preserve debounced value when rerendered with same reference

Divya-rajendran009:
## Summary
- Add a useDebouncedValue test for rerendering with the same object reference.
- Verify no extra debounced commit fires when the value reference does not change.

## Why
The hook already covers new object references. This adds same-reference coverage so rerenders do not accidentally schedule extra updates.

## Impact
Test-only change. No runtime behavior changes.

## Validation
- `npx.cmd vitest run __tests__/hooks/use-debounced-value.test.tsx`


Divya-rajendran009:
Closing as superseded by #2774, which now consolidates both hook test cases into a single PR to avoid shared-file overlap risk.

PR #2867, opened 2026-06-17
fix(kai-chat): use UTC-aware datetime for analyzed_at PKM writes

anshul23102:
## Summary

- `hushh_mcp/services/kai_chat_service.py` line 1230: `datetime.now()` produces a naive (timezone-unaware) ISO string stored in PKM under `analyzed_at` and `analysis_last_updated`
- Naive datetimes are ambiguous across DST boundaries and break any downstream timezone-aware comparison
- Replaces `datetime.now()` with `datetime.now(UTC)` and adds `UTC` to the import

## Test plan

- [ ] `tests/test_kai_chat_analyzed_at_utc.py` -- 3 source-level assertions:
  - `UTC` is imported from `datetime`
  - `datetime.now(UTC)` is present in the file
  - No bare `datetime.now()` calls remain
- All 3 pass against the fix; all 3 would fail against `main`

### DIFF
--- a/consent-protocol/hushh_mcp/services/kai_chat_service.py
+++ b/consent-protocol/hushh_mcp/services/kai_chat_service.py
@@ -23,7 +23,7 @@ import logging
 import os
 import re
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import UTC, datetime
 from enum import Enum
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`consent-protocol/hushh_mcp/services/kai_chat_service.py`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `consent-protocol/hushh_mcp/services/kai_chat_service.py` and push to this branch.

### Proof Needed

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `faac0c59`

PR #2868, opened 2026-06-17
fix(portfolio): replace deprecated utcnow() with UTC-aware datetime.now(UTC)

anshul23102:
## Summary

- `hushh_mcp/services/portfolio_import_service.py` line 3646: `datetime.utcnow().isoformat()` is deprecated (Python 3.12) and returns a naive datetime with no timezone info embedded
- `hushh_mcp/services/portfolio_parser.py` line 63: `datetime.utcnow()` stores a naive datetime in `parsed_at`, which breaks any downstream timezone-aware comparison
- Both replaced with `datetime.now(UTC)` and `UTC` added to the respective import lines

## Test plan

- [ ] `tests/test_portfolio_utcnow_deprecated.py` -- 6 source-level assertions across both files:
  - `UTC` imported in each file
  - No `datetime.utcnow()` calls remain
  - `datetime.now(UTC)` call-sites present
- All 6 pass against the fix; all 6 would fail against `main`

### DIFF
--- a/consent-protocol/hushh_mcp/services/portfolio_import_service.py
+++ b/consent-protocol/hushh_mcp/services/portfolio_import_service.py
@@ -14,7 +14,7 @@ import json
 import logging
 import re
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import Any, Optional
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`consent-protocol/hushh_mcp/services/portfolio_import_service.py`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `consent-protocol/hushh_mcp/services/portfolio_import_service.py` and push to this branch.

### Proof Needed

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `5655776e`

PR #2921, opened 2026-06-17
fix(utils): add explicit unknown type to catch clauses in session-storage

imsharukhan:
## Summary

Adds explicit `unknown` annotations to catch variables in `session-storage.ts`.

## What

Replaces:

```ts
catch (e) {
```

with:

```ts
catch (e: unknown) {
```

throughout the file.

## Why

The repository commonly uses explicit `unknown` catch variables to document error contracts and prevent accidental use of untyped caught values.

This change aligns `session-storage.ts` with that pattern.

## Impact

* No runtime behavior changes
* No logic changes
* No API changes
* Existing logging behavior preserved
* Single-file change

## Checklist

* [x] Type-only change
* [x] Single file
* [x] No behavior changes
* [x] TypeScript clean

### DIFF
--- a/hushh-webapp/lib/utils/session-storage.ts
+++ b/hushh-webapp/lib/utils/session-storage.ts
@@ -51,7 +51,7 @@ function purgeStaleSessionKeysOnNative(): void {
     }
 
     storage.setItem(BOOT_ID_KEY, currentBootId);
-  } catch (e) {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #3046, opened 2026-06-18
types(session-storage): replace catch(e) with catch(unknown)

imsharukhan:
## Summary

Adds explicit `unknown` typing to catch variables in `lib/utils/session-storage.ts`.

## What Changed

Replaced:

```ts
catch (e) {
```

with:

```ts
catch (e: unknown) {
```

across all catch blocks in the file.

## Why

TypeScript treats caught values as unknown at runtime.

Using an explicit `unknown` annotation makes that contract visible in the source code and aligns the file with the project's type-safety conventions.

## Safety

All catch blocks currently pass the caught value directly to `console.warn(...)`.

No narrowing is required because `console.warn` accepts values of any type.

Runtime behavior remains identical.

## Scope

Affected functions include:

* purgeStaleSessionKeysOnNative
* getSessionLikeStorage
* getPersistentStorage
* setSessionItem
* getSessionItem
* removeSessionItem
* removeSessionItemsByPrefix
* clearSessionStorage
* setLocalItem
* getLocalItem
* removeLocalItem
* clearLocalStorage

## Risk

* No runtime changes
* No logic changes
* No API changes
* No import changes
* No behavior changes

Type-safety improvement only.

## Validation

* TypeScript compilation passes
* No remaining untyped `catch (e)` blocks in the file
* Logging behavior unchanged

### DIFF
--- a/hushh-webapp/lib/utils/session-storage.ts
+++ b/hushh-webapp/lib/utils/session-storage.ts
@@ -51,7 +51,7 @@ function purgeStaleSessionKeysOnNative(): void {
     }
 
     storage.setItem(BOOT_ID_KEY, currentBootId);
-  } catch (e) {
+  } catch (e: unknown) {
     console.warn("[SessionStorage] Boot ID check failed:", e);
   }
 }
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #3056, opened 2026-06-18
fix(storage): add defensive storage fallback

abdulbeast4-creator:
## Overview
This PR hardens the existing web storage utility used by local cache callers against strict browser storage failures. It keeps normal localStorage behavior intact while adding silent fallback handling for disabled storage access and quota exhaustion.

## Impact
Low risk: the change is attached to the existing localized storage helper, adds defensive catch paths only, and does not alter server, schema, route, or auth behavior. High impact: the UI can continue running when localStorage is blocked in strict privacy/incognito modes or when persistent storage writes exceed browser quota.

## Changes Cleared
- Added a memory fallback for persistent local cache values when localStorage is unavailable.
- Added QuotaExceededError detection for storage limit failures.
- Silenced local persistent storage access/write/read/remove/clear failures to avoid console noise and runtime crashes.
- Preserved standard localStorage reads and writes when browser storage is healthy.
- Added focused tests for normal storage, quota failure, blocked storage access, and fallback clearing.

## Impact Map
- Routes: none.
- API/schema: none.
- Storage/database: browser local cache utility only; no backend persistence changes.
- Trust/privacy boundary: no new collected fields or user identifiers.
- Runtime: localStorage failure paths now degrade to in-memory state.

## Verification
- npm run test:ci -- __tests__/utils/session-storage-defensive.test.ts
- npm run typecheck


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

kushaltrivedi5 (changes requested):
## Changes Requested

PR #3127, opened 2026-06-19
test(use-debounced-value): cover null value debouncing

Divya-rajendran009:
## Summary
- Add focused hook test coverage for null value debouncing.
- Verify `useDebouncedValue` waits for the delay before committing a transition from a string value to `null`.

## Scope Proof
- Test file touched: `hushh-webapp/__tests__/hooks/use-debounced-value.test.tsx`.
- No runtime hook code changed.
- Appends one new `it()` block inside the existing `describe("useDebouncedValue")` suite.

## Impact
Test-only change. This locks the existing null transition debounce behavior without changing application output.

## Validation
- `npx.cmd vitest run __tests__/hooks/use-debounced-value.test.tsx`

### DIFF
--- a/hushh-webapp/__tests__/hooks/use-debounced-value.test.tsx
+++ b/hushh-webapp/__tests__/hooks/use-debounced-value.test.tsx
@@ -194,4 +194,27 @@ describe("useDebouncedValue", () => {
 
     expect(onDebounced).toHaveBeenLastCalledWith(next);
   });
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `0df0d81a`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3155, opened 2026-06-19
test(api-client): cover empty message fallback

gracyezhilkumar07:
## Summary

Adds test coverage for API error responses where the `message` field is present but empty.

## Changes

* Added a test case for an empty string `message` field (`""`)
* Verifies that `apiJson` treats an empty message as unreadable content
* Confirms fallback to the default status-based error message
* Extends existing error handling edge-case coverage

## Test

```bash
npx vitest run __tests__/services/api-client.test.ts
```

Result:

* All tests passed
* No regressions introduced


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #3228, opened 2026-06-20
fix(ui): replace pointer-events-none with cursor-not-allowed in sidebar menu buttons for better a11y cursor feedback

Roopmann30:
This PR replaces the disabled:pointer-events-none class with disabled:cursor-not-allowed in sidebar menu buttons. The previous pointer-events-none utility prevented normal cursor behavior, which stopped users from seeing the expected not-allowed cursor when hovering over disabled actions. Using cursor-not-allowed restores standard visual feedback, improves accessibility, and provides a clearer indication that the menu button is disabled.


kushaltrivedi5 (approved):

Re-reviewed at head `c81703f5`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3381, opened 2026-06-23
test(config): cover whitespace-only runtime URL fallbacks

KARTHI-77:
## Summary

Adds coverage for whitespace-only runtime URL values in configuration resolution.

### Added tests

* Development backend URL falls back to the localhost default when the runtime backend URL contains only whitespace.
* Development frontend URL falls back to the localhost default when the runtime frontend URL contains only whitespace.

### Why

`normalizeUrl()` trims runtime URL values before evaluation. These tests verify that whitespace-only values are treated as empty values and correctly trigger the existing fallback behavior.

### DIFF
--- a/hushh-webapp/__tests__/lib/config.test.ts
+++ b/hushh-webapp/__tests__/lib/config.test.ts
@@ -49,6 +49,32 @@ describe("config", () => {
 
     expect(config.BACKEND_URL).toBe("http://127.0.0.1:8000");
   });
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `c9e7bb19`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3386, opened 2026-06-23
feat(iam): ZK device-to-device sharing — drop plaintext location twin, heal migration chain, per-grant HCT capability tokens

kushaltrivedi5:
## Summary
Establishes a zero-knowledge device-to-device sharing posture for live location so cross-device sharing rides the consent protocol as a standard, and cleans up the supporting IAM surfaces.

### Phase 1 — remove the plaintext location prototype (IAM Invariant #1)
- Delete the unmounted KAI location router + `KaiLocationService` + tests.
- Migration `069_drop_kai_location_plaintext.sql` drops all six `kai_location_*` tables (children before parents, idempotent; indexes cascade).
- Strip `kai_location_*` from `account_service` deletion order.

### Phase 2 — heal the migration chain (developer-token registry)
- `070_developer_registry.sql`: canonical versioned DDL for `developer_applications` / `developer_apps` / `developer_tokens`, byte-compatible with the runtime `ensure_tables()` safety net (incl. historical renames). New `developer` manifest lane; UAT schema contract bumped to v70 with the three developer tables (hash-only, never raw token).

### Phase 3 — per-grant HCT capability tokens (reuse existing token machinery, no bloat)
- `create_grant()` mints a signed HCT scoped `cap.location.live.view`, bound to `device:<recipient>`, expiring with the grant, stored in grant metadata.
- `store_encrypted_envelope()` cryptographically validates that token (signature, expiry, scope) before accepting ciphertext. Tokenless legacy grants fall back to DB checks, so the change is backward compatible.

### Phase 4 — docs + audit tooling
- IAM architecture, scope catalog, one-location-agent, data-plane contract updated for the ZK model + developer-registry storage boundary.
- `data_model_audit.py` now honors `DROP TABLE` / net-schema, so decommissioned prototype tables are not flagged unclassified.

## Verification (real runs)
- **151 targeted tests pass** (one_location, account-cleanup, migration-contract, developer-registry, developer-API, granular-scopes, token-timing).
- `verify_release_migration_contract.py` aligned at head **70**.
- `data_model_audit.py` EXIT 0, zero unclassified tables.

## Notes
- Scoped strictly to the ZK/IAM workstream; the concurrent `071_enterprise_crm_registry` work is intentionally excluded so it lands under its own authorship.
- DCO signed-off.


kushaltrivedi5:
Re-basing: this work is migration-chain-dependent on main's 068_one_location_circle_invites, which integration/pr-train does not yet carry (train head is 067). Targeting the train would force resolution of hundreds of unrelated train-vs-main drift conflicts and would regress the 069/070 migration ordering. Per the PR-base-policy SOP, re-opening against main via the pre-authorized promotion branch since the change is cleanly based on main and verified green (151 tests, both governance verifiers).

PR #3389, opened 2026-06-23
test(browser-navigation): cover replace navigation events

KARTHI-77:
## Summary

Adds test coverage for internal replace navigation behavior.

### Added coverage

Verifies that `replaceWindowLocation()` dispatches an internal navigation event when given a routable application URL.

### Why

`replaceWindowLocation()` has different behavior from `assignWindowLocation()` because it emits navigation requests with `replace: true`.

This test documents and protects that behavior against future regressions.


kushaltrivedi5 (approved):

Re-reviewed at head `c121314a`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3403, opened 2026-06-23
test(navigation): characterize normalizeInternalRouteHref guards

abdulbeast4-creator:
## Summary

Adds an isolated characterization test for `normalizeInternalRouteHref` and
`resolveInternalRouteHref` in `lib/navigation/routes.ts`. These helpers had no
direct unit coverage in `__tests__/navigation/routes.test.ts`.

The test documents the function's **actual, intentionally narrow** contract:

- trims surrounding whitespace
- returns `null` for nullish / empty input
- accepts single-slash internal paths verbatim (case and query preserved)
- rejects non-internal and protocol-relative `//` hrefs
- rejects hrefs containing CR/LF (header/script injection guard)
- `resolveInternalRouteHref` returns the normalized value or the fallback

## Scope correction (truth-first)

The original ask described "legal navigation route utilities" that "decode
uppercase Unicode anchor links to segments." That surface does not exist:
`normalizeInternalRouteHref` performs no Unicode decoding, no case
normalization, and no `#` fragment parsing, and `app/api/legal/[doc]/route.ts`
only proxies two hardcoded URLs. To avoid asserting behavior that isn't
implemented, this test pins the real guards instead.

## Verification

- `npm test -- __tests__/navigation/normalize-internal-route-href.test.ts` → 7/7 pass

## Notes

- Test-only surface change. No production runtime behavior is modified.
- Targets `integration/pr-train`.

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-internal-route-href.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-internal-route-href.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  normalizeInternalRouteHref,
+  resolveInternalRouteHref,
+} from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref / resolveInternalRouteHref
+// in lib/navigation/routes.ts. These pin the *actual* implemented guards:
+//   - trims surrounding whitespace
+//   - returns null for empty / nullish input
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `1d0cce24`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3440, opened 2026-06-23
test(ui): cover switch default checked state contract

gracyezhilkumar07:
## Summary

Adds test coverage for the Switch component's default checked state contract.

## Changes Made

* Added a test verifying that a Switch rendered with `defaultChecked` exposes the expected checked state.
* Validates that the component sets the correct `data-state` value when initialized as checked.
* Helps prevent regressions in Switch state behavior.

## Test

```bash
npx vitest run __tests__/ui/switch.test.tsx
```


kushaltrivedi5 (approved):

Re-reviewed at head `673ef3bc`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3465, opened 2026-06-24
fix(vault): enforce explicit button types in recovery key dialog

Roopmann30:
Adds explicit `type="button"` attributes to the action buttons ("Copy Key", "Download", "I've Saved My Recovery Key") in the `RecoveryKeyDialog` component.

**Why**: Without this attribute, HTML `<button>` elements default to `type="submit"`. If the vault layout or dialog overlay happens to be placed within a `<form>`, clicking these actions will inadvertently trigger a form submission. This brings the vault recovery flow in line with the safe button enforcement patterns already merged into the codebase.

### DIFF
--- a/hushh-webapp/components/vault/recovery-key-dialog.tsx
+++ b/hushh-webapp/components/vault/recovery-key-dialog.tsx
@@ -84,6 +84,7 @@ export function RecoveryKeyDialog({
 
           <div className="grid grid-cols-2 gap-2">
             <Button
+              type="button"
               onClick={handleCopy}
               className="w-full"
             >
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current blockers: frontend_component_without_reachable_caller, existing_capability_overlap_requires_review.

### Path To Merge
Maintainer patch is not permitted without a concrete attachment plan. Maintainer patch is denied because the PR contains standalone helper, component, export, agent, or runtime code without a proven canonical caller. Blocking findings: frontend_component_without_reachable_caller.

### Proof Needed

PR #3470, opened 2026-06-24
fix(profile): enforce explicit button types in runtime secret settings

Roopmann30:
Adds explicit `type="button"` attributes to the action buttons ("Save", "Remove") in the `RuntimeSecretSettingsCard` component.

**Why**: Without this attribute, HTML `<button>` elements default to `type="submit"`. If the profile layout or API key overlay happens to be placed within a `<form>`, clicking these actions will inadvertently trigger a full-page form submission instead of the intended client-side save or delete operations. This brings the developer settings flow in line with the safe button enforcement patterns already merged into the codebase.


kushaltrivedi5 (approved):

Re-reviewed at head `9f895c26`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3472, opened 2026-06-24
refactor(api): parse upstream vault wrapper upsert error payloads as json

Roopmann30:
Refactors the error parsing logic inside `app/api/vault/wrapper/upsert/route.ts` to cleanly handle structured JSON errors from the upstream Python API.

**Why**: Previously, the route strictly used `await response.text()` for all non-200 responses. If the upstream API returned an actual JSON error object, it would either get double-stringified or wrapped awkwardly as `{ error: "{\"detail\":...}" }`. This refactor brings it in line with the error proxy patterns used in the other vault routes (e.g., `delete/route.ts`), ensuring the frontend receives clean, predictable error objects to map.

### PR Compliance

- [x] **DCO Signed-off**: Yes
- [x] **Scope**: Single-file, backend route API refactor
- [x] **Safety**: Safe error-handling change; no state/database modifications
- [x] **Behavior**: Does not change successful response shapes, only makes error shapes safer and strictly parsed
- [x] **Hygiene**: Passes all local typecheck and linting constraints
- [x] **Focus**: Targeted strictly to the exact bug surface to avoid `pr_claim_changed_surface_mismatch`

### DIFF
--- a/hushh-webapp/app/api/vault/wrapper/upsert/route.ts
+++ b/hushh-webapp/app/api/vault/wrapper/upsert/route.ts
@@ -90,11 +90,12 @@ export async function POST(request: NextRequest) {
     });
 
     if (!response.ok) {
-      const errorText = await response.text();
-      return NextResponse.json(
-        { error: errorText || "Backend error" },
-        { status: response.status }
-      );
+      const errorPayload = await response
+        .json()
+        .catch(async () => ({
+          error: (await response.text().catch(() => "")) || "Backend error",
+        }));
+      return NextResponse.json(errorPayload, { status: response.status });
     }
 
     const result = await response.json();


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge
A maintainer will stage the bounded patch onto the canonical attach point (`hushh-webapp/app/api/vault/wrapper/upsert/route.ts`), preserving your intent and adding the missing guard/validation/test, then land it co-authored to you. No action is required from you while this is queued; if you would prefer to apply the change yourself, attach it to `hushh-webapp/app/api/vault/wrapper/upsert/route.ts` and push to this branch.

### Proof Needed

PR #3485, opened 2026-06-24
test(utils): characterize formatCompleteJson on null and mixed arrays

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/utils/json-to-human.ts`) that pins down its exact, narrow
contract when a payload array contains nested `null` elements or mixed
primitives.

The literal assertions captured by this test:

- Generic (unrecognized-key) arrays render `null` elements as the literal
  string `null` **without throwing**.
- Generic arrays stringify mixed primitives (number / string / boolean).
- Generic arrays truncate after 3 elements and emit a `... and N more` line.
- Generic-array object elements render their first non-nullish value.
- Empty arrays are skipped (no section header emitted).
- Top-level `null` / `undefined` sections are skipped.
- The specialized `holdings` and `asset_allocation` branches **throw a
  `TypeError`** on a `null` element — they dereference each element as an
  object. This is the real, current boundary, documented rather than glossed.
- Well-formed `holdings` entries with missing optional fields format safely.

## Scope correction (truth-first)

- The original task referenced the path
  `hushh-research/__tests__/utils/...`. There is no top-level `__tests__` in
  this repo; vitest (`hushh-webapp/vitest.config.ts`) only includes
  `__tests__/**/*.test.{ts,tsx}` under `hushh-webapp`. The runnable file lives
  at `hushh-webapp/__tests__/utils/format-complete-json-bounds.test.ts`.
- The original "handles these invariants safely without throwing" framing is
  only partially true. `formatCompleteJson` is **not** uniformly null-safe;
  this test documents both the safe generic-array boundary and the throwing
  specialized-array boundary, so the contract is captured as-is.

## Verification

- `npm test -- __tests__/utils/format-complete-json-bounds.test.ts` → 9/9 pass

## CI gate checklist

- [x] Truth-First — documents real, existing behavioral boundaries (including the throw path)

### DIFF
--- a/hushh-webapp/__tests__/utils/format-complete-json-bounds.test.ts
+++ b/hushh-webapp/__tests__/utils/format-complete-json-bounds.test.ts
@@ -1,0 +1,113 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+/**
+ * Characterization tests for `formatCompleteJson` from
+ * `hushh-webapp/lib/utils/json-to-human.ts`.
+ *
+ * Scope: pin down the *exact, narrow* contract of how this formatter behaves
+ * when a payload contains arrays with nested `null` elements or mixed
+ * primitive members.
+ *
+ * Truth-first note: the formatter is NOT uniformly null-safe across array
+ * shapes. The generic array branch (for unrecognized section keys) tolerates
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `b5938a55`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3544, opened 2026-06-25
test(navigation): characterize resolveInternalRouteHref multi-param order lock

abdulbeast4-creator:
## Summary

Documents the parameter-preservation contract of `resolveInternalRouteHref`
(`hushh-webapp/lib/navigation/routes.ts`) when a pre-formed multi-key query
string is passed through resolution.

## Truth-first premise correction

  `hushh-webapp/__tests__/navigation/resolve-param-order.test.ts`.
- **The task premise was wrong and reading the source corrected it.**
  `resolveInternalRouteHref` does **not** append query keys, parse a query, or
  apply any lexicographical/declaration-order sorting. Its entire body is:

  ```ts
  export function resolveInternalRouteHref(value, fallback) {
  }
  ```

  So it only trims, rejects empty / non-`/`-leading / protocol-relative (`//`) /
  CR-LF inputs, and otherwise returns the input string **verbatim**; on rejection
  it returns the provided `fallback`. The "multi-param order lock" it guarantees
  is therefore **raw pass-through order preservation**, not active ordering.
  (Query KEY ordering elsewhere in this module is the job of the private
  `withQuery` helper via `URLSearchParams` insertion order — but
  `resolveInternalRouteHref` never calls it.)

## Literal assertions pinned

- Raw order of multiple keys preserved verbatim (`/search?b=2&a=1&c=3` unchanged)
- Keys are **not** sorted lexicographically (`?z=1&a=2` stays, not `?a=2&z=1`)
- Duplicate keys keep their original positions
- Empty-value key + bare key preserved exactly (`/p?a=&b&c=3`)
- Already percent-encoded query values are **not** re-encoded
- `null` / `undefined` / whitespace-only → returns `fallback`
- Protocol-relative (`//evil.com/...`) and external (`https://...`) multi-param
  URLs → returns `fallback`
- Surrounding whitespace trimmed; interior query order intact

## Verification

- `npm test -- __tests__/navigation/resolve-param-order.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — corrects the "appends/sorts params" premise to the real
  validate-and-pass-through-or-fallback contract

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-param-order.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-param-order.test.ts
@@ -1,0 +1,79 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts).
+//
+// TRUTH-FIRST PREMISE CORRECTION:
+//   The task asked to verify that resolveInternalRouteHref "appends multiple
+//   query keys" and preserves "lexicographical or raw declaration order"
+//   throughout "resolution limits". Reading the source corrected that premise.
+//   resolveInternalRouteHref does NO query construction, parsing, sorting, or
+//   reordering. Its entire body is:
+//
+//     export function resolveInternalRouteHref(value, fallback) {
+//       return normalizeInternalRouteHref(value) ?? fallback;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `6e7ab5f3`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3545, opened 2026-06-25
test(utils): characterize formatCompleteJson on large numeric bigint inputs

abdulbeast4-creator:
## Summary

Characterizes the output structure of `formatCompleteJson`
(`hushh-webapp/lib/utils/json-to-human.ts`) under high-magnitude numeric,
BigInt-like string, and real `bigint` field inputs.

  `hushh-webapp`. The file therefore lives at
  `hushh-webapp/__tests__/utils/format-json-bigint.test.ts`.
- **There is no special BigInt handling in the code; the test pins what really
  happens.** `formatValue` only branches on `typeof === number | string |
  boolean`, otherwise `String(value)`. Verified consequences:
  - A top-level `number` → scalar branch → `formatNumber` (Intl en-US,
    `maximumFractionDigits: 4`) with grouping:
    `9007199254740991` → `9,007,199,254,740,991`.
  - A top-level `number` with >4 fraction digits is rounded:
    `1234.56789` → `1,234.5679`.
  - A **real top-level `bigint` is silently skipped** — it is not number/string,
    not an array, and `typeof bigint !== "object"`, so it matches no branch and
    emits no line.
  - A **nested `bigint`** falls through `formatValue` to `String(value)`: plain
    decimal digits, **no** grouping and **no** trailing `n`
    (`9007199254740993n` → `9007199254740993`).
  - A **BigInt-like string** stays a string (markdown-cleaned, else verbatim),
    so full precision is retained because it never becomes a `Number`
    (`"123456789012345678901234567890"` preserved exactly).
  - Nested currency fields still format as USD (`ending_value: 1000000000` →
    `$1,000,000,000.00`).
  - A mixed bigint/big-string/big-number object does not throw.

## Verification

- `npm test -- __tests__/utils/format-json-bigint.test.ts` → 7/7 pass

## CI gate checklist

- [x] Truth-First — pins the real number/bigint/string behavior incl. the
  top-level-bigint skip and the nested `String()` fallthrough

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-bigint.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-bigint.test.ts
@@ -1,0 +1,73 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on high-magnitude /
+// BigInt-like inputs.
+//
+// TRUTH-FIRST notes (verified against the source):
+//   - There is no special BigInt handling. formatValue branches on
+//     `typeof === "number" | "string" | "boolean"`, else `String(value)`.
+//   - A top-level `number` routes through the scalar branch → formatNumber
+//     (Intl en-US, maximumFractionDigits: 4) → grouped digits.
+//   - A real top-level `bigint` is NOT number/string, not an array, and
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):
## Approved: characterization test at head `f28ae126`

Nice, focused characterization test — it imports and exercises the real production helper from `@/lib` (vitest, no self-mock), locks in current behavior, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to the merge queue.

PR #3549, opened 2026-06-25
test(navigation): characterize normalizeInternalRouteHref multi-dot path extensions

abdulbeast4-creator:
## Summary

Characterizes how `normalizeInternalRouteHref`
(`hushh-webapp/lib/navigation/routes.ts`) treats multi-dot / explicit relative
path segments such as `/app/../legal/..`.

  `hushh-webapp/__tests__/navigation/normalize-multi-dots.test.ts`.
- **There is no "tracking logic" and no path normalization.** The guard is a
  pure prefix/character check:

  ```ts
  if (!href) return null;
  return href;
  ```

  It never collapses `..`, `.` or repeated dots. Dot segments are ordinary path
  characters that survive verbatim when the input passes the prefix check.

## Behavior pinned

- `/app/../legal/..` → preserved verbatim (no traversal collapse)
- `/a/./b` → preserved verbatim
- `/../secret` → preserved verbatim (does not resolve above root)
- `/x/..../y` → long dot runs preserved as literal text
- `/files/archive.tar.gz` → dotted filename preserved
- `//evil.com/../x` → `null` (protocol-relative still rejected before dots matter)
- `../legal` and `./legal` → `null` (do not start with `/`)
- `   /app/../legal   ` → ends trimmed, dot segments kept (`/app/../legal`)

## Verification

- `npm test -- __tests__/navigation/normalize-multi-dots.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — pins the no-normalization, verbatim-or-null guard rather than
  an assumed traversal-resolving / tracking-isolation behavior

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-multi-dots.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-multi-dots.test.ts
@@ -1,0 +1,61 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on multi-dot / explicit
+// relative path segments (e.g. "/app/../legal/..", "/a/./b", "/x..y").
+//
+// Real guard:
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `3690736e`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3569, opened 2026-06-25
test(navigation): characterize normalizeInternalRouteHref percent-encoded slashes

abdulbeast4-creator:
## Summary

Pins how `normalizeInternalRouteHref`
(`hushh-webapp/lib/navigation/routes.ts`) reacts when a percent-encoded forward
slash (`%2F` / `%2f`) appears inside a path segment or query string parameter
(e.g. `/docs/legal%2Farchive`). Test-only; no production change.

  `hushh-webapp/__tests__/navigation/normalize-encoded-slashes.test.ts`.
- **There is no path splitting, decoding, or sub-segment isolation.** The guard
  is a pure prefix/character gate:

  ```ts
  if (!href) return null;
  return href;
  ```

  `%2F` is **not** decoded to `/` and is **never** treated as a path delimiter —
  it is three opaque characters returned verbatim. The only `/`-aware logic is
  the leading-character guard (first raw char must be a single `/`, not `//`).

## Behavior pinned

- `/docs/legal%2Farchive` → preserved verbatim (no decode/split)
- `/docs/legal%2farchive` → lowercase `%2f` preserved (no case-folding)
- `/search?path=a%2Fb%2Fc` → encoded slashes in a query value preserved
- `/a%2F%2Fb` → consecutive encoded slashes not collapsed
- `/%2Fevil.com` → accepted; leading char is the real `/`, char 2 is `%` (not a
  `//` protocol-relative prefix), so `%2F` stays encoded
- `//host%2Fpath` → `null` (genuine protocol-relative prefix still rejected)
- `   /docs/legal%2Farchive   ` → ends trimmed, encoded slashes intact

## Real boundary discovered during verification

The CR/LF guard runs **after** `trim()`. A test asserting that a *trailing*
newline is rejected initially failed — `trim()` strips it first, so the href
passes. The suite now pins both true behaviors:

- `/docs/legal%2F\narchive` (interior newline) → `null`
- `/docs/legal%2Farchive\n` (trailing newline) → `"/docs/legal%2Farchive"`
  (trimmed, accepted)

## Verification

- `npm test -- __tests__/navigation/normalize-encoded-slashes.test.ts` → 9/9 pass

## CI gate checklist

- [x] Truth-First — pins the verbatim-or-null gate (no `%2F` decode, no path
  isolation) and the trim-before-CR/LF interior-vs-trailing newline boundary

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-encoded-slashes.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-encoded-slashes.test.ts
@@ -1,0 +1,82 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on percent-encoded forward
+// slashes (`%2F` / `%2f`) appearing inside path or query string parameters.
+//
+// Real implementation:
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
+//   if (!href.startsWith("/") || href.startsWith("//")) return null;
+//   if (/[\r\n]/.test(href)) return null;
+//   return href;
+//
+// TRUTH-FIRST: normalizeInternalRouteHref does NOT decode, parse, split, or
+// segment the path. It performs no `%2F` → `/` decoding and no "sub-segment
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `f97bdf8b`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3573, opened 2026-06-25
test(table): assert container role="region" accessibility contract

imsharukhan:
## Summary

Adds an accessibility contract test for the table container.

## What changed

- Appended one test to `__tests__/ui/table.test.tsx`
- Verifies the table container exposes:
  - `role="region"`
  - `aria-label="Scrollable table"`

## Why

The accessibility attributes already exist in the component but were not covered by tests.

## Risk

- Test-only change
- Append-only
- No production code changes
- No runtime behaviour changes
- No new dependencies

### DIFF
--- a/hushh-webapp/__tests__/ui/table.test.tsx
+++ b/hushh-webapp/__tests__/ui/table.test.tsx
@@ -21,4 +21,24 @@ describe("Table", () => {
 
     expect(tableContainer?.getAttribute("tabindex")).toBe("0");
   });
+  it("renders the table container with an accessible region role and label", () => {
+    const { container } = render(
+      <Table>
+        <tbody>
+          <tr>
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Appreciated work — the added coverage is welcome and lands on the right surface. It just needs a refresh against the current train before it can merge.

**Blocker:** This PR is currently CONFLICTING with `integration/pr-train` (merge state DIRTY), so it cannot be queued as-is.

**Path To Merge:** Rebase this branch on the latest `integration/pr-train` and resolve the conflict, then push; it will be re-reviewed and queued at the fresh head.

**Proof Needed:** Clean merge (no conflict markers) plus the CI Status Gate green at the rebased head.

PR #3610, opened 2026-06-25
fix(db): URL-encode database credentials

parthmawai:
## Summary

URL-encode database usernames and passwords when constructing PostgreSQL connection URLs.

## Problem

PostgreSQL connection URLs become invalid when database usernames or passwords contain reserved URI characters (for example `@`, `:`, `/`, `%`, or spaces). This can cause connection failures even when the supplied credentials are correct.

## Solution

* URL-encode `db_user` and `db_password` before constructing PostgreSQL connection URLs.
* Apply the same behavior consistently to both direct connection and SQLAlchemy connection paths.

## Testing

* Verified that all PostgreSQL connection URL builders now URL-encode database credentials.
* Confirmed the change is limited to credential encoding with no unrelated behavioral changes.

### DIFF
--- a/consent-protocol/db/connection.py
+++ b/consent-protocol/db/connection.py
@@ -173,8 +173,8 @@ def get_database_url() -> str:
         )
     if db_unix_socket:
         # Cloud SQL Unix socket path must be provided via query host parameter.
-        return f"postgresql://{db_user}:{db_password}@/{db_name}?host={quote_plus(db_unix_socket)}"
-    return f"postgresql://{db_user}:{db_password}@{db_host}:{db_port}/{db_name}"
+        return f"postgresql://{quote_plus(db_user)}:{quote_plus(db_password)}@/{db_name}?host={quote_plus(db_unix_socket)}"
+    return f"postgresql://{quote_plus(db_user)}:{quote_plus(db_password)}@{db_host}:{db_port}/{db_name}"
 
 
 def get_database_ssl():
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Thank you for this contribution — the intent here is sound and it's aimed at the right surface. This change lands on a trust-boundary file (`consent-protocol/db/connection.py`), so before it can merge it needs a maintainer-authored security patch staged on top to make the enforcement independently stronger, not just preserved while touched.

**Blocker:** The diff touches `consent-protocol/db/connection.py`, a consent/identity/finance trust boundary. Per the PR-train write contract, a change on this surface must demonstrably strengthen DB-backed enforcement / scope / caller-credential shape — a maintainer patch will be staged to guarantee that bar before queueing.

**Path To Merge:** Keep the PR open; a maintainer security patch will be staged on top at `consent-protocol/db/connection.py` (co-authored to you), then the combined head is re-reviewed and queued. No action needed from you unless you want to tighten validation/error-surface at that attach point yourself first.

**Proof Needed:** A test asserting the strengthened behavior at `consent-protocol/db/connection.py` (an existing/your/maintainer-added test) green, plus the CI Status Gate green at the patched head.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `d365d60a`

PR #3611, opened 2026-06-25
fix(ticker-cache): distinguish empty cache from unloaded cache

parthmawai:
## Summary

Update `TickerCache.loaded` to reflect whether the cache has been initialized rather than whether it currently contains rows.

## Problem

The previous implementation returned `bool(self._rows)`, which reported the cache as "not loaded" whenever it contained zero rows. An empty cache is a valid state after a successful load and should not be treated as uninitialized.

## Solution

* Base `loaded` on the cache initialization timestamp (`loaded_at`) instead of the current row count.
* Preserve the distinction between:

  * an uninitialized cache, and
  * a successfully loaded cache with zero results.

## Testing

* Ran the ticker cache test suite locally.

```text
28 passed in 1.35s
```

### DIFF
--- a/consent-protocol/hushh_mcp/services/ticker_cache.py
+++ b/consent-protocol/hushh_mcp/services/ticker_cache.py
@@ -88,7 +88,7 @@ class TickerCache:
     @property
     def loaded(self) -> bool:
         with self._lock:
-            return bool(self._rows)
+            return self._loaded_at > 0.0
 
     @property
     def loaded_at(self) -> float:


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Thank you for this contribution — the intent here is sound and it's aimed at the right surface. This change lands on a trust-boundary file (`consent-protocol/hushh_mcp/services/ticker_cache.py`), so before it can merge it needs a maintainer-authored security patch staged on top to make the enforcement independently stronger, not just preserved while touched.

**Blocker:** The diff touches `consent-protocol/hushh_mcp/services/ticker_cache.py`, a consent/identity/finance trust boundary. Per the PR-train write contract, a change on this surface must demonstrably strengthen DB-backed enforcement / scope / caller-credential shape — a maintainer patch will be staged to guarantee that bar before queueing.

**Path To Merge:** Keep the PR open; a maintainer security patch will be staged on top at `consent-protocol/hushh_mcp/services/ticker_cache.py` (co-authored to you), then the combined head is re-reviewed and queued. No action needed from you unless you want to tighten validation/error-surface at that attach point yourself first.

**Proof Needed:** A test asserting the strengthened behavior at `consent-protocol/hushh_mcp/services/ticker_cache.py` (an existing/your/maintainer-added test) green, plus the CI Status Gate green at the patched head.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `81de36a6`

PR #3614, opened 2026-06-25
test(navigation): characterize resolveInternalRouteHref null parameter values

abdulbeast4-creator:
## Summary

Pins how `resolveInternalRouteHref`
(`hushh-webapp/lib/navigation/routes.ts`) behaves when an internal href carries
query parameters set to `null`/empty states (e.g. `/dashboard?filter=null&mode=`).

  `hushh-webapp/__tests__/navigation/resolve-null-params.test.ts`.
- **There is no per-field query mapping or stripping.** The function is a thin
  wrapper:

  ```ts
  export function resolveInternalRouteHref(value, fallback) {
  }
  ```

  `normalizeInternalRouteHref` trims, then returns `null` if the value is empty,
  not `/`-prefixed, `//`-prefixed, or contains `\r`/`\n`; otherwise it returns
  the href **verbatim**. The query string is opaque text — `filter=null` keeps
  the literal 4-char `null`, and `mode=` keeps its trailing `=`. Nothing is
  parsed, decoded, coalesced, or removed.
- The premise that fields are "mapped exactly as text or stripped seamlessly"
  is only half-true: there is no stripping at all. The real contract is
  **whole-href accept (verbatim) vs. reject (replaced wholesale by `fallback`)**.

## Behavior pinned

- `/dashboard?filter=null&mode=` → returned verbatim (literal `null`, empty `=`)
- `/x?a=` → trailing `=` kept; `/x?flag` → bare key kept
- `/p?v=null` → literal `null` never coalesced
- `/p?a=&b=&c=` → all empty params + `&` separators preserved
- `null` value → `fallback`; `""` value → `fallback`
- `//evil.com?filter=null` → `fallback` (protocol-relative rejected)
- `  /dashboard?filter=null&mode=  ` → ends trimmed, query intact

## Verification

- `npm test -- __tests__/navigation/resolve-null-params.test.ts` → 9/9 pass

## CI gate checklist

- [x] Truth-First — pins the verbatim-or-fallback contract (no query parsing,
  mapping, or stripping) rather than an assumed per-field resolver

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-null-params.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-null-params.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on hrefs whose query strings
+// contain "null" / empty parameter values (e.g. /dashboard?filter=null&mode=).
+//
+// Real implementation:
+//   export function resolveInternalRouteHref(value, fallback) {
+//     return normalizeInternalRouteHref(value) ?? fallback;
+//   }
+//   normalizeInternalRouteHref: trims; null if empty, not "/"-prefixed,
+//   "//"-prefixed, or containing \r/\n; otherwise returns the href verbatim.
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):
## Approved

Green CI Status Gate, MERGEABLE, base integration/pr-train at head `b3f872bb`. Test/contract coverage addition with no trust-boundary weakening. Enqueued.

PR #3632, opened 2026-06-25
test(utils): characterize formatCompleteJson on highly saturated int array matrices

abdulbeast4-creator:
## Summary

Pins the literal contract of `formatCompleteJson`
(`hushh-webapp/lib/utils/json-to-human.ts`) when a dense multi-dimensional
integer matrix is supplied under an unrecognized top-level key (coordinate rows
/ weight indices). Test-only; no production change.

## Truth-first scope correction (literal assertion being made)

  `hushh-webapp/__tests__/utils/format-json-matrices.test.ts`.
- A matrix under an unknown key hits the **"Generic array handling"** branch:

  ```ts
  lines.push(`--- ${label} (${value.length} items) ---`);
  for (const item of value.slice(0, 3)) {
    if (typeof item === "object" && item !== null) {
      const firstValue = Object.values(item).find(v => v !== null && v !== undefined);
      lines.push(`  • ${String(firstValue || "Item")}`);
    } else {
      lines.push(`  • ${String(item)}`);
    }
  }
  if (value.length > 3) lines.push(`  ... and ${value.length - 3} more`);
  ```

- **The premise that it "maps these arrays without altering numerical value
  ordering or index bounds" is FALSE.** The branch is destructively lossy:
  1. Only the **first three rows** are visited; the rest collapse to a single
     `... and N more` line — row bounds are not retained.
  2. Each visited row is an array (`typeof === "object"`), so it prints only the
     row's **first element** (`Object.values(row).find(...)`); every other
     column is discarded — column bounds/ordering are not preserved.
  3. `String(firstValue || "Item")` means a row whose first element is the
     **falsy integer 0** renders the literal `"Item"`, not `0` — the single
     surviving value is corrupted for zero-led rows.

## Literal behavior pinned

- `weights:[[1,2,3],[4,5,6],[7,8,9],[10,11,12]]` →
  `--- Weights (4 items) ---` / `• 1` / `• 4` / `• 7` / `... and 1 more`
- `matrix:[[100,200],[300,400]]` → `• 100` / `• 300` (200 and 400 dropped)
- `coords:[[0,9],[5,5]]` → `• Item` / `• 5` (zero-led row → "Item")
- `grid:[[-7,-8],[-9,-10]]` → `• -7` / `• -9` (negatives truthy; rest dropped)
- `rows:[[1],[2],[3]]` → three bullets, no overflow line
- `tensor:[[[1,2]],[[3,4]]]` → `• 1,2` / `• 3,4` (first inner array via String)

## Verification

- `npm test -- __tests__/utils/format-json-matrices.test.ts` → 6/6 pass

## CI gate checklist

- [x] Truth-First — documents the real lossy first-element/first-3-row collapse
  (incl. the `0 || "Item"` corruption) rather than assumed ordering/bound-safe
  matrix mapping

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-matrices.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-matrices.test.ts
@@ -1,0 +1,81 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on dense multi-dimensional
+// integer array matrices supplied under an unrecognized top-level key (e.g.
+// coordinate rows or weight indices), which fall into the "Generic array
+// handling" branch:
+//
+//   lines.push("");
+//   lines.push(`--- ${label} (${value.length} items) ---`);
+//   for (const item of value.slice(0, 3)) {
+//     if (typeof item === "object" && item !== null) {
+//       const obj = item;
+//       const firstValue = Object.values(obj).find(v => v !== null && v !== undefined);
+//       lines.push(`  • ${String(firstValue || "Item")}`);
+//     } else {
+//       lines.push(`  • ${String(item)}`);
+//     }
+//   }
+//   if (value.length > 3) lines.push(`  ... and ${value.length - 3} more`);
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `b8b6321f`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3637, opened 2026-06-25
test(navigation): characterize normalizeInternalRouteHref double question mark queries

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) to map the contract boundary when a raw
internal path contains non-standard **double question marks**
(`/search??query=value`, `/page?id=1?type=all`). Test-only; no production change.

  no top-level `__tests__` in this repo; vitest only includes `__tests__/**`
  under `hushh-webapp`. The file lives at
  `hushh-webapp/__tests__/navigation/normalize-double-questions.test.ts`.
- **The guard does NOT flag, collapse, or strip secondary `?` characters.**
  `normalizeInternalRouteHref` never parses the query string. It only applies
  whole-href structural guards (reject empty/whitespace-only, reject non-rooted,
  reject protocol-relative `//`, reject CR/LF) and otherwise returns the trimmed
  string **verbatim**. Every extra `?` is preserved exactly as-is. The premise
  that it might "flag the secondary characters" is **false**.
- A real ordering nuance is also pinned: `trim()` runs **before** the CR/LF
  guard, so a **trailing** newline is stripped and the path is **accepted**,
  while an **interior** newline is still **rejected** (→ `null`).

## Behavior pinned

- `/search??query=value` → returned verbatim
- `/page?id=1?type=all` → returned verbatim
- `/a???b` → returned verbatim (3+ `?`)
- `/results??` → returned verbatim (trailing bare double-?)
- `search??query=value` (no leading slash) → `null`
- `//evil.com??x=1` (protocol-relative) → `null`
- `/search??q=1\n` (trailing newline) → `/search??q=1` (trimmed, then accepted)
- `/search??q=1\nx=2` (interior newline) → `null`

## Verification

- `npm test -- __tests__/navigation/normalize-double-questions.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — pins the real verbatim pass-through of extra `?` (plus the
  trim-before-CR/LF ordering) rather than an assumed flagging/sanitizing guard

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-double-questions.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-double-questions.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on non-standard DOUBLE
+// question marks inside raw query paths (e.g. "/search??query=value" or
+// "/page?id=1?type=all").
+//
+// TRUTH-FIRST: normalizeInternalRouteHref does NOT parse, split, or validate the
+// query string. It performs only structural guards on the whole href:
+//   - reject empty / whitespace-only input  → null
+//   - reject input not beginning with "/"   → null
+//   - reject protocol-relative "//..."       → null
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `cc7403ac`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3652, opened 2026-06-25
test(utils): characterize formatCompleteJson on nested map and set objects

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/utils/json-to-human.ts`) documenting its exact, literal
serialization output when a section value is a native JavaScript `Map` or `Set`.

  `hushh-webapp/__tests__/utils/format-json-collections.test.ts`.
- **The premise that Map/Set "map safely without dropping data elements" is
  FALSE.** `formatCompleteJson` is written for plain JSON. A `Map`/`Set`:
  - is not a number/string → skips the scalar branch,
  - is not `Array.isArray()` → skips the array branch,
  - is `typeof === "object"` → enters the object branch, which enumerates with
    `Object.entries(value)`.
  `Object.entries(map)` / `Object.entries(set)` return `[]` because Map entries
  and Set members live in internal slots, not own enumerable string properties.
- **Literal assertion being pinned:** only the section HEADER is emitted and
  every contained element is silently DROPPED. `.size` is never read; Map
  keys/values and Set members never appear in the output.

## Literal output pinned (exact strings)

- `{ prefs: new Map([["a",1],["b",2]]) }` → `"\n--- Prefs ---"` (entries dropped)
- `{ tags: new Set([1,2,3]) }` → `"\n--- Tags ---"` (members dropped)
- `{ config: { prefs: Map } }` → `"\n--- Config ---\n  Prefs:"` (label only)
- `{ config: { tags: Set } }` → `"\n--- Config ---\n  Tags:"` (label only)
- `{ name: "x", prefs: Map }` → `"Name: x\n\n--- Prefs ---"` (sibling kept, Map dropped)
- `{ prefs: new Map([["secretKey","secretValue"]]) }` → output contains neither
  `secretKey` nor `secretValue`

## Verification

- `npm test -- __tests__/utils/format-json-collections.test.ts` → 6/6 pass

## CI gate checklist

- [x] Truth-First — documents the real data-loss boundary (Map/Set entries are
  dropped, header only) rather than an assumed safe-collection serializer

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-collections.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-collections.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) processing native JavaScript Map and
+// Set objects as section values.
+//
+// TRUTH-FIRST — IMPORTANT CORRECTION: the premise that Map/Set "map safely
+// without dropping data elements" is FALSE. formatCompleteJson is built for
+// plain JSON. A Map or Set is:
+//   - not a number/string  → skips the scalar branch
+//   - not Array.isArray()   → skips the array branch
+//   - typeof === "object"   → ENTERS the object branch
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `90495066`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3657, opened 2026-06-25
test(navigation): characterize normalizeInternalRouteHref percent-encoded spaces in query

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) to map how it treats percent-encoded
spaces (`%20`) inside the SEARCH QUERY portion of a rooted href
(e.g. `/search?q=open%20source`). Test-only; no production change.

  `hushh-webapp/__tests__/navigation/normalize-query-spaces.test.ts`.
- **`normalizeInternalRouteHref` does NOT parse, decode, or re-encode the query
  string.** Full contract:
  ```
  if (!href) return null;                                   // empty
  if (!href.startsWith("/") || href.startsWith("//")) return null; // rooted, not protocol-relative
  if (/[\r\n]/.test(href)) return null;                     // no CR/LF
  return href;                                              // else VERBATIM
  ```
- So `%20` in the query is **retained literally** — never decoded to a space,
  never re-encoded, never collapsed. The only mutation is an OUTER `.trim()`.

## Truth-first correction discovered while writing the test

An initial assertion assumed a trailing `\n` would be rejected by the CR/LF
guard. The local run proved that **FALSE**: `.trim()` executes *before* the
`/[\r\n]/` check, so a trailing newline is stripped first and the href is
**accepted**, not rejected. The CR/LF guard only rejects **interior** newlines
that survive the trim. The test was corrected to pin this real ordering.

## Behavior pinned

- `/search?q=open%20source` → verbatim (`%20` retained, not decoded)
- `%20` never becomes a literal space (no decode)
- `/search?q=a%20b%20c&sort=desc` → verbatim (multiple `%20` retained)
- `/search?q=open source` (raw space) → verbatim (no encoding added)
- `  /search?q=open%20source  ` → outer-trimmed, inner `%20` preserved
- `//evil.com/?q=open%20source` → `null` (protocol-relative rejected)
- `/search?q=open%20\nsource` (interior `\n`) → `null` (CR/LF guard)
- `/search?q=open%20source\n` (trailing `\n`) → trimmed → verbatim (NOT null)

## Verification

- `npm test -- __tests__/navigation/normalize-query-spaces.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — pins the real "query is opaque, `%20` retained verbatim,
  trim-before-CRLF-guard" contract rather than an assumed query normalizer

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-query-spaces.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-query-spaces.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on percent-encoded spaces
+// (`%20`) appearing inside the SEARCH QUERY portion of a rooted href.
+//
+// TRUTH-FIRST: normalizeInternalRouteHref does NOT parse or decode the query
+// string. Its full contract is:
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;                                   // empty
+//   if (!href.startsWith("/") || href.startsWith("//")) return null; // rooted, not protocol-relative
+//   if (/[\r\n]/.test(href)) return null;                     // no CR/LF
+//   return href;                                              // else VERBATIM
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `c918a3ea`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3658, opened 2026-06-25
test(navigation): characterize resolveInternalRouteHref parameter key case matching

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) documenting how it treats query
parameters that share a name but differ only by letter casing
(e.g. `/profile?userId=123&userid=456`). Test-only; no production change.

  `hushh-webapp/__tests__/navigation/resolve-key-casing.test.ts`.
- **The premise that `resolveInternalRouteHref` "parses search parameters" or
  enforces "case-sensitivity invariants" / an "override mapping" between
  duplicate-by-case keys is FALSE.** It does NOT parse the query string at all.
  Its entire body is:
  ```
  }
  ```
  `normalizeInternalRouteHref` only trims, rejects empty/non-rooted/protocol-
  relative/interior-CRLF hrefs, and otherwise returns the href **verbatim**.
- The query string is therefore **opaque**: every key is preserved as written, in
  original order, with **no dedupe, no case-folding, and no first/last-wins
  override**. There is no parameter-keyed map, so a "case collision" never occurs.
  `fallback` is used only when the whole href is rejected — never per-parameter.

## Behavior pinned

- `/profile?userId=123&userid=456` → verbatim (both casings kept, no override)
- `/profile?userid=456&userId=123` → verbatim (original order retained)
- `/profile?UserID=1&USERID=2&userid=3` → verbatim (no case-folding)
- `/profile?userId=123&userId=456` → verbatim (exact duplicates also opaque)
- `//evil.com/profile?userId=1&userid=2` → `fallback` (whole href rejected)
- `""` → `fallback`

## Verification

- `npm test -- __tests__/navigation/resolve-key-casing.test.ts` → 6/6 pass
- (An IDE-only `@/` module-resolution warning is a false positive — the same
  alias resolves under vitest, as the passing run confirms.)

## CI gate checklist

- [x] Truth-First — pins the real "query is opaque, no key parsing / casing
  override" contract rather than an assumed case-aware parameter resolver

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-key-casing.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-key-casing.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on query parameters that share
+// a name but differ ONLY by letter casing (e.g. `userId` vs `userid`).
+//
+// TRUTH-FIRST — IMPORTANT CORRECTION: the premise that resolveInternalRouteHref
+// "parses search parameters" or applies "case-sensitivity invariants" / an
+// "override mapping" between duplicate-by-case keys is FALSE. It does NOT parse
+// the query string at all. Its entire body is:
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `78d36a0b`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3696, opened 2026-06-26
test(utils): characterize formatCompleteJson high-order array value slicing

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/utils/json-to-human.ts`) documenting how it serializes object
configs carrying ultra-long linear array fields (5,000+ items). Test-only; no
production change.

  `hushh-webapp/__tests__/utils/format-array-slicing.test.ts`.
- **The premise that `formatCompleteJson` serializes "the complete dataset
  without applying clipping filters" is FALSE — it is the OPPOSITE of the real
  behavior.** The array branch ALWAYS clips. For a generic (non-special) array key:
  1. it emits a blank line, then a header `--- <Label> (<N> items) ---` that cites
     the FULL length N,
  2. renders only the FIRST 3 items via `sectionValue.slice(0, 3)`,
  3. if N > 3, appends `  ... and <N-3> more`.
  So a 5,000-item array yields a header citing `5000`, exactly 3 bullet lines, and
  `... and 4997 more` — the full dataset is NEVER serialized.
- **Gotcha pinned during local verification (one initial assertion failed):** the
  generic object branch renders `String(firstNonNullValue || "Item")`. `find`
  returns the first non-null/undefined value, so for `{ id: 0 }` it returns `0` —
  but `0 || "Item"` is falsy, so the bullet renders `Item`, not `0`. The test
  asserts the real output.
- Special-cased keys clip at different bounds: `holdings` → first 10,
  `activity_and_transactions` → first 5, while `legal_and_disclosures` /
  `historical_values` emit ONLY a count line and zero item bullets.

## Behavior pinned

- 5,000 primitives → header `(5000 items)`, bullets `0/1/2`, `... and 4997 more`
- 7,500 strings → full-length header, exactly 3 bullet lines, `... and 7497 more`
- generic object items → first non-null value, with falsy `0` → `Item`
- `legal_and_disclosures` (5,000) → `5000 disclosure(s) extracted`, no bullets
- exactly-3-item array → 3 bullets, no remainder line

## Verification

- `npm test -- __tests__/utils/format-array-slicing.test.ts` → 5/5 pass

## CI gate checklist

- [x] Truth-First — pins the real "array branch ALWAYS clips (first 3 + remainder),
  header cites full length, falsy-0 → 'Item'" contract rather than the assumed
  "serializes complete dataset without clipping" premise

### DIFF
--- a/hushh-webapp/__tests__/utils/format-array-slicing.test.ts
+++ b/hushh-webapp/__tests__/utils/format-array-slicing.test.ts
@@ -1,0 +1,86 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on object configs that carry
+// ultra-long linear array fields (5,000+ items).
+//
+// TRUTH-FIRST — IMPORTANT CORRECTION: the premise that formatCompleteJson
+// "serializes the COMPLETE dataset WITHOUT applying clipping filters" is FALSE —
+// it is the OPPOSITE of the real behavior. The array branch ALWAYS clips. For a
+// generic (non-special) array key it:
+//   1. pushes a blank line, then a header `--- <Label> (<N> items) ---` that
+//      reports the FULL length N,
+//   2. renders only the FIRST 3 items via `sectionValue.slice(0, 3)`,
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `621b9fd2`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3700, opened 2026-06-26
test(navigation): characterize normalizeInternalRouteHref multi-segment hash query chains

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) documenting how the path guard behaves
when an href places the query string AFTER the hash fragment
(e.g. `/profile#settings?theme=dark&status=active`). Test-only; no production
change.

`normalizeInternalRouteHref` is an **opaque allow-list guard, not a URL parser**.
Its full body is:

```ts
if (!href) return null;
return href;
```

It has **zero awareness of `#` or `?`** — it never splits, reorders, re-encodes,
or canonicalizes hash vs. query segments. Therefore, for any path-rooted
same-origin href, whatever trailing `#...?...` layout is supplied is preserved
**byte-for-byte** after a leading/trailing whitespace trim. The only rejections
are: empty, not starting with `/`, protocol-relative `//`, or containing CR/LF.

## Behavior pinned (7 cases)

- `/profile#settings?theme=dark&status=active` → returned unchanged
- `/dashboard#tab=reports?sort=desc&page=2` → NOT reordered; `#tab=reports?sort=desc` survives
- `/a#x?b=1?c=2#y&d=3` (multiple `?`/`#`) → returned unchanged
- `/search#q=a b&list=[1,2]?flag=true%20` (special chars) → NOT re-encoded
- `  /profile#settings?theme=dark  ` → trimmed to `/profile#settings?theme=dark`, inner layout intact
- `//evil.example#settings?theme=dark` → `null` (protocol-relative rejected)
- `/profile#settings?theme=dark\nstatus=active` → `null` (CRLF rejected)

## Truth-first scope note

The original task referenced `hushh-research/__tests__/navigation/...` and an
import from `lib/navigation/routes.ts`. There is no top-level `__tests__` in this
repo; vitest only includes `__tests__/**` under `hushh-webapp`, and the in-repo
alias for that module is `@/lib/navigation/routes`. The file therefore lives at
`hushh-webapp/__tests__/navigation/normalize-hash-queries.test.ts` and imports
via `@/lib/navigation/routes`.

## Verification

- `npm test -- __tests__/navigation/normalize-hash-queries.test.ts` → 7/7 pass

## CI gate checklist

- [x] Truth-First — documents the real opaque-passthrough boundary
  (no hash/query parsing, reordering, or re-encoding); no un-implemented
  generalizations

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-hash-queries.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-hash-queries.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on hrefs that place the query
+// string AFTER the hash fragment (e.g. "/profile#settings?theme=dark").
+//
+// TRUTH-FIRST — LITERAL CONTRACT: normalizeInternalRouteHref is an OPAQUE
+// allow-list guard, not a URL parser. Its full body is:
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
+//   if (!href.startsWith("/") || href.startsWith("//")) return null;
+//   if (/[\r\n]/.test(href)) return null;
+//   return href;
+// It has ZERO awareness of `#` or `?`. It does NOT split, reorder, re-encode, or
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `bf29cb57`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3703, opened 2026-06-26
test(navigation): characterize resolveInternalRouteHref parameter string boolean arrays

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) documenting its exact behavior when an
href's query string carries repeated / array-style boolean params
(e.g. `/api/v1?flags=true&flags=false`). Test-only; no production change.

## Truth-first correction to the premise

The task framed this as a "parameter extraction process" that "isolates and
registers" array boolean fields. **That premise is false** —
`resolveInternalRouteHref` does **no query parsing at all**. Its full body is:

```ts
```

and `normalizeInternalRouteHref` is an opaque allow-list guard:

```ts
if (!href) return null;
return href;
```

There is **no `URLSearchParams`, no de-duplication, no boolean coercion, no array
registration**. A valid same-origin href is returned **verbatim** (after a
whitespace trim); an invalid one short-circuits to the supplied `fallback`. The
tests pin that real contract instead of an imagined parser.

## Behavior pinned (7 cases, fallback = `/home`)

- `/api/v1?flags=true&flags=false` → returned unchanged (repeated params kept, in order)
- `/settings?on=true&on=true&on=false&off=false` → unchanged; all 4 pairs survive (no folding)
- `/api/v1?flags[]=true&flags[]=false` → bracket array syntax preserved untouched
- `/x?a=TRUE&b=False&c=1&d=0` → no boolean coercion; values kept as raw strings
- `//evil.example?flags=true&flags=false` → `/home` (protocol-relative → fallback)
- `""` / `null` / `undefined` → `/home` (empty/nullish → fallback)
- `/api/v1?flags=true\n&flags=false` → `/home` (CRLF → fallback)

## Truth-first scope note

The original task referenced `hushh-research/__tests__/navigation/...`, an import
from `lib/navigation/routes.ts`, and `npm test -- hushh-research/...`. There is no
`hushh-webapp`, and the module's in-repo alias is `@/lib/navigation/routes`. The
file therefore lives at
`hushh-webapp/__tests__/navigation/resolve-boolean-arrays.test.ts`.

## Verification

- `npm test -- __tests__/navigation/resolve-boolean-arrays.test.ts` → 7/7 pass

## CI gate checklist

- [x] Truth-First — corrects the false "param extraction" premise and pins the
  real opaque-passthrough-with-fallback contract; no un-implemented
  generalizations

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-boolean-arrays.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-boolean-arrays.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on hrefs whose query string
+// carries repeated/array-style boolean params (e.g. "?flags=true&flags=false").
+//
+// TRUTH-FIRST — IMPORTANT CORRECTION: the premise that this utility runs an
+// "extraction process" that "isolates and registers" array boolean fields is
+// FALSE. `resolveInternalRouteHref` performs NO query parsing whatsoever. Its
+// full body is literally:
+//   return normalizeInternalRouteHref(value) ?? fallback;
+// and `normalizeInternalRouteHref` is an opaque allow-list guard:
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `83a306a4`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3713, opened 2026-06-26
test(navigation): characterize normalizeInternalRouteHref absolute file paths

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) documenting how it treats absolute
file-system / `file://` style inputs (e.g. `file:///C:/app/index.html`,
`C:\app\index.html`, `/etc/passwd`). Test-only; no production change.

`normalizeInternalRouteHref` is a same-origin **allow-list guard**, not a path
normalizer or filesystem safety check. Its full body:

```ts
if (!href) return null;
return href;
```

It performs **no path resolution, no `..` collapsing, no drive-letter parsing,
no percent-decoding**. A surviving value is returned **byte-for-byte** (after a
whitespace trim). The single-`/` allow-list is a scheme/origin gate.

## Behavior pinned (9 cases)

External → `null`:
- `file:///C:/app/index.html` (starts with `f`, not `/`)
- `file:///etc/passwd`
- `C:\app\index.html` (starts with `C`)
- `//server/share/index.html` (explicit `startsWith("//")` UNC / protocol-relative guard)
- `/var/data\r\n/report.html` (CRLF rejected even after trim)

Same-origin → returned **verbatim** (the guard does NOT sanitize traversal-looking paths):
- `/etc/passwd` → `/etc/passwd`
- `/C:/app/index.html` → `/C:/app/index.html` (no drive parsing)
- `/../../etc/passwd` → `/../../etc/passwd` (no `..` collapsing)
- `   /var/data/report.html   ` → `/var/data/report.html` (trim, then echo)

## Security framing (truth-first)

This is **not** a path-traversal sanitizer and the tests do not pretend it is.
The guard's job is to keep `next/router` navigation same-origin: anything not
beginning with exactly one `/` (schemes, UNC, protocol-relative) is rejected as
external; anything that does is handed back unchanged. Filesystem-style strings
like `/etc/passwd` survive because they are valid same-origin *route* strings —
they are never resolved against a real filesystem by this function.

## Truth-first scope note

The original task referenced `hushh-research/__tests__/navigation/...` and
alias is `@/lib/navigation/routes`. The file therefore lives at
`hushh-webapp/__tests__/navigation/normalize-absolute-files.test.ts`.

## Verification

- `npm test -- __tests__/navigation/normalize-absolute-files.test.ts` → 9/9 pass

## CI gate checklist

- [x] Truth-First — pins the real "external → null, single-slash → verbatim"
  allow-list contract; explicitly does NOT claim traversal/drive sanitizing the
  function does not perform

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-absolute-files.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-absolute-files.test.ts
@@ -1,0 +1,76 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) against absolute file-system / file://
+// style inputs.
+//
+// TRUTH-FIRST — LITERAL CONTRACT: normalizeInternalRouteHref is a same-origin
+// allow-list guard, NOT a path normalizer. Its full body is:
+//
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
+//   if (!href.startsWith("/") || href.startsWith("//")) return null;
+//   if (/[\r\n]/.test(href)) return null;
+//   return href;
+//
+// Consequences for "absolute file" shapes:
+//  * It does NO path resolution, NO `..` collapsing, NO drive-letter handling,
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `ed8d9a73`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3714, opened 2026-06-26
test(utils): characterize formatCompleteJson deep object value cyclic ref guards

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/utils/json-to-human.ts`) documenting its behavior when the
input object contains a cyclic self-reference (`obj.self = obj`). Test-only; no
production change.

`formatCompleteJson` has **no cycle guard, no `try/catch`, and no "strip looping
params" logic** — and it does not need one, because it is **not recursive**. It
walks a hard-coded, fixed depth of exactly three levels via plain
`for...of Object.entries(...)` loops:

- level 1: section entries (top-level keys)
- level 2: field entries (inside an object section)
- level 3: nested-field entries (inside a nested object field)

At level 3 each value goes to `formatValue`, whose fallthrough for a
non-null/number/string/boolean value is `return String(value)` →
`"[object Object]"`. There is no fourth level and no self-call, so a cyclic
reference can **never** cause infinite recursion or a `RangeError`
(stack overflow). The cycle is simply truncated by the fixed depth.

## Behavior pinned (6 cases)

- `obj.self = obj` → does **not** throw
- `obj.self = obj` → `"\n--- Self ---\n  Self:\n    • Self: [object Object]"`
- mutual cycle `a.b = b; b.a = a` → does **not** throw
- mutual cycle → `"\n--- B ---\n  A:\n    • B: [object Object]"`
- `{ portfolio_summary: node }` with `node.cash_balance = 100; node.loop = node`
  → sibling scalar still renders `Cash Balance: $100.00`; the cyclic field is
  **not stripped**, it bottoms out at `[object Object]`
- a 1000-deep chain closed into a loop → completes, does **not** throw
  (fixed-depth walk ignores chain depth entirely)

## Truth-first framing

The task asked whether the utility "includes safety protections, throws a
standard error, or strips looping parameters." The honest answer is **none of
those** — and crucially, none are required. The function is structurally
cycle-safe by virtue of being a fixed 3-level iterator, not a recursive
serializer like `JSON.stringify` (which *would* throw
`Converting circular structure to JSON`). The tests assert that real property and
deliberately do **not** claim any WeakSet/seen-set guard, thrown error, or key
stripping that the code does not implement.

## Truth-first scope note

alias is `@/lib/utils/json-to-human`. The file therefore lives at
`hushh-webapp/__tests__/utils/format-cyclic-guards.test.ts`.

## Verification

- `npm test -- __tests__/utils/format-cyclic-guards.test.ts` → 6/6 pass

## CI gate checklist

- [x] Truth-First — pins the real "fixed-depth, naturally cycle-safe, no guard /
  no throw / no stripping" behavior; does NOT overclaim any circular-structure
  protection the code lacks

### DIFF
--- a/hushh-webapp/__tests__/utils/format-cyclic-guards.test.ts
+++ b/hushh-webapp/__tests__/utils/format-cyclic-guards.test.ts
@@ -1,0 +1,86 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) against cyclic self-references.
+//
+// TRUTH-FIRST — LITERAL CONTRACT: formatCompleteJson has NO cycle guard, NO
+// try/catch, and NO "strip looping params" logic. It does NOT need one, because
+// it is NOT recursive. It walks a HARD-CODED, FIXED depth of exactly three
+// levels via plain `for...of Object.entries(...)` loops:
+//
+//   level 1: section entries        (top-level keys)
+//   level 2: field entries          (inside an object section)
+//   level 3: nested-field entries   (inside a nested object field)
+//
+// At level 3 each value is passed to `formatValue`, whose fallthrough for a
+// non-null/number/string/boolean value is `return String(value)` ->
+// "[object Object]". There is no fourth level and no self-call, so a cyclic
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `af374722`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3715, opened 2026-06-26
test(navigation): characterize resolveInternalRouteHref string encoded colon properties

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) documenting how percent-encoded colons
(`%3A`) and raw `:` behave inside route/query text values. Test-only; no
production change.

`resolveInternalRouteHref` is a one-line wrapper:

```ts
```

and `normalizeInternalRouteHref` is purely:

```ts
if (!href) return null;
return href;
```

## Truth-first correction to the task framing

The task described a "search query extractor" that "reads percent-encoded colons"
and "isolates these from standard protocol or port selectors." **No such
extractor exists.** `resolveInternalRouteHref` does **not** parse query strings,
does **not** split on `?`/`&`/`=`, does **not** percent-decode, and has **no**
concept of protocol or port selectors. A `%3A` (or a raw `:`) is just ordinary
text inside the href string. The only decision is: same-origin (single leading
`/`, not `//`, no CRLF) → echo the string **verbatim**; otherwise → return the
caller's `fallback`.

## Behavior pinned (8 cases)

Kept **verbatim** (single leading `/`, `%3A`/`:` never decoded or special-cased):
- `/search?q=time%3A12%3A30` → unchanged
- `/search?q=ratio=3:2` (raw `:` not treated as a port) → unchanged
- `/items/id%3A42` (in path segment) → unchanged
- `/a?x=1%3A1&y=2%3A2` (multiple params) → unchanged
- `   /q?t=a%3Ab   ` → trimmed then echoed `/q?t=a%3Ab`

Returns `fallback` (the colon is irrelevant — the leading char fails the guard):
- `https://evil.example.com/path` → fallback (starts with `h`)
- `javascript:alert(1)` → fallback (starts with `j`)
- `//host%3A8080/path` → fallback (explicit `startsWith("//")` rejection)

## Security framing

The guard's protection comes entirely from the leading-character allow-list, not
from any colon analysis. Scheme/pseudo-scheme/protocol-relative payloads are
rejected because they don't begin with exactly one `/`, so their colons never
matter. Conversely, a same-origin href keeps its encoded colons byte-for-byte —
the function is a navigation origin gate, not a URL parser or sanitizer. The
tests deliberately do **not** claim any decoding, query extraction, or port
detection the code does not perform.

## Truth-first scope note

The original task referenced `hushh-research/__tests__/navigation/...` and
alias is `@/lib/navigation/routes`. The file therefore lives at
`hushh-webapp/__tests__/navigation/resolve-encoded-colons.test.ts`.

## Verification

- `npm test -- __tests__/navigation/resolve-encoded-colons.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — pins the real "no query parsing / no decoding / leading-char
  allow-list" behavior; explicitly corrects the "query extractor / protocol-port
  isolation" framing the code does not implement

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-encoded-colons.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-encoded-colons.test.ts
@@ -1,0 +1,92 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) against percent-encoded colons (%3A)
+// embedded inside query/parameter text values.
+//
+// TRUTH-FIRST — LITERAL CONTRACT: resolveInternalRouteHref is a one-line wrapper:
+//
+//   return normalizeInternalRouteHref(value) ?? fallback;
+//
+// and normalizeInternalRouteHref is purely:
+//
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
+//   if (!href.startsWith("/") || href.startsWith("//")) return null;
+//   if (/[\r\n]/.test(href)) return null;
+//   return href;
+//
+// IMPORTANT CORRECTION to the framing: there is NO "search query extractor" here.
+// resolveInternalRouteHref does NOT parse query strings, does NOT split on "?",
+// "&", or "=", does NOT percent-decode, and has NO concept of protocol or port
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `3a7b71d4`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3718, opened 2026-06-26
test(utils): characterize formatCompleteJson array element key-value inversion

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/utils/json-to-human.ts`) for the generic array branch when
array elements are objects whose keys are stringified indices
(e.g. `[{ "0": "a" }]`). Test-only; no production change.

```ts
for (const item of sectionValue.slice(0, 3)) {
  if (typeof item === "object" && item !== null) {
    const obj = item as Record<string, unknown>;
    const firstValue = Object.values(obj).find(v => v !== null && v !== undefined);
    lines.push(`  • ${String(firstValue || "Item")}`);
  } else {
    lines.push(`  • ${String(item)}`);
  }
}
if (sectionValue.length > 3) lines.push(`  ... and ${sectionValue.length - 3} more`);
```

The task framing ("indexes hold objects that reuse the stringified indices as
explicit key names", "isolates these matching value tags") implies the formatter
reads or correlates object **keys**. It does not. In the generic array branch the
formatter **never reads keys** — it takes the first non-null/undefined **value**
via `Object.values(obj).find(...)`. So `{ "0": "a" }` and `{ foo: "a" }` render
**identically**. There is no "matching value tag isolation"; the only thing the
layout records is the first kept value. The tests pin that reality rather than
the implied (incorrect) key-correlation behavior.

## Behavior pinned (8 cases)

- `[{ "0": "a" }]` → `--- Items (1 items) ---` / `• a` (value rendered, key ignored)
- `[{ "0": "a" }]` renders **identically** to `[{ foo: "a" }]`
- `[{ "0": "a", "1": "b" }]` → `• a` (first value only)
- 4 index-named objects → first 3 bullets + `... and 1 more`
- `[{ "0": 0 }]` → `• Item` — sharp edge: fallback is `firstValue || "Item"`
  (truthiness), so a falsy-but-present `0` is replaced by `"Item"`
- `[{ "0": "" }]` → `• Item` (same `||` truthiness fallback)
- `[{ "0": false }]` → `• Item` (same)
- `[{ "0": null, "1": "b" }]` → `• b` (`.find()` skips null, keeps next value)

## Sharp edge

`String(firstValue || "Item")` uses truthiness, **not** nullish coalescing.
`.find(v => v !== null && v !== undefined)` will legitimately select `0`/`""`/`false`,
but the subsequent `|| "Item"` then discards them — so a present falsy first value
displays as the literal `"Item"`. This is documented, not asserted as desirable.

## Truth-first scope note

`hushh-webapp/__tests__/utils/format-array-inversion.test.ts`.

## Verification

- `npm test -- __tests__/utils/format-array-inversion.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — corrects the "key correlation" premise and pins the real
  first-value / truthiness-fallback behavior

### DIFF
--- a/hushh-webapp/__tests__/utils/format-array-inversion.test.ts
+++ b/hushh-webapp/__tests__/utils/format-array-inversion.test.ts
@@ -1,0 +1,86 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) for the GENERIC array branch when
+// array elements are objects whose keys are stringified indices (e.g.
+// `[{ "0": "a" }]`).
+//
+// TRUTH-FIRST — LITERAL CONTRACT (generic, non-special-cased section key):
+//
+//   for (const item of sectionValue.slice(0, 3)) {
+//     if (typeof item === "object" && item !== null) {
+//       const obj = item as Record<string, unknown>;
+//       const firstValue = Object.values(obj).find(v => v !== null && v !== undefined);
+//       lines.push(`  • ${String(firstValue || "Item")}`);
+//     } else {
+//       lines.push(`  • ${String(item)}`);
+//     }
+//   }
+//   if (sectionValue.length > 3) lines.push(`  ... and ${sectionValue.length - 3} more`);
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `af80ff06`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3722, opened 2026-06-26
test(theme-toggle): characterise resolveActiveTheme normalisation contract

M-Rajith:
Characterises the resolveActiveTheme normalisation boundary in ThemeToggle:
- Returns correct ThemeOption for "light", "dark", "system"
- Falls back to "system" for undefined or unknown values
- Normalises whitespace and casing before comparison

Attach point: hushh-webapp/components/theme-toggle.tsx
Single file, single surface. No DOM rendering or mocks needed.

### DIFF
--- a/hushh-webapp/__tests__/components/theme-toggle.test.tsx
+++ b/hushh-webapp/__tests__/components/theme-toggle.test.tsx
@@ -1,15 +1,5 @@
 import { describe, expect, it } from "vitest";
-
-// resolveActiveTheme is not exported, so we re-implement the contract inline.
-// This characterises the normalisation behaviour at the boundary.
-type ThemeOption = "light" | "dark" | "system";
-function resolveActiveTheme(theme: string | undefined): ThemeOption {
-  const normalized = (theme ?? "").trim().toLowerCase();
-  if (normalized === "light" || normalized === "dark" || normalized === "system") {
-    return normalized as ThemeOption;
-  }
-  return "system";
-}
+import { resolveActiveTheme } from "@/components/theme-toggle";
 
 describe("resolveActiveTheme", () => {
   it("returns light for light input", () => {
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: self-contained test re-implements the contract instead of importing it

### Direction
Thank you for adding characterisation tests around theme normalisation — pinning down the "light / dark / system, fall back to system, trim + lowercase" contract is exactly the right instinct, and the case coverage here (undefined, unknown value, whitespace + casing) is genuinely thorough. That edge-case thinking is what makes a test suite valuable, so the work to merge is small and worth finishing.

### Blocker
The test re-implements `resolveActiveTheme` inline rather than importing it from production code (your own comment notes "resolveActiveTheme is not exported, so we re-implement the contract inline"). Because the assertions run against the copy in the test file, the suite can pass forever while the real `resolveActiveTheme` in the theme-toggle component drifts or regresses — it exercises no production code path, so it cannot protect the behaviour it describes.

### Path To Merge
Export the real `resolveActiveTheme` from the theme-toggle source module (a named `export` on the existing function is enough), then `import { resolveActiveTheme } from` that module in the test and delete the inline re-implementation. Keep every one of your existing cases — they become real regression coverage the moment they call the production function.

### Proof Needed
`npx vitest run __tests__/components/theme-toggle.test.tsx` green against the imported (not re-implemented) function, so a future change to the real normalisation logic would actually fail this suite.

M-Rajith:
Exported resolveActiveTheme from the production module and updated the test to import it directly. All 6 cases now exercise the real function — a future regression in the normalisation logic will fail this suite.

kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current blockers: new_export_without_app_or_backend_caller.

### Path To Merge

### Proof Needed

M-Rajith:
resolveActiveTheme is called by ThemeToggle and ThemeToggleCompact within the same module — the export makes the internal normalisation contract testable without duplicating logic. No new external caller is introduced; the function was already the production behaviour, now it's verifiable.

kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current blockers: new_export_without_app_or_backend_caller.

### Path To Merge

### Proof Needed

PR #3730, opened 2026-06-26
test(status-bar-manager): cover null DOM render contract

M-Rajith:
Covers the null DOM render contract of StatusBarManager — the component 
always returns null and mounts nothing into the DOM. It is a native-only 
runtime bridge used in app/providers.tsx.

Attach point: hushh-webapp/app/providers.tsx imports StatusBarManager.
Single file, single surface.


kushaltrivedi5 (approved):

Re-reviewed at head `1a039249`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3742, opened 2026-06-26
test(utils): characterize formatCompleteJson on non-enumerable properties

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/utils/json-to-human.ts`) documenting its exact contract for
properties created via `Object.defineProperty`, specifically **non-enumerable**
properties. Test-only; no production change.

Top-level iteration is `for (const [sectionKey, sectionValue] of Object.entries(json))`,
and nested object iteration is also `Object.entries(...)`. `Object.entries`
returns **only own enumerable string-keyed** properties. `Object.defineProperty`
defaults `enumerable` to `false`, so a property added that way is invisible to
`Object.entries` unless `enumerable: true` is passed explicitly.

## Truth-first note on the task premise

The task asks how the formatter "ignores or parses these hidden attributes." The
literal answer is: it **never parses** them. There is no hidden-attribute reading
path — non-enumerable properties are simply not iterated, so the output is
identical to an object that never had the property. The property boundary is
exactly: own + enumerable + string-keyed.

## Behavior pinned (7 cases)

- Top-level non-enumerable `account_metadata` → `""` (skipped entirely)
- Same key with `enumerable: true` → `\n--- Account Information ---\n  Account Number: 123`
  (boundary control proving the value itself is renderable; only enumerability gates it)
- Non-enumerable sibling alongside an enumerable section → only the enumerable
  section renders
- An object with a non-enumerable key renders **identically** to one that never
  had the key (`toBe` equality of both outputs)
- **Nested** non-enumerable property (`ssn` inside a section object) → invisible,
  because nested iteration also uses `Object.entries`
- All-non-enumerable object → `""`
- **Non-enumerable accessor (getter)** → `""` **and the getter is never invoked**
  (`getterCalls === 0`), proving `Object.entries` never even touches it

## Header-format correction (caught locally)

My first draft asserted `=== Label ===`; the local run showed the real header is
`--- Label ---`. Expectations were corrected to the actual output before the
final green run — pinning real behavior, not an assumed format.

## Truth-first scope note

`hushh-webapp/__tests__/utils/format-json-enumerable.test.ts`.

## Verification

- `npm test -- __tests__/utils/format-json-enumerable.test.ts` → 7/7 pass

## CI gate checklist

- [x] Truth-First — pins the real "enumerable-own-string only; non-enumerable
  skipped, getter never invoked" boundary

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-enumerable.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-enumerable.test.ts
@@ -1,0 +1,113 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) for properties defined via
+// `Object.defineProperty` — specifically NON-ENUMERABLE properties.
+//
+// TRUTH-FIRST — LITERAL CONTRACT:
+//
+// Top-level iteration is `for (const [sectionKey, sectionValue] of
+// Object.entries(json))`, and nested object iteration is also
+// `Object.entries(...)`. `Object.entries` returns ONLY own ENUMERABLE
+// string-keyed properties. `Object.defineProperty` defaults `enumerable` to
+// `false`, so a property added that way is invisible to `Object.entries` unless
+// `enumerable: true` is explicitly passed.
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `b090646c`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3743, opened 2026-06-26
test(navigation): characterize normalizeInternalRouteHref verbatim %3F preservation (no decode, no query split)

abdulbeast4-creator:
## Summary

Adds an isolated characterization test for `normalizeInternalRouteHref` (`hushh-webapp/lib/navigation/routes.ts`) documenting how the path guard handles percent-encoded question marks (`%3F`) embedded as raw literal text inside the path (e.g. `/legal/faq%3Fpage`) rather than as a query-string marker. Test-only; no production change.

The blocker is addressed by tightening the claim to the verified surface. The prior title ("percent-encoded question marks") could read as if the util performs `?`/query handling on `%3F`. It does not — the real surface is **verbatim preservation with no decode and no query splitting** (identity for any href that passes the guard). The title now states that exact boundary so the claim matches what the test pins.

```ts
if (!href) return null;
return href;
```

The guard does **no percent-decoding, no `?`/query splitting, and no re-encoding**. It only checks: non-empty, single leading slash (rejecting protocol-relative `//`), and absence of CR/LF. For any internal href that passes those three checks, `normalizeInternalRouteHref` is an **identity function** — so a literal `%3F` token is preserved byte-for-byte.

## Truth-first note on the task premise

The task asks whether the routing utility "preserves the literal text token exactly." It does — and the reason is the more important boundary: there is no decode/split path at all, so `%3F` is never interpreted as a `?`. The util is a pure allow/deny gate on the trimmed string, not a URL parser. The tests pin both the preservation and the absence of any decoding/normalization.

## Behavior pinned (9 cases)

- `/legal/faq%3Fpage` → returned verbatim
- `%3F` is **not** decoded into `?` (output contains `%3F`, not `?`)
- lowercase `%3f` preserved with no case normalization
- multiple `%3F` tokens preserved (`/a%3Fb%3Fc`)
- a `%3F` literal **plus** a real `?tab=1` query string both preserved verbatim (no merging/splitting)
- leading/trailing whitespace trimmed while the inner `%3F` literal is kept
- protocol-relative `//evil.example/faq%3Fpage` → `null` (guard still applies)
- non-rooted `legal/faq%3Fpage` → `null` (guard still applies)
- `%3F`-bearing href containing `\n` → `null` (CR/LF guard still applies)

## Truth-first scope note

The original task referenced `hushh-research/__tests__/navigation/...`, an import from `lib/navigation/routes.ts`, and `npm test -- hushh-research/...`. There is no top-level `__tests__` in this repo; vitest only includes `__tests__/**` under `hushh-webapp`, and the module alias is `@/lib/navigation/routes`. The file therefore lives at `hushh-webapp/__tests__/navigation/normalize-encoded-questions.test.ts` and imports via the `@/` alias.

## Verification

- `npm test -- __tests__/navigation/normalize-encoded-questions.test.ts` → 9/9 pass

## CI gate checklist

- [x] Truth-First — pins the real "no decode/no split; identity for passing hrefs; `%3F` preserved literally" boundary
- [x] Claim matches surface — title corrected to verbatim `%3F` preservation (no decode, no query split)

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-encoded-questions.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-encoded-questions.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) for percent-encoded question marks
+// (`%3F`) embedded as raw literal text inside the path, NOT as a query marker.
+//
+// TRUTH-FIRST — LITERAL CONTRACT:
+//
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
+//   if (!href.startsWith("/") || href.startsWith("//")) return null;
+//   if (/[\r\n]/.test(href)) return null;
+//   return href;
+//
+// The guard performs NO percent-decoding, NO query/`?` splitting, and NO
+// re-encoding. It only checks: non-empty, single leading slash, and absence of
+// CR/LF. Therefore a literal `%3F` token inside the path text is preserved
+// byte-for-byte — `normalizeInternalRouteHref` is an identity function for any
+// internal href that passes the three guards.
+
+describe("normalizeInternalRouteHref — percent-encoded question marks (%3F)", () => {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #3745, opened 2026-06-26
test(navigation): characterize resolveInternalRouteHref parameter array empty strings

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) documenting its exact behavior when a
candidate href carries repeated array-style query params with consecutive empty
strings, e.g. `/api/v1?filters=&filters=&filters=active`. Test-only; no
production change.

```ts
}

// normalizeInternalRouteHref:
if (!href) return null;
return href;
```

The task asks how "the extraction process constructs and retains these blank
positions." **There is no extraction process.** `resolveInternalRouteHref` does
no query parsing, no array splitting, no decoding, and no construction of any
position list. It is a pure allow/deny gate over the trimmed string:

- If the href passes the three guards (non-empty, single leading `/`, no CR/LF),
  it is returned **verbatim** — the entire query block, including every empty
  `filters=` slot, is preserved exactly in place.
- If it fails any guard, the supplied **fallback** is returned unchanged.

So the consecutive empty array positions are "retained" only in the trivial sense
that the whole string is passed through untouched — not because any parser
models them. These tests pin that real boundary so future readers don't assume a
URL/array parser exists here.

## Behavior pinned (10 cases)

- `/api/v1?filters=&filters=&filters=active` returned verbatim
- repeated empty params are not collapsed/deduped/reordered (3 `filters=` kept)
- fully-empty array `filters=&filters=&filters=` preserved verbatim
- leading empty position `?a=&a=1` preserved
- bracketed array syntax `?a[]=&a[]=&a[]=z` preserved verbatim
- outer whitespace trimmed while empty array positions stay intact
- protocol-relative `//evil.example/...?filters=&filters=` → fallback
- non-rooted `api/v1?filters=&filters=` → fallback
- newline-bearing `?filters=&filters=\ninject` → fallback
- `null`/empty input → fallback

## Truth-first scope note

The original task referenced `hushh-research/__tests__/navigation/...`, a bare
`lib/navigation/routes.ts` import, and `npm test -- hushh-research/...`. There is
`hushh-webapp`, and the module alias is `@/lib/navigation/routes`. The file
therefore lives at
`hushh-webapp/__tests__/navigation/resolve-array-empty-strings.test.ts` and
imports via the `@/` alias.

## Verification

- `npm test -- __tests__/navigation/resolve-array-empty-strings.test.ts` → 10/10 pass

## CI gate checklist

- [x] Truth-First — corrects the "extraction/construction" premise; pins the real
  "no parse; verbatim passthrough or fallback" contract

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-array-empty-strings.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-array-empty-strings.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) for query strings that carry repeated
+// array-style parameters with consecutive empty-string values, e.g.
+// `/api/v1?filters=&filters=&filters=active`.
+//
+// TRUTH-FIRST — LITERAL CONTRACT:
+//
+//   export function resolveInternalRouteHref(value, fallback): string {
+//     return normalizeInternalRouteHref(value) ?? fallback;
+//   }
+//
+// and normalizeInternalRouteHref:
+//
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
+//   if (!href.startsWith("/") || href.startsWith("//")) return null;
+//   if (/[\r\n]/.test(href)) return null;
+//   return href;
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `239997b6`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3748, opened 2026-06-26
test(utils): characterize formatCompleteJson on native error instances

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/utils/json-to-human.ts`) documenting its exact output when
native JavaScript `Error` instances are embedded directly as dictionary values
(e.g. `new Error("failed task")`). Test-only; no production change.

## Truth-first mapping baseline

`formatCompleteJson` walks every level with `Object.entries(...)`, which returns
only **own + enumerable + string-keyed** properties. On a native `Error`:

- `message` and `stack` are **own but non-enumerable**
- `name` lives on the **prototype** (non-enumerable)

So `Object.entries(error)` yields **nothing** for a bare Error. The function
therefore treats an Error like an empty object — it emits the section/label
header and **no** message or stack lines. The built-in message and the full
stack trace are **never** rendered. Error content surfaces **only** through
properties that code explicitly assigns (e.g. `err.code = "E_FAIL"`), which are
own + enumerable.

## Behavior pinned (8 cases)

- bare Error as a top-level section → `"\n--- Failure ---"` only
- Error message never leaks into output
- Error stack trace never leaks (`stack` / `Error:` / `.test.ts` absent)
- only explicitly-assigned own enumerable props render (`code`, `detail`)
- nested Error value → `"Cause:"` label header with no children, message hidden
- Error in an array (generic branch) → falls back to literal `"• Item"`
- Error in an array with an assigned enumerable value → uses that first value
- subclass that assigns `this.name`/`this.status` → those render (assignment
  creates an own enumerable `name` that shadows the prototype), but the `super()`
  message stays hidden

## Truth-first correction recorded during local verification

The local run corrected one of my initial assumptions: I expected a subclass's
`name` to stay hidden, but assigning `this.name = "HttpError"` in the constructor
creates an **own enumerable** `name` that shadows the prototype's non-enumerable
one — so it **does** render. The test now pins that real behavior (assigned
fields render; only the built-in `super()` message stays non-enumerable/hidden).

## Truth-first scope note

`'formatCompleteJson'` import. There is no top-level `__tests__` here; vitest only
`hushh-webapp/__tests__/utils/format-json-errors.test.ts` and imports via `@/`.

## Verification

- `npm test -- __tests__/utils/format-json-errors.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — pins the real "Object.entries → own+enumerable only; built-in
  message/stack hidden" mapping for Error instances

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-errors.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-errors.test.ts
@@ -1,0 +1,125 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) when native JavaScript Error
+// instances are embedded directly as dictionary values.
+//
+// TRUTH-FIRST — LITERAL CONTRACT:
+//
+// formatCompleteJson iterates exclusively with `Object.entries(...)` at every
+// level. `Object.entries` returns only OWN + ENUMERABLE + string-keyed
+// properties. On a native Error, the standard `message`, `stack`, and `name`
+// properties are NON-ENUMERABLE (message/stack are own-but-non-enumerable; name
+// lives on the prototype). Therefore:
+//
+//   - An Error passed as a top-level SECTION value is `typeof === "object"` and
+//     not an array, so it emits the section header block:
+//         ""                       (blank separator line)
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `7ba38e5b`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3752, opened 2026-06-26
test(navigation): characterize normalizeInternalRouteHref windows dynamic path slashes

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) documenting how it handles path
sequences that resemble Windows file delimiters / backslashes, e.g.
`/app\legal\privacy`. Test-only; no production change.

```ts
  if (!href) return null;
  return href;
}
```

The task asks whether the routing engine "normalizes [backslashes] into standard
web forward slashes." **It does not.** `normalizeInternalRouteHref` performs **no
slash normalization** — it never converts `\` to `/` and never splits on path
delimiters. It is a pure allow/deny gate, and a backslash is just an ordinary
character to every check:

- `trim()` strips only outer ASCII whitespace (backslashes untouched)
- `startsWith("/")` / `startsWith("//")` inspect only the first 1–2 chars
- the reject regex matches only CR/LF, **not** backslash

So any rooted href (single leading `/`, not `//`, no CR/LF) that contains
backslashes is returned **byte-for-byte verbatim** — the engine treats them as
**literal characters**. The correct answer to the task's question is: **literal,
never normalized.**

## Behavior pinned (11 cases)

- `/app\legal\privacy` returned verbatim (no `\`→`/` conversion)
- explicit assertion that output is NOT `/app/legal/privacy` and still contains `\`
- mixed `/app/legal\privacy` preserved exactly
- `/\legal` (leading `/` then `\`) preserved
- leading backslash `\app\legal` → `null` (doesn't start with `/`)
- whitespace-padded leading backslash → `null`
- outer whitespace trimmed, interior `\` intact
- trailing backslash `/app\` preserved
- consecutive/UNC-style `/\\server\share` preserved
- protocol-relative `//host\path` → `null`
- CR/LF injection alongside backslashes → `null`

## Truth-first scope note

`'normalizeInternalRouteHref'` import. There is no top-level `__tests__` here;
`hushh-webapp/__tests__/navigation/normalize-windows-slashes.test.ts` and imports
via the `@/` alias.

## Verification

- `npm test -- __tests__/navigation/normalize-windows-slashes.test.ts` → 11/11 pass

## CI gate checklist

- [x] Truth-First — corrects the "normalizes backslashes to forward slashes"
  premise; pins the real "literal passthrough / no normalization" contract

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-windows-slashes.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-windows-slashes.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) for path sequences that resemble
+// Windows file delimiters / backslashes, e.g. "/app\\legal\\privacy".
+//
+// TRUTH-FIRST — LITERAL CONTRACT:
+//
+//   export function normalizeInternalRouteHref(value): string | null {
+//     const href = String(value ?? "").trim();
+//     if (!href) return null;
+//     if (!href.startsWith("/") || href.startsWith("//")) return null;
+//     if (/[\r\n]/.test(href)) return null;
+//     return href;
+//   }
+//
+// CORRECTION TO THE TASK PREMISE: there is NO slash normalization. The utility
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `3951db7a`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3753, opened 2026-06-26
test(navigation): characterize resolveInternalRouteHref encoded ampersand text bounds

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) documenting how it handles a
percent-encoded ampersand (`%26`) inside a single query field's string value,
e.g. `/one?utm_campaign=spring%26summer`. Test-only; no production change.

```ts
}

  if (!href) return null;
  return href;
}
```

The task frames this as "tracking and extraction parameters within
resolveInternalRouteHref when analyzing query properties." `resolveInternalRouteHref`
**does no query analysis at all.** It does **not** parse, decode, split, or
re-serialize the query string — there is **no `URLSearchParams` round-trip**. It
is a thin same-origin safety wrapper that returns either the trimmed href
**verbatim** (when it passes the rooted/no-`//`/no-CR-LF gate) or the caller's
`fallback`.

Therefore a `%26` inside a query value is preserved **byte-for-byte** as the
three literal characters `%26` — it is never decoded into `&`, and the value is
never split into two parameters. The "tracking chain stays intact" precisely
**because nothing tokenizes it**, not because of any deliberate ampersand
handling.

## Behavior pinned (8 cases)

- `/one?utm_campaign=spring%26summer` returned verbatim (no decode, no split)
- explicit assertion output keeps `%26` and is NOT `spring&summer`
- `/one?ref=a%26utm_source=b` → no second param introduced (`split("&").length === 1`)
- `/one?a=x%26y&b=z` → real `&` splits to 2 fields, `%26` does not add a third
- outer whitespace trimmed, encoded ampersand intact
- lowercase `%26` preserved exactly (no escape-case normalization)
- fails-gate inputs (`//evil.example?q=a%26b`, non-rooted `one?q=a%26b`) → fallback
- nullish input → fallback

## Truth-first scope note

`'resolveInternalRouteHref'` import. There is no top-level `__tests__` here;
`hushh-webapp/__tests__/navigation/resolve-encoded-ampersands.test.ts` and imports
via the `@/` alias.

## Verification

- `npm test -- __tests__/navigation/resolve-encoded-ampersands.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — corrects the "query analysis / tracking extraction" premise;
  pins the real "literal passthrough, no decode/split" contract for `%26`

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-encoded-ampersands.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-encoded-ampersands.test.ts
@@ -1,0 +1,96 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) for query segments that contain a
+// percent-encoded ampersand (%26) inside a single field's string parameter,
+// e.g. "/one?utm_campaign=spring%26summer".
+//
+// TRUTH-FIRST — LITERAL CONTRACT:
+//
+//   export function resolveInternalRouteHref(value, fallback): string {
+//     return normalizeInternalRouteHref(value) ?? fallback;
+//   }
+//
+//   export function normalizeInternalRouteHref(value): string | null {
+//     const href = String(value ?? "").trim();
+//     if (!href) return null;
+//     if (!href.startsWith("/") || href.startsWith("//")) return null;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `5fb2dc5c`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3754, opened 2026-06-26
test(navigation): characterize normalizeInternalRouteHref relative dot segment chains

abdulbeast4-creator:
## Summary

Pins and documents the current characterization of `normalizeInternalRouteHref`
(`hushh-webapp/lib/navigation/routes.ts`) when a route path is built from
redundant single-dot navigation components, e.g. `/./legal/./privacy/./`.

```ts
  if (!href) return null;
  return href;
}
```

The task asks whether the path guard "normalizes or validates" dot chains. The
verified answer: it **validates only — it does not normalize.** There is **no
path resolution, no segment collapsing, no canonicalization.** A `.` segment is
treated as an ordinary character. Any value that starts with a single `/` (not
`//`) and contains no CR/LF is returned **byte-for-byte verbatim**, so
`/./legal/./privacy/./` stays exactly as written — it is **NOT** reduced to
`/legal/privacy`. The data is handled as a standard string fragment, never
resolved as filesystem-style relative navigation.

## Behavior pinned (10 cases)

- `/./legal/./privacy/./` returned verbatim (no collapse)
- explicit assertion: output is NOT `/legal/privacy` and still contains `/./`
- leading `/./` prefix preserved
- trailing `/.` preserved
- consecutive `/./././x` preserved with no dedupe
- bare `/.` accepted verbatim (starts with one slash)
- `./legal/privacy` rejected → `null` (does not start with `/`)
- outer whitespace trimmed, interior dot segments intact
- `//./legal` rejected → `null` (protocol-relative gate)
- CR/LF alongside dot segments rejected → `null`

## Truth-first scope note

`'normalizeInternalRouteHref'` import. There is no top-level `__tests__` here;
`hushh-webapp/__tests__/navigation/normalize-dot-chains.test.ts` and imports via
the `@/` alias.

## Verification

- `npm test -- __tests__/navigation/normalize-dot-chains.test.ts` → 10/10 pass

## CI gate checklist

- [x] Truth-First — corrects the "normalizes" premise; pins the real
  "validate-only, literal passthrough, no dot-segment resolution" contract

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-dot-chains.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-dot-chains.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) for route paths built from
+// consecutive relative single-dot navigation markers, e.g.
+// "/./legal/./privacy/./".
+//
+// TRUTH-FIRST — LITERAL CONTRACT:
+//
+//   export function normalizeInternalRouteHref(value): string | null {
+//     const href = String(value ?? "").trim();
+//     if (!href) return null;
+//     if (!href.startsWith("/") || href.startsWith("//")) return null;
+//     if (/[\r\n]/.test(href)) return null;
+//     return href;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `8d24469b`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3763, opened 2026-06-26
test(auth-provider-button): assert explicit button type contract

imsharukhan:
## Summary

Adds a focused test for `AuthProviderButton` asserting that the rendered
button exposes `type="button"`.

## What & Why

`AuthProviderButton` explicitly passes `type="button"` to the underlying
Button component to prevent accidental form submission when rendered
inside forms.

This test characterizes that existing behavior and helps prevent future
regressions.

## Scope

- New test file only
- No production code changes
- No existing tests modified
- One deterministic assertion
- No snapshots

## Validation

```bash
npx vitest run __tests__/components/auth-provider-button.test.tsx
```

## Checklist

- [x] Test-only change
- [x] New test file
- [x] No snapshots
- [x] Deterministic

### DIFF
--- a/hushh-webapp/__tests__/components/auth-provider-button.test.tsx
+++ b/hushh-webapp/__tests__/components/auth-provider-button.test.tsx
@@ -1,0 +1,35 @@
+import { render } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested

PR #3788, opened 2026-06-26
test(ui): cover textarea disabled state contract

gracyezhilkumar07:
## Summary

Adds test coverage for the Textarea component's disabled state contract.

## Changes Made

* Added a unit test to verify that the `Textarea` component correctly renders in a disabled state.
* Confirms that the rendered `<textarea>` element exposes the `disabled` attribute when the `disabled` prop is provided.
* Helps prevent regressions in the component's disabled behavior.

## Test

```bash
npx vitest run __tests__/ui/textarea.test.tsx
```

### DIFF
--- a/hushh-webapp/__tests__/ui/textarea.test.tsx
+++ b/hushh-webapp/__tests__/ui/textarea.test.tsx
@@ -4,37 +4,12 @@ import { describe, expect, it } from "vitest";
 import { Textarea } from "@/components/ui/textarea";
 
 describe("Textarea", () => {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `59bb4d97`: test-only coverage change on a non-sensitive UI path, exercises real production code, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue (CI re-runs as the backstop).

PR #3797, opened 2026-06-26
test(table): verify Table container has role='region' and aria-label

imsharukhan:
## Summary

Adds one focused accessibility contract test verifying that the `Table`
scroll container exposes the expected accessibility attributes.

## What

Appends a single `it()` block to the existing table test suite.

`data-slot="table-container"` exposes:

- `role="region"`
- `aria-label="Scrollable table"`

## Why

`components/ui/table.tsx` already renders these accessibility attributes
on the table container. The existing suite verifies other aspects of the
container but does not explicitly lock these attributes. This test
documents and protects that behavior.

## Scope

- Test-only change
- One new `it()` block
- No production code changes
- No import changes
- No snapshots
- No mocks

## Validation

```bash
npx vitest run __tests__/ui/table.test.tsx
```

## Checklist

- [x] Test-only change
- [x] Append-only update
- [x] No snapshots
- [x] Deterministic

### DIFF
--- a/hushh-webapp/__tests__/ui/table.test.tsx
+++ b/hushh-webapp/__tests__/ui/table.test.tsx
@@ -54,4 +54,26 @@ describe("Table", () => {
     expect(header).toBeTruthy();
     expect(header?.tagName).toBe("THEAD");
   });
+
+  it("renders the table container with role='region' and an aria-label", () => {
+    const { container } = render(
+      <Table>
+        <tbody>
+          <tr>
+            <td>Holding</td>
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `70f796a3`: test-only coverage change on a non-sensitive UI path, exercises real production code, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue (CI re-runs as the backstop).

PR #3802, opened 2026-06-26
test(separator): verify Separator renders as a div element

imsharukhan:
## Summary

Adds one focused contract test verifying that the `Separator` component
renders as a native `<div>` element.

## What

Appends a single `it()` block to the existing separator test suite.

`data-slot="separator"` renders as:

- `<div>`

## Why

`Separator` wraps Radix's `SeparatorPrimitive.Root`, which renders a native
`<div>` by default. The existing suite already verifies accessibility
attributes such as `role="separator"` and `data-orientation`, but it does
not explicitly protect the underlying HTML element. This test documents and
preserves that DOM contract.

## Scope

- Test-only change
- One new `it()` block
- No production code changes
- No import changes
- No snapshots
- No mocks

## Validation

```bash
npx vitest run __tests__/ui/separator.test.tsx
```

## Checklist

- [x] Test-only change
- [x] Append-only update
- [x] No snapshots
- [x] Deterministic

### DIFF
--- a/hushh-webapp/__tests__/ui/separator.test.tsx
+++ b/hushh-webapp/__tests__/ui/separator.test.tsx
@@ -40,4 +40,12 @@ describe("Separator", () => {
     expect(el?.getAttribute("role")).toBe("separator");
   });
 
+  it("renders as a div element", () => {
+    const { container } = render(<Separator />);
+
+    const el = container.querySelector('[data-slot="separator"]');
+
+    expect(el?.tagName).toBe("DIV");
+  });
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `49cc605a`: characterization test importing real prod code (`@/` import present in file head, outside the diff hunk), path-clean (test-only), green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3803, opened 2026-06-26
test(label): verify Label renders as a label element

imsharukhan:
## Summary

Adds one focused contract test verifying that the `Label` component renders
as a native `<label>` element.

## What

Appends a single `it()` block to the existing label test suite.

`data-slot="label"` renders as:

- `<label>`

## Why

`Label` wraps Radix's `LabelPrimitive.Root`, which renders a native
`<label>` element as part of its accessibility contract. The existing
suite already verifies styling and other behavior but does not explicitly
protect the underlying HTML element. This test documents and preserves
that DOM contract.

## Scope

- Test-only change
- One new `it()` block
- No production code changes
- No import changes
- No snapshots
- No mocks

## Validation

```bash
npx vitest run __tests__/ui/label.test.tsx
```

## Checklist

- [x] Test-only change
- [x] Append-only update
- [x] No snapshots
- [x] Deterministic

### DIFF
--- a/hushh-webapp/__tests__/ui/label.test.tsx
+++ b/hushh-webapp/__tests__/ui/label.test.tsx
@@ -11,4 +11,13 @@ describe("Label", () => {
       container.querySelector('[data-slot="label"]')
     ).not.toBeNull();
   });
+
+  it("renders as a label element", () => {
+    const { container } = render(<Label>Email</Label>);
+
+    const label = container.querySelector('[data-slot="label"]');
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `483ba82e`: characterization test importing real prod code (`@/` import present in file head, outside the diff hunk), path-clean (test-only), green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3804, opened 2026-06-26
test(input): verify Input renders as an input element

imsharukhan:
## Summary

Adds one focused contract test verifying that the `Input` component renders
as a native `<input>` element.

## What

Appends a single `it()` block to the existing input test suite.

`data-slot="input"` renders as:

- `<input>`

## Why

`Input` renders a native `<input data-slot="input">` element directly. The
existing suite already verifies other runtime behavior but does not
explicitly protect the underlying HTML element. This test documents and
preserves that DOM contract.

## Scope

- Test-only change
- One new `it()` block
- No production code changes
- No import changes
- No snapshots
- No mocks

## Validation

```bash
npx vitest run __tests__/ui/input.test.tsx
```

## Checklist

- [x] Test-only change
- [x] Append-only update
- [x] No snapshots
- [x] Deterministic

### DIFF
--- a/hushh-webapp/__tests__/ui/input.test.tsx
+++ b/hushh-webapp/__tests__/ui/input.test.tsx
@@ -34,4 +34,13 @@ describe("Input", () => {
 
     expect(el?.getAttribute("autocapitalize")).not.toBe("none");
   });
+
+  it("renders as an input element", () => {
+    const { container } = render(<Input />);
+
+    const el = container.querySelector('[data-slot="input"]');
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `bac6f685`: characterization test importing real prod code (`@/` import present in file head, outside the diff hunk), path-clean (test-only), green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3807, opened 2026-06-26
test(textarea): verify Textarea renders as a textarea element

imsharukhan:
## Summary

Adds one focused contract test verifying that the `Textarea` component
renders as a native `<textarea>` element.

## What

Appends a single `it()` block to the existing textarea test suite.

`data-slot="textarea"` renders as:

- `<textarea>`

## Why

`Textarea` renders a native `<textarea data-slot="textarea">` element
directly. The existing suite already verifies styling and runtime behavior
but does not explicitly protect the underlying HTML element. This test
documents and preserves that DOM contract.

## Scope

- Test-only change
- One new `it()` block
- No production code changes
- No import changes
- No snapshots
- No mocks

## Validation

```bash
npx vitest run __tests__/ui/textarea.test.tsx
```

## Checklist

- [x] Test-only change
- [x] Append-only update
- [x] No snapshots
- [x] Deterministic

### DIFF
--- a/hushh-webapp/__tests__/ui/textarea.test.tsx
+++ b/hushh-webapp/__tests__/ui/textarea.test.tsx
@@ -12,4 +12,13 @@ describe("Textarea", () => {
     expect(textarea).not.toBeNull();
     expect(textarea?.hasAttribute("disabled")).toBe(true);
   });
+
+  it("renders as a textarea element", () => {
+    const { container } = render(<Textarea />);
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `cdb2fd78`: characterization test importing real prod code (`@/` import present in file head, outside the diff hunk), path-clean (test-only), green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3812, opened 2026-06-26
test(utils): cover cn empty value handling

gracyezhilkumar07:
## Summary

Adds unit test coverage for the `cn` utility to verify that empty and falsy values are ignored when composing class names.

## Changes Made

* Added a test to ensure `cn()` ignores `null`, `undefined`, and empty string (`""`) values.
* Verifies that valid class names are preserved while empty inputs do not affect the output.
* Expands existing utility coverage with an additional edge-case scenario.

## Test

```bash
```


kushaltrivedi5 (approved):

Re-reviewed at head `733fce37`: characterization test importing real prod code (`@/` import present in file head, outside the diff hunk), path-clean (test-only), green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3813, opened 2026-06-26
test(utils): characterize formatCompleteJson array object deep property ordering

abdulbeast4-creator:
## Summary

Characterizes how `formatCompleteJson` (`hushh-webapp/lib/utils/json-to-human.ts`)
serializes property/key order when an array contains objects that share an
identical key set but were declared in completely different orders. Test-only;
no production change.

`formatCompleteJson` walks the input exclusively with `Object.entries(...)` and
`Object.values(...)`. It never calls `.sort()`, never alphabetizes, and never
canonicalizes keys. JavaScript object iteration order for ordinary string
(non-integer-like) keys is **insertion order**.

The task frames this as "serialization order metrics … when keys are
structurally out of sequence," implying the formatter might re-sequence or
normalize key order toward a single canonical layout. It does **not.** The
output is **deterministic**, but only because it faithfully **preserves each
object's own insertion order** — it does **not** impose a shared or alphabetized
ordering across array elements. Two objects with the same keys in different
declaration orders serialize to **different** line orders, each matching its own
source object. "Layout determinism" here means *stable passthrough of input
order*, not *canonicalization*.

## Behavior pinned (6 cases)

- generic array section emits `• <first-inserted value>` per item — for
  `{gamma,alpha,bravo}` → `G`, for reversed `{alpha,bravo,gamma}` → `A`
- object section keys printed in insertion order (`Gamma, Alpha, Bravo`),
  explicitly NOT alphabetical
- a second object with a different key order prints in its own order
  (`Charlie, Bravo, Alpha`)
- nested object key order preserved verbatim (`Zulu, Yankee, Xray`)
- determinism: identical input → byte-identical output across repeated calls
- top-level sections emitted in their own insertion order

## Truth-first scope note

`'formatCompleteJson'` import. There is no top-level `__tests__` here; vitest
only includes `__tests__/**` under `hushh-webapp`, and the module alias is
`hushh-webapp/__tests__/utils/format-property-ordering.test.ts` and imports via
the `@/` alias.

## Verification

- `npm test -- __tests__/utils/format-property-ordering.test.ts` → 6/6 pass

## CI gate checklist

- [x] Truth-First — corrects the "re-sequence/canonicalize" premise; pins the
  real "insertion-order passthrough, no sort" contract

### DIFF
--- a/hushh-webapp/__tests__/utils/format-property-ordering.test.ts
+++ b/hushh-webapp/__tests__/utils/format-property-ordering.test.ts
@@ -1,0 +1,105 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) documenting how key/property order
+// is serialized when an array contains objects that share an identical key set
+// but were declared in completely different orders.
+//
+// TRUTH-FIRST — LITERAL CONTRACT: formatCompleteJson walks the input with
+// `Object.entries(...)` / `Object.values(...)` and never sorts, canonicalizes,
+// or re-orders keys. JavaScript object iteration order for string
+// (non-integer-like) keys is INSERTION ORDER. Therefore the formatter is
+// deterministic but it preserves each object's *own declaration order* exactly
+// as given — it does NOT impose a single shared/alphabetized ordering across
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `4b449035`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3814, opened 2026-06-26
test(utils): characterize formatCompleteJson on proto property poisoning attempts

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/utils/json-to-human.ts`) documenting its exact behavior when
the input object carries the classic prototype-poisoning property names
(`__proto__`, `constructor`, `prototype`) as own enumerable keys. Test-only; no
production change.

`formatCompleteJson` is a **read-only serializer**: it only ever READS the input
via `Object.entries(...)` / `Object.values(...)` and appends strings to a local
`lines` array. It performs **no** assignment back onto the input, no
`obj[key] = ...`, no `Object.assign`, and no recursive merge. Therefore it
**cannot pollute `Object.prototype`** regardless of the input keys.

The real runtime path is `tryFormatComplete` → `JSON.parse(...)` →
`formatCompleteJson`. Critically, `JSON.parse('{"__proto__": ...}')` does **not**
reassign the object's prototype — it defines `__proto__` as an **own enumerable
data property**. The tests build payloads via `JSON.parse` to mirror that exact
hazard shape so `Object.entries` actually iterates the dangerous keys.

## Truth-first findings (real behavior pinned, not assumed)

The verification gate surfaced two genuine behaviors I corrected my assertions to
match rather than forcing a convenient expectation:

1. **No prototype pollution — confirmed.** Across `__proto__`, `constructor`, and
   `prototype` payloads (top-level, nested, and combined), a freshly created `{}`
   never sees injected keys and `"x" in Object.prototype` stays `false`. The
   serializer does not write.
2. **Harmless label-lookup quirk — documented.** Section labels are resolved with
   `SECTION_LABELS[key]`, a plain object literal. For `__proto__` and
   `constructor` that read **inherits from the literal's own prototype**, so the
   label becomes the inherited `Object.prototype` (stringified `"[object Object]"`)
   or the native `Object` constructor (`"function Object() { [native code] }"`),
   instead of a cleaned name. This is a cosmetic READ-only artifact in label
   rendering — it does **not** mutate anything and does **not** leak globally. The
   nested values are still rendered normally. The tests pin this real output
   rather than a prettier-but-false expectation.

## Behavior pinned (7 cases)

- `{"__proto__":{...}}` → no `Object.prototype` pollution; payload proto stays
  pristine; `__proto__` is a real own key
- `__proto__` object section label renders as `[object Object]` (inherited-label
  quirk), nested `Alpha: A` still emitted
- scalar `{"__proto__":"danger"}` serialized; `"danger"` not added to prototype
- `{"constructor":{...}}` label renders as native `Object` string;
  `Object.prototype.constructor` unchanged
- `{"prototype":{...}}` renders as ordinary `Prototype` field
- poisoning key nested in a section object → no pollution
- combined `__proto__`/`constructor`/`prototype` payload → does not throw, global
  prototype stays pristine

## Truth-first scope note

`'formatCompleteJson'` import. There is no top-level `__tests__` here; vitest only
`hushh-webapp/__tests__/utils/format-json-poisoning.test.ts` and imports via the
`@/` alias.

## Verification

- `npm test -- __tests__/utils/format-json-poisoning.test.ts` → 7/7 pass

## CI gate checklist

- [x] Truth-First — documents real, existing boundaries: confirmed no
  prototype pollution AND the cosmetic inherited-label quirk for
  `__proto__`/`constructor`, instead of asserting idealized output

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-poisoning.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-poisoning.test.ts
@@ -1,0 +1,130 @@
+import { afterEach, describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) documenting its exact behavior when
+// the input object carries hazardous own keys: "__proto__", "constructor", and
+// "prototype" — the classic prototype-poisoning property names.
+//
+// TRUTH-FIRST — LITERAL CONTRACT: formatCompleteJson is a READ-ONLY serializer.
+// It only ever READS the input via `Object.entries(...)` / `Object.values(...)`
+// and appends strings to a local `lines` array. It performs NO assignment back
+// onto the input, NO `obj[key] = ...`, NO `Object.assign`, and NO recursive
+// merge. Therefore it CANNOT pollute `Object.prototype` regardless of what keys
+// the input contains.
+//
+// REAL RUNTIME PATH: in production these objects arrive via `tryFormatComplete`,
+// which does `JSON.parse(...)` then calls `formatCompleteJson`. Critically,
+// `JSON.parse('{"__proto__": ...}')` does NOT set the object's prototype — it
+// defines "__proto__" as an OWN, ENUMERABLE data property. So these tests build
+// payloads with `JSON.parse` (mirroring the real path) so that the hazardous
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `e1211d91`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3815, opened 2026-06-26
test(navigation): characterize normalizeInternalRouteHref multi-segment hash encoding mixes

abdulbeast4-creator:
## Summary

(`hushh-webapp/lib/navigation/routes.ts`) to map the exact contract boundary when
an internal href combines raw interior spaces with percent-encoded tokens across
multiple `#` hash markers (e.g. `/legal#privacy%20policy#terms of use`).

`normalizeInternalRouteHref` is a **boundary guard, not a content normalizer**.
Its entire body:

```ts
if (!href) return null;
return href;
```

So it ONLY: (1) coerces to string, (2) trims **outer** whitespace, (3) rejects
empty, (4) rejects hrefs not starting with a single `/`, (5) rejects
protocol-relative `//`, (6) rejects CR/LF. On success the href is returned
**byte-for-byte unchanged**.

The task asks whether the utility "normalizes the trailing segment tokens
cohesively" across mixed hash encodings. **It does not.** There is no hash
parsing, no percent-decoding, no space encoding, no segment re-joining, and no
`#` handling whatsoever. Raw interior spaces and multiple `#` markers are
**preserved verbatim**. "Normalize" here means *outer-trim + reject*, never
*canonicalize the fragment*. The tests pin that real boundary rather than an
imagined cohesive-fragment behavior.

## Behavior pinned (10 cases)

- example `/legal#privacy%20policy#terms of use` returned verbatim
- `%20` is NOT decoded inside the fragment
- raw interior spaces are NOT percent-encoded
- every `#` in a multi-hash href is retained (no collapsing/joining)
- only OUTER whitespace is trimmed; interior fragment spaces preserved
- a fragment beginning with a space after `#` is preserved
- CR or LF anywhere (including the fragment) → `null`
- protocol-relative `//...` with mixed hashes → `null`
- a bare `#...` fragment not starting with `/` → `null`
- mixed raw/encoded `+`, `%2B`, `%20` tokens preserved verbatim

## Truth-first scope note

The task referenced `hushh-research/__tests__/navigation/...` and an import from
`lib/navigation/routes.ts`. There is no top-level `__tests__` here; vitest only
`hushh-webapp/__tests__/navigation/normalize-hash-encodings.test.ts` and imports
via the `@/` alias.

## Verification

- `npm test -- __tests__/navigation/normalize-hash-encodings.test.ts` → 10/10 pass

## CI gate checklist

- [x] Truth-First — corrects the "cohesive fragment normalization" premise; pins
  the real "outer-trim + reject, otherwise verbatim passthrough" guard contract

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-hash-encodings.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-hash-encodings.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) pinning its EXACT behavior when an
+// internal href combines raw spaces with percent-encoded tokens across multiple
+// "#" hash boundaries. Test-only; no production change.
+//
+// TRUTH-FIRST — LITERAL CONTRACT: normalizeInternalRouteHref is a BOUNDARY
+// GUARD, not a content normalizer. Its entire body is:
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
+//   if (!href.startsWith("/") || href.startsWith("//")) return null;
+//   if (/[\r\n]/.test(href)) return null;
+//   return href;
+// So it ONLY: (1) coerces to string, (2) trims OUTER whitespace, (3) rejects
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `eed6fc98`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3818, opened 2026-06-26
test(navigation): characterize resolveInternalRouteHref parameter array duplicate keys

abdulbeast4-creator:
## Summary

Documents the exact behavior of `resolveInternalRouteHref`
(`hushh-webapp/lib/navigation/routes.ts`) when a candidate href carries a query
string that mixes a scalar param key with a bracket-notation array of the same
name (e.g. `/api/v1?tags=dev&tags[]=prod`). Test-only; no production change.

`resolveInternalRouteHref` is a one-liner:

```ts
```

and `normalizeInternalRouteHref` only trims **outer** whitespace, rejects
empty / non-`/`-leading / protocol-relative `//` / CR-LF hrefs, and otherwise
returns the value **byte-for-byte unchanged**.

The task asks how "the extraction process resolves the conflict and structured
indexing" between `tags=dev` and `tags[]=prod`. **There is no extraction
process.** `resolveInternalRouteHref` performs zero query parsing: no
`URLSearchParams`, no `?`/`&` splitting, no bracket-notation `[]` handling, no
scalar-vs-array conflict resolution, no dedupe, no structured indexing. The full
query string — duplicate keys, `[]` tokens, ordering — is preserved verbatim on
a valid internal href; an invalid candidate yields the `fallback`. The only
"resolution" is the boundary guard + fallback. The tests pin that real boundary
rather than an imagined parameter-extraction behavior.

## Behavior pinned (10 cases)

- `/api/v1?tags=dev&tags[]=prod` returned verbatim (no extraction)
- original parameter order + both duplicate keys preserved (no "last wins")
- literal `[]` bracket tokens retained (no structured indexing)
- three repeats of the same scalar key are NOT deduplicated
- indexed `tags[0]=...&tags[1]=...` tokens preserved verbatim
- encoded brackets `%5B%5D` are NOT percent-decoded
- protocol-relative `//...` with array keys → fallback
- a bare `?...` query not starting with `/` → fallback
- `null` / whitespace-only input → fallback
- only outer whitespace trimmed; duplicate-key query kept intact

## Truth-first scope note

The task referenced `hushh-research/__tests__/navigation/...` and an import from
`lib/navigation/routes.ts`. There is no top-level `__tests__` here; vitest only
`hushh-webapp/__tests__/navigation/resolve-array-duplicate-keys.test.ts` and
imports via the `@/` alias.

## Verification

- `npm test -- __tests__/navigation/resolve-array-duplicate-keys.test.ts` → 10/10 pass

## CI gate checklist

- [x] Truth-First — corrects the "parameter extraction / structured indexing"
  premise; pins the real "boundary guard + fallback, otherwise verbatim
  passthrough" contract

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-array-duplicate-keys.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-array-duplicate-keys.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) pinning its EXACT behavior when the
+// candidate href carries a query string that mixes a scalar param key with a
+// bracket-notation array of the same name (e.g. '/api/v1?tags=dev&tags[]=prod').
+// Test-only; no production change.
+//
+// TRUTH-FIRST — LITERAL CONTRACT: resolveInternalRouteHref is a one-liner:
+//   return normalizeInternalRouteHref(value) ?? fallback;
+// and normalizeInternalRouteHref only:
+//   const href = String(value ?? "").trim();
+//   if (!href) return null;
+//   if (!href.startsWith("/") || href.startsWith("//")) return null;
+//   if (/[\r\n]/.test(href)) return null;
+//   return href;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `03f2ccb3`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3824, opened 2026-06-26
test(cn): verify cn handles undefined and null arguments without crashing

imsharukhan:
## Summary

Adds a single contract test to the existing `cn` utility test suite.

Verifies that `cn()` returns an empty string when called with `undefined` and `null` arguments.

## Why

The `cn` utility delegates to `clsx` and `tailwind-merge`.

`clsx` treats both `null` and `undefined` as no-op values and omits them from the resulting class string. When no class values remain, the resulting output is an empty string.

Existing tests verify conditional class handling and Tailwind conflict resolution, but do not verify null/undefined input handling.

Adding this assertion helps protect an existing invariant and documents expected behavior for callers.

## Changes

* Added 1 new `it()` block to `__tests__/utils/cn.test.ts`
* Verifies `cn(undefined, null)` returns an empty string
* No production code changes
* No existing tests modified
* No import changes
* No snapshots or mocks added

## Test Plan

```bash
```

All tests pass locally.

## Risk

Low.

This is a test-only change that adds a single assertion to an existing utility test suite and does not affect runtime behavior.

### DIFF
--- a/hushh-webapp/__tests__/utils/cn.test.ts
+++ b/hushh-webapp/__tests__/utils/cn.test.ts
@@ -13,8 +13,13 @@ describe("cn", () => {
     expect(cn("p-2", "p-4")).toBe("p-4");
   });
   it("ignores null, undefined, and empty string values", () => {
-  expect(cn("flex", null, undefined, "", "items-center")).toBe(
-    "flex items-center",
-  );
-});
+    expect(cn("flex", null, undefined, "", "items-center")).toBe(
+      "flex items-center",
+    );
+  });
+
+  it("returns empty string for undefined and null arguments", () => {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `78b8fd84`: characterization test for `cn` undefined/null handling, importing real prod code (`@/lib/utils`), complementary to #3812 (different test case), path-clean (test-only), green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #3827, opened 2026-06-26
test(navigation): characterize resolveInternalRouteHref multi-question hash splits

abdulbeast4-creator:
## Summary
Maps tracking and extraction parameters within resolveInternalRouteHref when analyzing query blocks appended directly onto anchor parameters.

Truth-first correction captured in the test header: there is NO "output extraction contract" and NO query map. `resolveInternalRouteHref` is literally `normalizeInternalRouteHref(value) ?? fallback`. For `/page#anchor?param=123` the value is returned byte-for-byte — the entire `#anchor?param=123` (including the embedded `?`) is opaque fragment text that is never split, parsed, or reordered. The only transforms are trim, single-leading-`/` guard (rejects `//`), and CR/LF rejection. The 11 tests pin exactly this.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-hash-splits.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-hash-splits.test.ts
@@ -1,0 +1,94 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `3e1b93c0`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3829, opened 2026-06-26
test(utils): characterize formatCompleteJson BigInt asset property invariants

abdulbeast4-creator:
## Summary
Establishes serialization structure rules within formatCompleteJson when encountering native Javascript BigInt parameters across discrete data planes.

Truth-first correction captured in the test header: there is NO uniform BigInt contract — behavior is plane-dependent, and a sibling suite (`format-json-bigint.test.ts`) already pins the top-level-skip / nested-`String()` / big-string paths. To avoid duplication, this suite pins the still-uncharacterized asset/currency planes:
- Currency-keyed nested bigint (e.g. `ending_value`) → plain digits via `String()`, NOT `$`/grouped, because the currency gate is `typeof === "number"` and a bigint bypasses it.
- Generic + percentage-keyed nested bigint → `String()` full-precision digits, no `n`.
- `holdings[].market_value` bigint → **throws TypeError** (`formatCurrency` runs `Math.abs(bigint)` unguarded) — the real, unguarded type boundary.
- `asset_allocation[].percentage` bigint → **throws TypeError** (`bigint.toFixed`).
- Top-level bigint → silently skipped.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-bigint-properties.test.ts
+++ b/hushh-webapp/__tests__/utils/format-bigint-properties.test.ts
@@ -1,0 +1,109 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on how native JavaScript
+// BigInt primitives behave across the distinct financial "data planes" of the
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `fe0a764f`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3831, opened 2026-06-26
test(utils): characterize formatCompleteJson array key length density tolerances

abdulbeast4-creator:
## Summary
Characterizes JSON data structure serialization limits inside formatCompleteJson when encountering abnormally large object property strings.

Truth-first correction (captured in the test header): there are no key-length "serialization limits" or "buffer allocation" boundaries to hit. `formatCompleteJson` builds plain strings via `Array.join("\n")` and routes every key through `getFieldLabel`:

```
FIELD_LABELS[key] || key.replace(/_/g, " ").replace(/\b\w/g, c => c.toUpperCase())
```

So a 1,000+ character key is humanized by two pure string transforms only — underscores → single spaces, and the first character of each word upper-cased. The full key length is preserved verbatim (no truncation, ellipsis, or cap), no exception is thrown, and recognized labels (e.g. `ending_value` → `Ending Value`) still take precedence even when surrounded by extreme keys.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-key-lengths.test.ts
+++ b/hushh-webapp/__tests__/utils/format-key-lengths.test.ts
@@ -1,0 +1,94 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) under object configurations whose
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `34c5db13`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3843, opened 2026-06-26
test(routes): characterize buildConnectedSystemRoute String/trim/encodeURIComponent boundaries

imsharukhan:
## What

Adds characterization tests for `buildConnectedSystemRoute` in `lib/navigation/routes.ts`.

## Why

The helper is a pure deterministic function and currently lacks direct test coverage.

These tests document and lock the current behavior for:

* null and undefined inputs
* empty and whitespace-only inputs
* normal system identifiers
* URL encoding via `encodeURIComponent`

## Scope

Test-only change.

No production code modified.

## Validation

```bash
npx vitest run __tests__/navigation/build-connected-system-route.test.ts
```

## Notes

The tests use only the public function interface and characterize existing behavior without changing implementation.

### DIFF
--- a/hushh-webapp/__tests__/navigation/build-connected-system-route.test.ts
+++ b/hushh-webapp/__tests__/navigation/build-connected-system-route.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from "vitest";
+
+import { buildConnectedSystemRoute } from "@/lib/navigation/routes";
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `2d113598`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3845, opened 2026-06-26
test(routes): characterize buildMarketplaceRiaProfileRoute query omission behavior

imsharukhan:
## What

Adds characterization tests for `buildMarketplaceRiaProfileRoute` in `lib/navigation/routes.ts`.

## Why

The helper is a pure deterministic function and currently lacks direct test coverage.

These tests document and lock the current behavior for:

* missing values
* null values
* whitespace-only values
* populated ria identifiers
* query-string serialization through the existing implementation

## Scope

Test-only change.

No production code modified.

## Validation

```bash
npx vitest run __tests__/navigation/build-marketplace-ria-profile-route.test.ts
```

## Notes

The tests use only the public function interface and characterize existing behavior without changing implementation.


kushaltrivedi5 (approved):

Re-reviewed at head `4af9ddee`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3864, opened 2026-06-26
test(routes): verify testProfile route mapping

imsharukhan:
## What

Adds tests for the `testProfile` option in `buildRiaClientWorkspaceRoute`.

## Why

The route builder contains logic that conditionally emits the `test_profile` query parameter and falls back to the clients route when no valid client ID is provided. These branches currently lack direct coverage.

The tests verify:

* `testProfile` enables the `test_profile=1` query parameter
* disabled `testProfile` omits the parameter
* invalid client IDs return the clients route fallback

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/navigation/build-ria-workspace-test-profile.test.ts

## Notes

The tests verify existing route-building behavior through the public API without depending on internal implementation details.

### DIFF
--- a/hushh-webapp/__tests__/navigation/build-ria-workspace-test-profile.test.ts
+++ b/hushh-webapp/__tests__/navigation/build-ria-workspace-test-profile.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, it } from "vitest";
+
+import { buildRiaClientWorkspaceRoute } from "@/lib/navigation/routes";
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `8ce4f653`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3866, opened 2026-06-26
test(routes): verify request route fallback behavior

imsharukhan:
## What

Adds tests for fallback handling in `buildRiaClientRequestRoute`.

## Why

The route builder contains a guard that falls back to the clients route when either required identifier is missing. This behavior currently lacks direct coverage.

The tests verify:

* missing client identifiers trigger the fallback route
* missing request identifiers trigger the fallback route
* valid identifiers produce the expected request route
* `testProfile` emits the expected query parameter on valid routes

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/navigation/build-ria-client-request-fallbacks.test.ts

## Notes

The tests verify existing route-building behavior through the public API without depending on implementation details.

### DIFF
--- a/hushh-webapp/__tests__/navigation/build-ria-client-request-fallbacks.test.ts
+++ b/hushh-webapp/__tests__/navigation/build-ria-client-request-fallbacks.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, it } from "vitest";
+
+import { buildRiaClientRequestRoute } from "@/lib/navigation/routes";
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `dc8b257b`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3886, opened 2026-06-27
test(kai): verify experience contract structure

imsharukhan:
## What

Adds coverage for the exported `KAI_EXPERIENCE_CONTRACT`.

## Why

The experience contract defines the structure used by Kai experience surfaces but currently lacks direct coverage.

The tests verify:

* the expected experience sections are exported
* portfolio clarity exposes the expected contract fields
* decision conviction exposes the expected contract fields

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/lib/kai/experience-contract.test.ts

## Notes

The tests verify the exported contract structure without coupling to implementation details or product copy.

### DIFF
--- a/hushh-webapp/__tests__/lib/kai/experience-contract.test.ts
+++ b/hushh-webapp/__tests__/lib/kai/experience-contract.test.ts
@@ -1,0 +1,77 @@
+import { describe, it, expect } from "vitest";
+import { KAI_EXPERIENCE_CONTRACT } from "@/lib/kai/experience-contract";
+
+describe("KAI_EXPERIENCE_CONTRACT", () => {
+  it("exposes exactly the portfolioClarity and decisionConviction top-level keys", () => {
+    expect(Object.keys(KAI_EXPERIENCE_CONTRACT).sort()).toEqual(
+      ["decisionConviction", "portfolioClarity"].sort()
+    );
+  });
+
+  describe("portfolioClarity", () => {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Appreciated work — adding a structural contract test for `@/lib/kai/experience-contract` is genuinely useful regression coverage, and it correctly imports the real production module rather than a self-mock. Because it touches the **Kai finance runtime** surface, it cannot be auto-merged on the engine's diff-safety gate alone; it's routed to a maintainer security review so the finance-facing contract is verified by a human before it ships.

**Blocker:** The change lands on the Kai finance runtime (`hushh-webapp/__tests__/lib/kai/experience-contract.test.ts` asserting Kai portfolio/decision copy and keys). Per the SOP, any diff on this family requires explicit maintainer/security sign-off (regulated-advice boundary, finance copy contract) that the automated lane cannot grant — even for a test-only PR.

**Path To Merge:** Keep the PR open. A maintainer will run the security/finance evidence lane against the current head; if the contract assertions match the canonical `experience-contract` source and introduce no regulated-advice drift it is approved and queued, otherwise a precise follow-up will be added here.

**Proof Needed:** The test green against the real `@/lib/kai/experience-contract` module (no mocked contract), plus the CI Status Gate green at the current head.

PR #3889, opened 2026-06-27
fix(frontend-a11y): add accessible name to consent search input

KARTHI-77:
## Summary

Add an accessible name to the consent search input.

## Changes

* Add `aria-label="Search consents"` to the consent search input.

## Why

The consent search input previously relied on placeholder text to communicate its purpose. Adding an accessible name improves screen reader support without changing the existing behavior or appearance of the component.

### DIFF
--- a/hushh-webapp/components/consent/consent-center-page.tsx
+++ b/hushh-webapp/components/consent/consent-center-page.tsx
@@ -1918,6 +1918,7 @@ export function ConsentCenterPage() {
                       setSearchValue(next);
                       setParam({ q: next || null, page: "1" });
                     }}
+                    aria-label="Search consents"
                     placeholder={searchPlaceholder}
                     className="pl-9"
                     data-voice-control-id="consent_search"


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Thank you for this contribution — the intent here is sound and it's aimed at the right surface. This change lands on a trust-boundary file (`consent-protocol/api/routes/consent.py`), so before it can merge it needs a maintainer-authored security patch staged on top to make the enforcement independently stronger, not just preserved while touched.

**Blocker:** The diff touches `consent-protocol/api/routes/consent.py`, a consent/identity/finance trust boundary. Per the PR-train write contract, a change on this surface must demonstrably strengthen DB-backed enforcement / scope / caller-credential shape — a maintainer patch will be staged to guarantee that bar before queueing.

**Path To Merge:** Keep the PR open; a maintainer security patch will be staged on top at `consent-protocol/api/routes/consent.py` (co-authored to you), then the combined head is re-reviewed and queued. No action needed from you unless you want to tighten validation/error-surface at that attach point yourself first.

**Proof Needed:** A test asserting the strengthened behavior at `consent-protocol/api/routes/consent.py` (an existing/your/maintainer-added test) green, plus the CI Status Gate green at the patched head.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `d7ecada9`

PR #3896, opened 2026-06-27
test(auth-provider-button): cover provider button type

Divya-rajendran009:
## Summary
- Adds focused coverage for the AuthProviderButton type contract.

## Reviewer Proof
- Surface: components/onboarding/AuthProviderButton.tsx, rendered through the real component.
- No overlap: test-only coverage for provider button type=button.
- DCO signed; base: integration/pr-train.

## Validation
- git diff --cached --check
- Web (Next.js) via GitHub Actions


kushaltrivedi5 (changes requested):
## Changes Requested

PR #3937, opened 2026-06-27
test(utils): characterize formatCompleteJson on native URL instance fields

abdulbeast4-creator:
## Summary
Adds an isolated characterization test for formatCompleteJson to document and verify serialization output paths when encountering native URL objects.

Literal assertions pinned by this test (no generalizations):
- `Object.entries(new URL(...))` and `Object.keys(new URL(...))` are both `[]` — every URL accessor (`href`, `origin`, `hostname`, etc.) lives on `URL.prototype`, not as an own-enumerable key.
- A top-level URL value (`{ source: url }`) produces exactly `"\n--- Source ---"` and no field lines.
- A nested URL value (`{ account_metadata: { website: url } }`) produces only `"  Website:"` with no children.
- `formatCompleteJson` never calls `URL.toString()`: the `href`, hostname, path, query, and hash strings never appear in the output.
- Sibling scalar fields still format normally while the URL section is emptied.
- An array of URL instances yields only the `"--- Links (2 items) ---"` header; no per-URL domain text leaks.

Truth-first correction: the premise that the formatter "maps URL objects safely without stripping domain metadata" is inverted — because a URL exposes zero own-enumerable keys, all domain metadata is in fact stripped. These tests pin that real, existing boundary so any future behavior change is caught.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-urls.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-urls.test.ts
@@ -1,0 +1,95 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) when native JavaScript `URL`
+// instances are embedded inside payload objects.
+//
+// TRUTH-FIRST — CURRENT CONTRACT (verified against source):
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `8d896419`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3938, opened 2026-06-27
fix(frontend-a11y): add accessible name to marketplace search input

KARTHI-77:
## Summary

Add an accessible name to the marketplace search input.

## Changes

* Add `aria-label="Search marketplace"` to the marketplace search input.

## Why

The marketplace search input relied only on placeholder text to communicate its purpose. Adding an accessible name improves support for assistive technologies without changing the existing behavior or appearance of the component.


kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `b3d7dadf`

PR #3939, opened 2026-06-27
fix(frontend-a11y): add accessible name to location request reason field

KARTHI-77:
## Summary

Add an accessible name to the location request reason field.

## Changes

* Add `aria-label="Reason for location request"` to the location request textarea.

## Why

The location request textarea previously relied only on placeholder text to communicate its purpose. Adding an accessible name improves support for assistive technologies without changing the existing behavior or appearance of the component.


kushaltrivedi5 (approved):

Re-reviewed at head `59b32aab`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3946, opened 2026-06-27
fix(frontend-a11y): add accessible name to market search input

KARTHI-77:
## Summary

Add an accessible name to the market search input.

## Changes

* Add `aria-label="Search stocks, ETFs, indices"` to the market search input.

## Why

The market search input previously relied only on placeholder text to communicate its purpose. Adding an accessible name improves support for assistive technologies without changing the existing behavior or appearance of the component.

### DIFF
--- a/hushh-webapp/components/kai/views/kai-market-preview-view.tsx
+++ b/hushh-webapp/components/kai/views/kai-market-preview-view.tsx
@@ -2031,6 +2031,7 @@ export function KaiMarketPreviewView() {
             <Search className="h-[17px] w-[17px] shrink-0 text-[color:var(--one-fg3)]" />
             <input
               type="text"
+              aria-label="Search stocks, ETFs, indices"
               placeholder="Search stocks, ETFs, indices"
               value={marketSearchQuery}
               onChange={(event) => setMarketSearchQuery(event.target.value)}


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Thanks for this — the change touches a real Kai finance surface and the intent is in the right place. It's been routed to the maintainer patch lane because the attach point is a sensitive finance-rendering component that needs a hand-verified change before it can land.

**Blocker:** The edit lands on `hushh-webapp/components/kai/views/kai-market-preview-view.tsx`, a Kai finance/market-rendering surface. A maintainer patch is needed to confirm the change preserves the stale-first/error-parity contract and doesn't alter what financial data the user sees, before it can be approved and queued.

**Path To Merge:** No action required from you right now — a maintainer will either patch this to the canonical shape on `hushh-webapp/components/kai/views/kai-market-preview-view.tsx` (with your co-authorship credited) or come back here with a specific, narrow follow-up if a product decision is needed.

**Proof Needed:** Maintainer patch green against the Kai finance test suite + CI Status Gate green at the patched head.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `1682af1e`

PR #3956, opened 2026-06-27
test(ui): cover accordion component className propagation contract

Roopmann30:
### Description

Adds missing test coverage to `__tests__/ui/accordion.test.tsx` to explicitly verify that the `Accordion` component correctly propagates custom `className` values to its underlying root element.


kushaltrivedi5 (changes requested):
## Changes Requested

**Proof Needed:** `npm run test -- hushh-webapp/__tests__/ui/accordion.test.tsx` green against the real import, and the CI Status Gate green at the new head.

PR #3971, opened 2026-06-27
fix(ui): silence uncaught route error boundary logs in production

Roopmann30:
### Description

Silences the development `console.error` inside the `componentDidCatch` lifecycle of `components/app-ui/route-error-boundary.tsx` by wrapping it in a `NODE_ENV !== "production"` check. This prevents the Error Boundary from leaking the internal React component stack trace to the production client console when an uncaught error occurs, while preserving the user-facing fallback rendering.


kushaltrivedi5 (approved):

Re-reviewed at head `d6e92579`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3973, opened 2026-06-27
fix(ui): silence kai flow import errors in production

Roopmann30:
### Description

Silences a development `console.error` inside `components/kai/kai-flow.tsx` by wrapping it in a `NODE_ENV !== "production"` check. This prevents the Kai Flow component from leaking internal import parsing error stack traces to the production client console when a data ingest fails, while preserving the user-facing error UI state fallback.

### DIFF
--- a/hushh-webapp/components/kai/kai-flow.tsx
+++ b/hushh-webapp/components/kai/kai-flow.tsx
@@ -2640,7 +2640,9 @@ export function KaiFlow({
           return;
         }
 
-        console.error("[KaiFlow] Import error:", err);
+        if (process.env.NODE_ENV !== "production") {
+          console.error("[KaiFlow] Import error:", err);
+        }
         const rawErrorMessage =
           err instanceof Error ? String(err.message || "") : String(err || "");
         const isTransientNetworkLoss =


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Thank you for the contribution here — the work is on a high-trust surface, which is exactly why it gets an extra, careful pass rather than an automatic merge. That's a reflection of where the code lives, not the quality of your PR.

**Blocker:** The change touches `hushh-webapp/components/kai/kai-flow.tsx` (Kai finance/voice-action runtime). Changes on this surface require a security/consent review to confirm they independently strengthen — and never weaken or merely touch — the trust boundary (caller-credential shape, scope enforcement, or data boundary) before they can be approved.

**Path To Merge:** A maintainer security review is queued. If the diff is clean and boundary-positive it will be approved and queued as-is; if a specific guard or validation is missing you'll get a precise, named follow-up here.

**Proof Needed:** Security review sign-off on the Kai finance/voice-action runtime surface + the relevant test green + CI Status Gate green at the current head.

PR #3982, opened 2026-06-27
test(navigation): characterize resolveInternalRouteHref parameter array inverted indexes

abdulbeast4-creator:
## Summary
Documents parameter parsing limits within resolveInternalRouteHref when handling explicitly non-sequential query index assignments (e.g. `/api/v1?items[1]=b&items[0]=a`).

Literal assertions pinned by this test (no generalizations):
- Inverted bracketed indices (`items[1]` before `items[0]`) are returned in their original literal order — no reordering.
- Multiple out-of-order indices (`x[2]`, `x[0]`, `x[1]`) keep positional order exactly; no sort/aggregation into a compiled structure.
- Duplicate indices (`a[0]=first&a[0]=second`) are preserved without deduplication.
- Percent-encoded brackets (`%5B1%5D`) remain encoded — no URL-decoding pass.
- Non-numeric / nested bracket keys (`obj[a][0]`) survive verbatim.
- Leading/trailing whitespace is trimmed, but the indexed query is otherwise untouched.
- Non-same-origin values (`//host?...`, `https://...?...`) and CRLF-injected values fall back regardless of indices.

Truth-first correction: `resolveInternalRouteHref(value, fallback)` is exactly `normalizeInternalRouteHref(value) ?? fallback`, and `normalizeInternalRouteHref` only trims and gates on a leading `/` (not `//`) and CR/LF. There is no query parser, "extraction engine," array aggregation, or index ordering — a valid same-origin path is returned byte-for-byte. These tests pin that verbatim passthrough so any future structural parsing is caught as an intentional contract shift.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-inverted-indexes.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-inverted-indexes.test.ts
@@ -1,0 +1,87 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) when a query string supplies
+// explicit bracketed array indices OUT OF ORDER, e.g.
+//   "/api/v1?items[1]=b&items[0]=a"
+//
+// TRUTH-FIRST — CURRENT CONTRACT (verified against source):
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `e45b578a`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3983, opened 2026-06-27
test(utils): characterize formatCompleteJson on fractional float precision

abdulbeast4-creator:
## Summary
Characterizes structural output metrics within formatCompleteJson when formatting values utilizing JavaScript floating point primitives (e.g. `0.1 + 0.2`, `1 / 3`).

Literal layout boundaries pinned by this test (generic, non-currency, non-percentage keys):
- `0.1 + 0.2` renders as `0.3` — no IEEE-754 artifact (`0.30000000000000004` never appears).
- `1 / 3` renders as `0.3333` — rounded to exactly four fraction digits, never the repeating expansion.
- `2 / 3` renders as `0.6667` — rounded, not truncated.
- `0.5` renders as `0.5` — trailing zeros stripped (no `0.5000`).
- `0.12345` rounds at the 4-digit boundary to `0.1235`.
- Integer `42` renders as `42` — no decimal point (no `42.0`).

Truth-first correction: generic numbers flow through `formatNumber`, which is `new Intl.NumberFormat("en-US", { minimumFractionDigits: 0, maximumFractionDigits: 4 }).format(value)`. There is no raw floating-point passthrough and no preservation of trailing binary digits — the fraction is bounded to at most four rounded places with trailing zeros stripped. These assertions lock that layout so any future precision/format change is caught as an intentional contract shift. (Currency/percentage-classified keys use separate `toFixed(2)` paths and are out of scope here.)

## CI gate checklist
- [x] DCO Verified
- [x] Truth-First — documents real, existing rounding/layout behavior

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-floats.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-floats.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) on dense / repeating
+// floating-point values such as `0.1 + 0.2` and `1 / 3`.
+//
+// TRUTH-FIRST — CURRENT CONTRACT (verified against source):
+//
+//   Top-level scalar numbers render as `${Label}: ${formatValue(key, value)}`.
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `1157749a`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3984, opened 2026-06-27
test(navigation): characterize normalizeInternalRouteHref protocol relative slurs

abdulbeast4-creator:
## Summary
Pins down specific routing isolation rules when protocol-relative URLs (double-slash, no protocol word, e.g. `//hushh.ai/legal`) interface with normalizeInternalRouteHref.

Literal assertions pinned by this test:
- `//hushh.ai/legal` → `null` (rejected as external, not parsed as an internal sub-path).
- `//hushh.ai` and bare `//` → `null`.
- Deeper variants `//evil.example.com/a/b/c`, `...?x=1`, `...#frag` → `null`.
- `   //hushh.ai/legal   ` → `null` (rejection survives whitespace trimming).
- `///hushh.ai/legal` and `////a` → `null` (3+ leading slashes still start with `//`).
- CONTRAST: single-slash `/hushh.ai/legal` → `/hushh.ai/legal` and `/legal` → `/legal` (genuine internal paths returned verbatim).

Truth-first answer to the premise question: protocol-relative inputs are treated as EXTERNAL and rejected, never interpreted as nested internal pathways. `normalizeInternalRouteHref` guards with `if (!href.startsWith("/") || href.startsWith("//")) return null;` after trimming, plus a CR/LF reject. There is no "routing engine" doing host parsing — the single explicit `//` guard is the whole isolation rule. These assertions lock that boundary so any future relaxation of the open-redirect guard is caught.

## CI gate checklist
- [x] DCO Verified
- [x] Truth-First — documents real, existing open-redirect rejection boundary

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-protocol-relative.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-protocol-relative.test.ts
@@ -1,0 +1,61 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) on protocol-relative inputs that
+// begin with double forward slashes and omit the protocol word, e.g.
+//   "//hushh.ai/legal"
+//
+// TRUTH-FIRST — CURRENT CONTRACT (verified against source):
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `84606b30`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3986, opened 2026-06-27
test(utils): characterize formatCompleteJson array nested object circular key baselines

abdulbeast4-creator:
## Summary
Establishes truth-first validation benchmarks on formatCompleteJson when multiple sibling objects inside a layout array each reference a SINGLE shared object further up the tree (shared-reference matrix).

Literal behavior pinned by this test (generic top-level array → else branch):
- Three siblings each holding the same `shared` object → exactly 3 bullets, each rendering `[object Object]` (one per sibling, no collapsing).
- No tracking markers are ever emitted: no `[Circular]`, no `[Reference]`, and the shared object's keys (`root-shared`) do not surface.
- Header is verbatim `--- Widgets (2 items) ---`.
- Preview caps at 3 siblings; 5 items → 3 bullets + `... and 2 more` (shared refs counted as distinct slots).
- The else branch uses the FIRST non-null property: `{ name: "Alpha", ref: shared }` surfaces `• Alpha`, not the object.
- Does not throw and does not recurse into the shared object (`should-not-appear` / `deep` never appear).

Truth-first correction: `formatCompleteJson` has NO reference tracking — no WeakSet/seen-set, no cycle detection, no de-duplication, and no `[Circular]` sentinel. So there are no "specialized tracking exceptions or omissions" for shared references; each sibling is walked independently and a shared object simply stringifies to `[object Object]` once per occurrence (up to the 3-item generic preview cap). These assertions lock that repeat-without-dedup contract.

## CI gate checklist
- [x] DCO Verified
- [x] Truth-First — documents real, existing no-tracking traversal behavior

### DIFF
--- a/hushh-webapp/__tests__/utils/format-circular-keys.test.ts
+++ b/hushh-webapp/__tests__/utils/format-circular-keys.test.ts
@@ -1,0 +1,100 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) when multiple sibling objects
+// inside a layout array each hold a reference to a SINGLE shared object that
+// lives further up the tree.
+//
+// TRUTH-FIRST — CURRENT CONTRACT (verified against source):
+//
+//   formatCompleteJson has NO reference tracking whatsoever. There is no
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `30ec4bd8`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3987, opened 2026-06-27
test(navigation): characterize resolveInternalRouteHref multi-delimiter fragment splits

abdulbeast4-creator:
## Summary
Maps tracking and extraction boundaries within resolveInternalRouteHref when resolving queries embedded within multi-hash strings (e.g. `/search?q=tags#react#node`).

Literal behavior pinned by this test:
- `/search?q=tags#react#node` → returned unchanged (query `?q=tags` + both `#react` / `#node` anchors preserved as one string).
- No stripping/reordering/splitting: result is NOT reduced to `/search?q=tags` or `/search`.
- `/list?tab=open#a#b#c` → verbatim (additional `#` delimiters preserved).
- `   /search?q=tags#react#node   ` → trimmed at the edges, inner query+hash payload intact.
- `//evil.example.com/search?q=tags#react#node` → returns the fallback (protocol-relative rejected even with multi-hash payload).
- `/search?q=tags#react\n#node` (CR/LF injected) → returns the fallback.
- `null` / `""` → fallback; `/p?x=#` → verbatim.

Truth-first correction: `resolveInternalRouteHref(value, fallback)` is literally `normalizeInternalRouteHref(value) ?? fallback`, and `normalizeInternalRouteHref` only trims, rejects empty / non-`/` / `//` / CRLF, then returns the string verbatim. There is NO query parser, NO `URL`/`URLSearchParams` parsing, and NO splitting on `?` or `#` — so there is no "parsing matrix" that isolates the active query block from trailing anchor parameters. A valid internal href passes through opaque (hashes and all). These assertions lock that pass-through-verbatim contract plus the existing security rejections (protocol-relative + CRLF).

## CI gate checklist
- [x] DCO Verified
- [x] Truth-First — documents real, existing verbatim pass-through behavior

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-fragment-splits.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-fragment-splits.test.ts
@@ -1,0 +1,79 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) when an internal route value carries
+// a query string whose value payload itself contains multiple nested hash
+// symbols, e.g. "/search?q=tags#react#node".
+//
+// TRUTH-FIRST — CURRENT CONTRACT (verified against source):
+//
+//   resolveInternalRouteHref(value, fallback) is literally:
+//       normalizeInternalRouteHref(value) ?? fallback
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `e7ad6294`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3989, opened 2026-06-27
test(navigation): characterize normalizeInternalRouteHref percent-encoded control tokens

abdulbeast4-creator:
## Summary
Pins and documents current existing characterization criteria inside normalizeInternalRouteHref when escaped control sequences cross route strings.

Literal behavior pinned by this test:
- `/path%0Dvalue`, `/path%0Avalue`, `/inject%0D%0ASet-Cookie=evil`, `/p%0d%0aq`, `/a%09b%00c` → all returned **verbatim** (percent-encoded control tokens are NOT filtered).
- Raw (decoded) control chars are rejected → `null`: `/path\r...`, `/path\n...`, `/inject\r\n...`.
- Other guards still compose: `//evil%0D%0A/x` → `null` (protocol-relative); `   /ok%0Dvalue   ` → trimmed to `/ok%0Dvalue`.

Truth-first correction: `normalizeInternalRouteHref` trims, returns `null` for empty / non-`/` / `//`, then rejects only **raw** CR/LF via the regex `/[\r\n]/`, otherwise returns the string verbatim. It does NOT call `decodeURIComponent` or decode percent escapes, so the *encoded* tokens `%0D`/`%0A` are ordinary ASCII to the guard and pass through unchanged — the path guard does NOT filter percent-encoded control tokens, it leaves them verbatim. These assertions lock that raw-vs-encoded distinction plus the existing protocol-relative + trim guards.

## CI gate checklist
- [x] DCO Verified
- [x] Truth-First — documents real, existing guard behavior (raw CR/LF rejected; encoded verbatim)

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-control-tokens.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-control-tokens.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) for route strings carrying
+// percent-encoded control characters / line endings, e.g. "%0D" (CR) and
+// "%0A" (LF), versus their raw counterparts.
+//
+// TRUTH-FIRST — CURRENT CONTRACT (verified against source):
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `2844740d`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #3995, opened 2026-06-27
test(spinner): cover status role and loading aria label

M-Rajith:
Covers the accessibility contract of Spinner — renders with role="status" 
and aria-label="Loading" so screen readers announce loading state correctly.

Attach point: hushh-webapp/components/ui/spinner.tsx
Single file, single surface.


kushaltrivedi5 (approved):

Re-reviewed at head `61eb96ac`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4000, opened 2026-06-27
test(utils): characterize formatCompleteJson array nested null property compressions

abdulbeast4-creator:
## Summary
Establishes truth-first whitespace benchmarks on formatCompleteJson under dense configurations of null values.

Truth-first note: there is NO special "null compression", placeholder, or indentation variation for dense-null structures. `formatCompleteJson` simply skips `null`/`undefined` via `continue`: a fully-null top-level object → empty string; an all-null object section → a bare `--- Label ---` header (no value lines); a nested all-null object → just its `  Label:` line. Inside arrays, null *items* are NOT skipped — they stringify to the literal `null`, and all-null object items collapse to the literal `Item`. These tests pin that exact layout.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-null-compressions.test.ts
+++ b/hushh-webapp/__tests__/utils/format-null-compressions.test.ts
@@ -1,0 +1,101 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `2323cc0a`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4001, opened 2026-06-27
test(navigation): characterize resolveInternalRouteHref multi-question parameter truncations

abdulbeast4-creator:
## Summary
Maps tracking and extraction boundaries within resolveInternalRouteHref when dealing with multi-question parameter queries.

Truth-first note: `resolveInternalRouteHref(value, fallback) = normalizeInternalRouteHref(value) ?? fallback`. There is NO query parser — it does NOT split on `?`, does NOT treat only the first `?` as a delimiter, does NOT truncate at a second `?`, and does NOT decode. For a valid internal href the ENTIRE string (path + every raw `?` + params) is returned verbatim; the safety guards (must start with a single `/`, no `//`, no CR/LF) still reject protocol-relative, absolute, and newline-bearing inputs to the fallback. These tests pin that exact contract.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-query-truncations.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-query-truncations.test.ts
@@ -1,0 +1,103 @@
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `e6923eaa`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4005, opened 2026-06-27
test(navigation): characterize normalizeInternalRouteHref redundant hash slash paths

abdulbeast4-creator:
## Summary
Pins and documents current existing characterization criteria inside normalizeInternalRouteHref when slash boundaries are passed within trailing hash parameters.

Truth-first note: `normalizeInternalRouteHref` is not a URL/hash parser. It trims, rejects empty / non-`/`-leading / `//`-leading / CR-LF inputs, and otherwise returns the string verbatim. The `//` rejection is anchored to the START of the whole string (protocol-relative guard) — slashes living inside the `#` fragment (even `#//privacy///terms`) are irrelevant to the guard and are never collapsed, validated, or decoded. These tests pin both that verbatim pass-through and the still-active start-anchored guards.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-hash-slashes.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-hash-slashes.test.ts
@@ -1,0 +1,91 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `62e6f10a`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4007, opened 2026-06-27
test(utils): characterize formatCompleteJson array string value escape sequences

abdulbeast4-creator:
## Summary
Characterizes JSON structural string output metrics inside formatCompleteJson when parsing fields heavily relying on control code escape formatting sequences.

Truth-first note: string values route through `cleanMarkdown`, which strips markdown tokens (`***`, `**`, `*`, `` ` ``) and then `String.prototype.trim()`. There is no escaping, no whitespace collapsing, and no control-character stripping beyond `trim()`. JS `trim()` removes `\t \n \r \f` (and space) only at the string edges and does NOT treat `\b` (U+0008 backspace) as whitespace. Interior control characters — including `\n`, `\t`, `\r`, `\f`, `\b` — pass into the joined output verbatim; the layout (section headers, indentation, newline joins) is built around values and never sanitizes their contents. These tests pin that exact behavior.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-escape-sequences.test.ts
+++ b/hushh-webapp/__tests__/utils/format-escape-sequences.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `85fb6c39`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4013, opened 2026-06-27
test(navigation): verify isRiaRoute boundary handling

imsharukhan:
## What

Adds coverage for `isRiaRoute` in `lib/navigation/routes.ts`.

## Why

`isRiaRoute` relies on an exact-match check and a slash-delimited prefix check. The slash-boundary behavior is important because routes that merely share the `/ria` prefix should not be classified as RIA routes.

The tests verify:

* exact RIA route classification
* nested RIA route classification
* slash-boundary handling for prefix lookalikes
* non-RIA route classification

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/navigation/is-ria-route.test.ts

## Notes

The tests focus on the public route-classification contract and specifically verify the slash-boundary behavior that distinguishes valid RIA routes from prefix lookalikes. No shared test setup or production code is modified.

### DIFF
--- a/hushh-webapp/__tests__/navigation/is-ria-route.test.ts
+++ b/hushh-webapp/__tests__/navigation/is-ria-route.test.ts
@@ -1,0 +1,62 @@
+import { describe, expect, it } from "vitest";
+
+import { isRiaRoute } from "@/lib/navigation/routes";
+
+/**
+ * Characterization tests for isRiaRoute.
+ *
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `ea573c3c`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4014, opened 2026-06-27
test(navigation): verify isOneOnboardingRoute boundaries

imsharukhan:
## What

Adds coverage for `isOneOnboardingRoute` in `lib/navigation/routes.ts`.

## Why

The canonical onboarding-route classifier is also exposed through an alias, but important namespace and prefix-boundary behavior is not currently covered directly.

The tests verify:

* canonical onboarding route classification
* legacy onboarding route classification
* RIA onboarding namespace separation
* non-onboarding prefix boundary behavior

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/navigation/is-one-onboarding-route.test.ts

## Notes

The tests focus on route-classification boundaries and the canonical export without modifying shared test setup or production code. They intentionally verify namespace and prefix distinctions that are easy to regress during route refactors.

### DIFF
--- a/hushh-webapp/__tests__/navigation/is-one-onboarding-route.test.ts
+++ b/hushh-webapp/__tests__/navigation/is-one-onboarding-route.test.ts
@@ -1,0 +1,76 @@
+import { describe, expect, it } from "vitest";
+
+import { isOneOnboardingRoute } from "@/lib/navigation/routes";
+
+/**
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4019, opened 2026-06-27
test(kbd): verify className propagation

imsharukhan:
## What

Adds coverage for custom `className` propagation in the Kbd component.

## Why

The component merges consumer-provided class names with its default styling, but this behavior is not currently verified by the test suite.

The test verifies:

* custom `className` values are applied to the rendered kbd element

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/ui/kbd.test.tsx

## Notes

The test verifies the public rendering contract directly through the rendered DOM. It does not use mocks, snapshots, browser APIs, interactions, or shared test setup changes.

### DIFF
--- a/hushh-webapp/__tests__/ui/kbd.test.tsx
+++ b/hushh-webapp/__tests__/ui/kbd.test.tsx
@@ -18,4 +18,17 @@ describe("Kbd", () => {
     expect(el?.tagName).toBe("KBD");
   });
 
+  it("merges a custom className onto the kbd element", () => {
+    const { container } = render(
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `931e58cd`: test-only change on `hushh-webapp/__tests__/ui/kbd.test.tsx` that imports and exercises the real production module, path-clean, MERGEABLE, base integration/pr-train. Enqueued to the merge queue (CI is the backstop).

PR #4024, opened 2026-06-27
test(table): verify className target element

imsharukhan:
## What

Adds coverage for className propagation in the Table component.

## Why

The Table component applies consumer-provided class names to the inner table element rather than the outer container wrapper. This behavior is not currently verified by the test suite.

The test verifies:

* custom class names are applied to the rendered table element
* the outer table container does not receive the custom class
* the rendered target remains a table element

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/ui/table.test.tsx

## Notes

The test verifies the public rendering contract directly through the rendered DOM and ensures the custom class is applied to the correct element. No mocks, snapshots, browser APIs, interactions, asynchronous behavior, or shared test setup changes are required.

### DIFF
--- a/hushh-webapp/__tests__/ui/table.test.tsx
+++ b/hushh-webapp/__tests__/ui/table.test.tsx
@@ -76,4 +76,26 @@ describe("Table", () => {
     );
   });
 
+  it("merges custom className onto the inner table element, not the container", () => {
+    const { container } = render(
+      <Table className="custom-table-class">
+        <tbody>
+          <tr>
+            <td>Holding</td>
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `1f3efadd`: test-only change on `hushh-webapp/__tests__/ui/table.test.tsx` that imports and exercises the real production module, path-clean, MERGEABLE, base integration/pr-train. Enqueued to the merge queue (CI is the backstop).

PR #4025, opened 2026-06-27
test(collapsible): verify root className propagation

imsharukhan:
## What

Adds coverage for className propagation on the Collapsible root element.

## Why

The Collapsible component forwards props to its root primitive, allowing consumers to provide custom class names. This behavior is not currently verified by the test suite.

The test verifies:

* custom class names are applied to the rendered root element

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/ui/collapsible.test.tsx

## Notes

The test verifies the public rendering contract through the existing root selector already used elsewhere in the suite. No mocks, snapshots, browser APIs, interactions, asynchronous behavior, or shared test setup changes are required.

### DIFF
--- a/hushh-webapp/__tests__/ui/collapsible.test.tsx
+++ b/hushh-webapp/__tests__/ui/collapsible.test.tsx
@@ -59,4 +59,19 @@ describe("Collapsible", () => {
         ?.classList.contains("custom-content-class"),
     ).toBe(true);
   });
+
+  it("passes className through to the root element", () => {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested

**Proof Needed:** `npm run test -- hushh-webapp/__tests__/ui/collapsible.test.tsx` green against the real import, and the CI Status Gate green at the new head.

PR #4027, opened 2026-06-27
test(button): verify default button type

imsharukhan:
## What

Adds coverage for the Button component's default type attribute.

## Why

Button explicitly defaults to `type="button"` when rendered as a native button element. This prevents accidental form submission when the component is used inside forms, but the behavior is not currently verified by the test suite.

The test verifies:

* Button renders with `type="button"` by default

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/ui/button.test.tsx

## Notes

The test verifies a public safety contract through a simple DOM attribute assertion. No mocks, snapshots, browser APIs, interactions, asynchronous behavior, or shared test setup changes are required.

### DIFF
--- a/hushh-webapp/__tests__/ui/button.test.tsx
+++ b/hushh-webapp/__tests__/ui/button.test.tsx
@@ -12,4 +12,13 @@ describe("Button", () => {
     expect(button).not.toBeNull();
     expect(button?.hasAttribute("disabled")).toBe(true);
   });
+
+  it("defaults to type='button' to prevent accidental form submission", () => {
+    const { container } = render(<Button>Save</Button>);
+
+    const button = container.querySelector("button");
+
+    expect(button?.getAttribute("type")).toBe("button");
+  });
+
 });


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Really appreciate this — you're steadily building out real regression coverage for the UI primitives, and the case you added here is a genuinely useful one. The only issue is a merge-ordering collision, not the test itself.

**Blocker:** This PR edits `hushh-webapp/__tests__/ui/button.test.tsx`, the same file as a sibling PR (#4018) that is already approved and in the merge queue this cycle. Two PRs can't land the same file in one queue pass, so this one needs to come in right after the sibling merges.

**Path To Merge:** Once #4018 merges, rebase this branch on the latest `integration/pr-train` (`git fetch origin && git rebase origin/integration/pr-train`), resolve the trivial overlap in `hushh-webapp/__tests__/ui/button.test.tsx` so your case sits alongside the sibling's, and push. It will then be clean to queue.

**Proof Needed:** `npm run test -- hushh-webapp/__tests__/ui/button.test.tsx` green after the rebase, and the CI Status Gate green at the new head.

PR #4030, opened 2026-06-28
fix(agent): move welcome suggestion resolution out of render

parthmawai:
## Summary

Fixes a React lifecycle warning caused by resolving welcome suggestions during the render phase of `AgentChatWorkspace`.

The change moves welcome suggestion resolution from `useMemo()` into a React effect so that cache population occurs after render, eliminating render-phase side effects while preserving the existing `CacheService` behavior.

---

## Problem

`AgentChatWorkspace` previously resolved welcome suggestions inside a `useMemo()`:

```tsx
const welcomeSuggestions = useMemo(() =>
  resolveAgentWelcomeSuggestions(...),
  [...]
);
```

`resolveAgentWelcomeSuggestions()` is not render-pure. On a cache miss it populates `CacheService`, and `CacheService.set()` synchronously emits updates to subscribers.

One subscriber (`KaiCommandBarGlobal`) updates React state in response to those cache events, which resulted in the runtime warning:

```
Cannot update a component (`KaiCommandBarGlobal`) while rendering a different component (`AgentChatWorkspace`)
```

---

## Solution

Move welcome suggestion resolution into a `useEffect()` and store the resolved suggestions in component state.

This ensures that:

* render remains free of side effects,
* cache population still occurs,
* existing cache semantics are preserved,
* subscriber notifications occur after React completes the current render.

The implementation intentionally leaves `resolveAgentWelcomeSuggestions()` and the underlying cache implementation unchanged.

---

## Validation

Manual verification:

* Reproduced the React warning before the change.
* Verified the warning no longer appears after the change.
* Welcome suggestions continue to render correctly.
* Suggestions continue to update when opening a new chat and navigating between routes.
* No additional console warnings or errors observed.

Project checks:

* Typecheck passes.
* Existing `agent-welcome-suggestions` tests continue to pass.

---

### Why not keep `useMemo`?

`useMemo` is appropriate only for render-pure computations.

Because `resolveAgentWelcomeSuggestions()` may populate `CacheService` on a cache miss, invoking it during render can synchronously notify subscribers, leading to render-phase state updates.

Moving the invocation into a React effect keeps render pure while preserving the existing cache behavior.

### Why not remove the cache?

The existing implementation and tests indicate that caching is intentional. This change preserves the cache contract rather than altering application behavior.

### Why not defer with `setTimeout`?

A deferred timer would avoid the warning, but it introduces scheduling-based behavior. Using React's lifecycle (`useEffect`) keeps the side effect aligned with React's execution model instead of relying on timing.

### DIFF
--- a/hushh-webapp/components/agent/agent-chat-workspace.tsx
+++ b/hushh-webapp/components/agent/agent-chat-workspace.tsx
@@ -2719,16 +2719,18 @@ export function AgentChatWorkspace({
     // eslint-disable-next-line react-hooks/exhaustive-deps
     [activePersona, conversationId, freshOpenKey, hasStartedConversation, pathname, user?.uid],
   );
-  const welcomeSuggestions = useMemo(
-    () =>
+  const [welcomeSuggestions, setWelcomeSuggestions] = useState<AgentWelcomeSuggestion[]>([]);
+
+  useEffect(() => {
+    setWelcomeSuggestions(
       resolveAgentWelcomeSuggestions({
         userId: user?.uid,
         pathname,
         persona: activePersona,
         randomSeed: welcomeSuggestionSeed,
-      }),
-    [activePersona, pathname, user?.uid, welcomeSuggestionSeed],
-  );
+      })
+    );
+  }, [activePersona, pathname, user?.uid, welcomeSuggestionSeed]);
   const latestRetryableAssistantId =
     [...visibleMessages]
       .reverse()


parthmawai:
Quick update on this PR:

While validating the fix, I continued testing the Kai chat flow locally and uncovered another issue that appears to be unrelated to the original render-phase bug addressed in this PR.

Current findings:

* The original React warning (`Cannot update a component while rendering a different component`) is resolved by moving welcome suggestion resolution out of the render phase.
* During further testing (after resetting the local vault state), I consistently hit a separate React error: `Maximum update depth exceeded`.
* The current stack traces point to `lib/navigation/kai-bottom-chrome-visibility.ts`, specifically through the `useSyncExternalStore` subscription path (`emit()` → `animateProgress()` → `forceStoreRerender`), suggesting a separate feedback loop rather than a regression from this change.
* I've narrowed the investigation down to the bottom chrome visibility subscribers and am isolating the component responsible for the repeated store updates.

Since this appears to be a distinct issue with a different root cause, I'd prefer to keep this PR focused on the render-phase fix and submit a follow-up PR for the bottom chrome/update-depth issue once I've identified and verified the root cause.

I'll post another update once I have a confirmed fix.

kushaltrivedi5 (approved):

Re-reviewed at head `ee92bb6a`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4031, opened 2026-06-28
test(utils): characterize formatCompleteJson on plain object vs custom class instances

abdulbeast4-creator:
## Summary
Adds an isolated characterization test for formatCompleteJson to document and verify serialization output paths when encountering instantiated custom class objects.

Literal code assertion being pinned: `formatCompleteJson` traverses values with `Object.entries(...)` at every level (top-level sections, object sections, one nested level). `Object.entries` returns only own enumerable string-keyed properties. For a JS class instance:
- field declarations (`id = 1`) and constructor assignments (`this.x = 5`) become own enumerable instance properties — identical to a plain literal.
- prototype methods and getters are non-enumerable, so they are never read.
- inherited base-class field declarations are initialized as own enumerable properties on the subclass instance, so they are included.

The tests assert that the formatted string for a class instance is byte-for-byte equal to the output for a plain literal carrying the same own enumerable fields, that prototype methods/getters are excluded, that inherited class fields are included, and that field-specific currency formatting applies identically. There is no `instanceof`/constructor/class-name branch in the utility; this PR documents that real existing behavior only.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-classes.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-classes.test.ts
@@ -1,0 +1,134 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on the EXACT contract when a
+// top-level section value is an instance of a user-defined class versus a plain
+// object literal with the same own fields.
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

kushaltrivedi5 (approved):

Re-reviewed at head `7ec0e0d6`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4033, opened 2026-06-28
test(navigation): characterize resolveInternalRouteHref verbatim %00 preservation (no decode, no truncation)

abdulbeast4-creator:
## Summary
Characterizes `resolveInternalRouteHref` (hushh-webapp/lib/navigation/routes.ts) for internal hrefs whose query values carry escaped null-byte tokens (`%00`) and raw NUL.

The blocker was correct: the prior title claimed "parameter **array** null byte **truncations**," asserting array handling and truncation the surface does **not** have. The verified truth is the opposite — `%00` is preserved verbatim and **never truncated**, and there is no array-parameter handling. The title now states the real surface: verbatim `%00` preservation with no truncation.

### Truth-first note (verified against source)
`resolveInternalRouteHref(value, fallback)` is `normalizeInternalRouteHref(value) ?? fallback`. `normalizeInternalRouteHref` only trims, rejects empty / non-`/`-rooted / `//`-prefixed / CR-LF-containing input, and otherwise returns the href verbatim. There is **no** `decodeURIComponent`, **no** query parsing, and **no** null-byte handling. A `%00` token is therefore preserved as three literal characters — never decoded and never truncated — so text after the "null character boundary" is kept intact. A raw NUL (`\x00`) is likewise not in the CR/LF reject set, so it does not trigger the fallback (in contrast to CR/LF, which does). These tests pin that exact behavior and confirm the fallback path still applies to protocol-relative and non-rooted hrefs that carry `%00`.

## CI gate checklist
- [x] DCO Verified
- [x] Claim matches surface — title corrected (no array handling, no truncation)

### DIFF
--- a/hushh-webapp/__tests__/navigation/resolve-null-bytes.test.ts
+++ b/hushh-webapp/__tests__/navigation/resolve-null-bytes.test.ts
@@ -1,0 +1,92 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for resolveInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on percent-encoded null bytes
+// (`%00`) appearing inside query parameter values.
+//
+// TRUTH-FIRST (verified against the source):
+//   resolveInternalRouteHref(value, fallback) === normalizeInternalRouteHref(value) ?? fallback
+//   normalizeInternalRouteHref only:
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4034, opened 2026-06-28
test(utils): characterize formatCompleteJson on JavaScript global context fields

abdulbeast4-creator:
## Summary
Characterizes structural output metrics within formatCompleteJson when parsing fields referencing global runtime environments.

Truth-first note (verified against `hushh-webapp/lib/utils/json-to-human.ts` and confirmed by the live run): `formatCompleteJson` does NOT use `JSON.stringify`. It iterates `Object.entries(json)` at the top level and dispatches on `typeof`, descending at most one nested level. Three baselines are pinned:
1. A top-level section whose value is a function (`typeof === "function"`) matches no branch and is silently dropped (produces no output); the SAME function held as an object FIELD is rendered via `String(fn)` (label `OnClick` preserves the inner capital).
2. A global root (`globalThis`) reached as a bounded LEAF (nested under a normal object) is coerced via `String()` to `"[object ...]"` and the walk TERMINATES — no deep/infinite recursion.
3. A global root placed DIRECTLY as a top-level section value is enumerated one level deep and hits a member that cannot be coerced, so the formatter THROWS `TypeError: Cannot convert object to primitive value`. Documented guidance: never feed raw execution-scope roots in as section values.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-globals.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-globals.test.ts
@@ -1,0 +1,104 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on objects that hold direct
+// references to native JavaScript global / execution-scope roots (globalThis,
+// window) and to callable scope members (functions).
+//
+// TRUTH-FIRST (verified against the source AND the live run):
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

kushaltrivedi5 (approved):

Re-reviewed at head `a8280b91`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4035, opened 2026-06-28
test(utils): characterize formatCompleteJson array nested object string constructor overrides

abdulbeast4-creator:
## Summary
Establishes truth-first validation benchmarks on formatCompleteJson under customized object property tags.

Truth-first finding (verified against `hushh-webapp/lib/utils/json-to-human.ts` and confirmed by the live run): `{ constructor: "..." }` is an own enumerable property, so `Object.entries`/`Object.values` include it and the string VALUE always passes through correctly — there is no prototype-pollution guard and no special omission for `constructor`.

The surprising, pinned behavior is in the LABEL path, not the value path: the field-label lookup resolves the key against a plain object map (`fieldMappings[key]`). For `key === "constructor"` that lookup does NOT miss — it hits the INHERITED `Object.prototype.constructor` (the native `Object` function), which is then stringified into the label, producing `function Object() { [native code] }: <value>` rather than a title-cased `Constructor: <value>`. The generic ARRAY branch renders items via `Object.values` with no key label, so it is immune to the leak and shows the plain string.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-constructor-overrides.test.ts
+++ b/hushh-webapp/__tests__/utils/format-constructor-overrides.test.ts
@@ -1,0 +1,103 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on payload dictionaries
+// whose standard `constructor` reference tag has been overwritten by a plain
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

kushaltrivedi5 (approved):

Re-reviewed at head `4deb32d1`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4037, opened 2026-06-28
test(utils): characterize formatCompleteJson on JavaScript generator function attributes

abdulbeast4-creator:
## Summary
Establishes serialization structure rules within formatCompleteJson when encountering native JavaScript Generator definitions across discrete data planes.

Truth-first finding (verified against `hushh-webapp/lib/utils/json-to-human.ts`): there is **no `typeof === "function"` branch** anywhere in `formatCompleteJson`. The premise that it returns "blank properties or standard function omissions" is only **partly** correct — the behavior is plane-dependent:

- **Top-level value:** the loop only matches `number|string`, `Array`, and `typeof === "object"`. A function's `typeof` is `"function"` (not `"object"`), so no branch matches and the entry is **silently dropped** (no line emitted). This is the omission case.
- **Inside an object section:** the per-field `else` calls `formatValue`, which falls through to `String(value)` — rendering the generator's **source text** (`function* demo() { yield 1; }`), not a blank.
- **Generic array of objects:** the first non-null value is printed via `String(firstValue)` — again the **source text**.

These 5 tests pin all three planes exactly.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-generator-functions.test.ts
+++ b/hushh-webapp/__tests__/utils/format-generator-functions.test.ts
@@ -1,0 +1,86 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on objects that carry an
+// un-invoked JavaScript Generator function as a property value, e.g.
+//   function* demo() { yield 1; }
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

kushaltrivedi5 (approved):

Re-reviewed at head `c2d86a54`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4038, opened 2026-06-28
test(navigation): characterize normalizeInternalRouteHref multi-byte UTF-8 directory segments

abdulbeast4-creator:
## Summary
Pins and documents current existing characterization criteria inside normalizeInternalRouteHref when international localized text structures route fragments.

Truth-first finding (verified against `hushh-webapp/lib/navigation/routes.ts`): `normalizeInternalRouteHref` does **not** "normalize" text content in any locale-aware sense. Its full body is: `.trim()` the outer edges, reject empty / non-`/` / protocol-relative (`//`) / CRLF inputs (returning `null`), and otherwise return the string **verbatim**. There is **no** Unicode normalization (no NFC/NFD), no percent-encoding, no transliteration, and no per-segment validation. Multi-byte UTF-8 directory names satisfy the guards, so they pass through byte-for-byte identical.

These 11 tests pin: verbatim pass-through for Korean, Japanese, Arabic (RTL), Cyrillic, emoji/astral-plane, and accented Latin segments; that a decomposed sequence is **not** folded into its composed form; and that the existing guards (edge-trim, `//` rejection, non-`/` rejection, newline rejection) still apply when UTF-8 text is present.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-utf8-directories.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-utf8-directories.test.ts
@@ -1,0 +1,81 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on internal route paths whose
+// directory segments are composed of multi-byte UTF-8 / international localized
+// text, e.g. '/legal/데이터/보안'.
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

kushaltrivedi5 (approved):

Re-reviewed at head `7f9d47a0`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4072, opened 2026-06-28
test(utils): characterize formatCompleteJson on native Set payload instances

abdulbeast4-creator:
## Summary
Adds an isolated characterization test for formatCompleteJson to document and verify serialization output paths when encountering native Set objects.

Literal assertion being pinned (verified against `hushh-webapp/lib/utils/json-to-human.ts`): formatCompleteJson has **no** Set special-casing. A `Set` is `typeof "object"` and not `Array.isArray`, so it enters the object branch and is iterated with `Object.entries(...)`. Because Set members are stored internally (not as own enumerable string-keyed properties) and `.size` is a non-enumerable prototype getter, `Object.entries(new Set([...]))` is always `[]`. Therefore:

- A top-level Set section emits only the `--- <Label> ---` header; **all members are silently dropped** (no bullet lines, no `item(s)` count — that count path is reserved for `Array.isArray` values).
- A nested Set field emits only the `  <Field Label>:` line; members and size are dropped.
- It never throws; an empty Set renders identically to a populated one; sibling primitive fields are unaffected.

These 7 tests pin exactly that observed, existing behavior — no un-implemented generalization is asserted.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-sets.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-sets.test.ts
@@ -1,0 +1,106 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on native JavaScript `Set`
+// instances embedded inside payload objects.
+//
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `790b0f34`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4077, opened 2026-06-28
test(navigation): characterize normalizeInternalRouteHref encoded slash guard intersections

abdulbeast4-creator:
## Summary
Adds a distinct, non-overlapping characterization for `normalizeInternalRouteHref` covering the **intersections** of percent-encoded slashes (`%2F`/`%2f`) with the function's other guard clauses.

A sibling file `normalize-encoded-slashes.test.ts` already pins the basic `%2F` directory-fragment behavior, so this PR intentionally uses a new filename and a different angle to avoid duplicating the existing contract: leading `//` rejection vs. later `%2F`, a single `/` followed by literal `%2F` (accepted because escapes are not decoded), raw CR/LF rejection adjacent to encoded slashes, opaque encoded-newline pass-through (`%0A`/`%0D`), and whitespace-trim ordering.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-encoded-slash-segments.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-encoded-slash-segments.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `3d0fc4b9`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4090, opened 2026-06-28
test(switch): cover data-slot and size data attribute contract

M-Rajith:
Covers three data-attribute contracts on Switch:
- data-slot="switch" is always present
- data-size="default" when no size specified
- data-size="sm" when sm size passed

Attach point: hushh-webapp/components/ui/switch.tsx
Single file, single surface.


kushaltrivedi5 (approved):

Re-reviewed at head `b6e78cf1`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4094, opened 2026-06-28
chore(ui): silence debate stream view warnings in production

Roopmann30:
### Description

Silences internal development logs inside `components/kai/debate-stream-view.tsx` by wrapping several `console.warn` statements in a `NODE_ENV !== "production"` check. This prevents network fallback traces (e.g., active run cancellation failures, Kai profile context load errors, and financial PKM domain hydration errors) from leaking into the production client console, while preserving the application's intended error-handling and resilient fallback states.


kushaltrivedi5 (changes requested):
## Changes Requested

PR #4098, opened 2026-06-28
chore(ui): silence gmail receipt summary build error stack trace in production

Roopmann30:
### Description

Wraps a raw `console.error` inside the `ProfileReceiptsPage` summary generation handler with a production environment check. This prevents exposing internal API exceptions and stack traces to end-users if a shopping summary build fails, while preserving the intended UI error toast.

### DIFF
--- a/hushh-webapp/components/gmail/gmail-receipts-page.tsx
+++ b/hushh-webapp/components/gmail/gmail-receipts-page.tsx
@@ -839,10 +839,12 @@ export default function ProfileReceiptsPage() {
         setReceiptMemoryArtifact(artifact);
         setReceiptMemoryMessage("Your shopping summary is ready to review.");
       } catch (error) {
-        console.error(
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Appreciated work — this is a real contribution to a sensitive part of the platform. Because it touches the **Gmail data-access scope**, it cannot be auto-merged on the engine's diff-safety gate alone; it's routed to a maintainer security review so the trust boundary is verified by a human before it ships.

**Blocker:** The change lands on the Gmail data-access scope. Per the SOP, any diff on this family requires explicit maintainer/security sign-off (consent enforcement, scope shape, caller-credential binding, or regulated-advice boundary) that the automated lane cannot grant.

**Proof Needed:** A test exercising the Gmail data-access scope behavior at the changed path green, plus the CI Status Gate green at the current head.

PR #4101, opened 2026-06-28
test(navigation): characterize normalizeInternalRouteHref multi-stage percent decoded hash targets

abdulbeast4-creator:
## Summary

Adds an isolated characterization test for `normalizeInternalRouteHref` (`hushh-webapp/lib/navigation/routes.ts`) documenting its exact contract when an internal href carries nested, percent-encoded hash symbols (e.g. `/legal%2523privacy-policy`, where `%25` encodes `%`, so `%2523` is the encoded form of the literal text `%23`, which itself encodes `#`). Test-only; no production change.

```ts
if (!href) return null;
return href;
```

It performs **no** `decodeURIComponent`, **no** percent-decoding of any stage, and **no** `#`/anchor splitting. For any href that passes the three checks (non-empty, single leading slash, no CR/LF) it is an **identity** function. The test pins exactly that boundary:

- `"/legal%2523privacy-policy"` → returned byte-for-byte as `"/legal%2523privacy-policy"` (never collapsed to `%23`, never to an active `#` anchor)
- output still contains `%2523`, never `%23privacy`, never `#`
- single-stage `"/legal%23privacy-policy"` is also returned verbatim with no anchor split
- multiple layered tokens `"/a%2523b/c%2523d"` preserved across segments
- a real literal `#` alongside `%2523` (`"/legal%2523privacy#section"`) is kept verbatim — no normalization
- leading/trailing whitespace is trimmed while the inner `%2523` stays intact
- the deny rules still hold even with `%2523` present: protocol-relative (`//evil.example/...`) → `null`, non-rooted (`legal%2523...`) → `null`, embedded newline → `null`

The boundary in one line: `normalizeInternalRouteHref` treats a layered escaped hash as opaque path text — it neither decodes nor splits it; the only thing that ever happens is trimming and an allow/deny verdict.

## Truth-first scope note

The original task referenced `hushh-research/__tests__/navigation/...` and `npm test -- hushh-research/...`. There is no top-level `__tests__` in this repo; vitest only includes `__tests__/**` under `hushh-webapp`, and the module alias is `@/lib/navigation/routes`. The file therefore lives at `hushh-webapp/__tests__/navigation/normalize-decoded-hashes.test.ts` and imports via the `@/` alias.

## Verification

- `npm test -- __tests__/navigation/normalize-decoded-hashes.test.ts` → 10/10 pass

## CI gate checklist

- [x] Truth-First — documents the real, existing behavioral boundary (no decodeURIComponent, no anchor split; identity-on-allow, null-on-deny)

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-decoded-hashes.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-decoded-hashes.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+/**
+ * Characterization tests for normalizeInternalRouteHref
+ * (hushh-webapp/lib/navigation/routes.ts) when the href contains nested,
+ * percent-encoded hash symbols (e.g. `/legal%2523privacy-policy`, where `%25`
+ * is an encoded `%`, so `%2523` is the encoded form of the literal text
+ * `%23`, which is itself the encoded form of `#`).
+ *
+ * TRUTH-FIRST (verified against the source):
+ *   The guard is a pure allow/deny gate on the trimmed string:
+ *     const href = String(value ?? "").trim();
+ *     if (!href) return null;
+ *     if (!href.startsWith("/") || href.startsWith("//")) return null;
+ *     if (/[\r\n]/.test(href)) return null;
+ *     return href;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `26a508ec`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4102, opened 2026-06-28
test(utils): characterize formatCompleteJson on custom object Symbol.toStringTag overrides

abdulbeast4-creator:
## Summary

Adds an isolated characterization test for `formatCompleteJson` (`hushh-webapp/lib/utils/json-to-human.ts`) documenting its exact contract when a payload object carries an explicit `Symbol.toStringTag` modifier (e.g. `{ [Symbol.toStringTag]: "HushhSecurePayload" }`). Test-only; no production change.

`formatCompleteJson` walks values with `Object.entries(...)` and dispatches on `typeof`. Two facts drive the behavior:

1. `Object.entries(obj)` returns **only string-keyed enumerable own properties**. A `Symbol.toStringTag` entry is keyed by a Symbol, so it is excluded from every `Object.entries` pass at every depth.
2. The formatter **iterates** objects; it never calls `String(obj)` on a section/nested object, so the default-tag artifact `[object HushhSecurePayload]` is never produced either.

Net: the `Symbol.toStringTag` override is **structurally invisible** and has **zero** effect on the produced layout. The test pins exactly that:

- a section object whose only own key is the symbol tag → `"\n--- Payload ---"` (bare header, no field lines)
- the custom tag string `"HushhSecurePayload"` never appears in output
- the `[object` default-tag artifact never appears
- string-keyed siblings render normally while the symbol tag is skipped (`"\n--- Payload ---\n  Status: active"`)
- known field-label mapping still applies to string keys around the tag (`institution_name` → `Institution`, `account_type` → `Account Type`)
- a `Symbol.toStringTag` on a nested object is also ignored (`"\n--- Payload ---\n  Meta:\n    • Label: x"`)
- output is byte-identical to the same object with the symbol tag removed
- the formatter never throws on a tagged payload

## Truth-first scope note

The original task referenced `hushh-research/__tests__/utils/...` and `npm test -- hushh-research/...`. There is no top-level `__tests__` in this repo; vitest only includes `__tests__/**` under `hushh-webapp`, and the module alias is `@/lib/utils/json-to-human`. The file therefore lives at `hushh-webapp/__tests__/utils/format-json-string-tags.test.ts` and imports via the `@/` alias.

## Verification

- `npm test -- __tests__/utils/format-json-string-tags.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — documents the real, existing behavioral boundary (`Object.entries` skips symbol keys; objects iterated not stringified; tag is layout-invisible)

### DIFF
--- a/hushh-webapp/__tests__/utils/format-json-string-tags.test.ts
+++ b/hushh-webapp/__tests__/utils/format-json-string-tags.test.ts
@@ -1,0 +1,120 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+/**
+ * Characterization tests for formatCompleteJson
+ * (hushh-webapp/lib/utils/json-to-human.ts) when payload objects carry an
+ * explicit `Symbol.toStringTag` modifier
+ * (e.g. `{ [Symbol.toStringTag]: "HushhSecurePayload" }`).
+ *
+ * TRUTH-FIRST (verified against the source):
+ *   formatCompleteJson walks values with `Object.entries(...)` and dispatches on
+ *   `typeof`. It NEVER calls `String(obj)` on a section/nested object, and
+ *   `Object.entries` returns ONLY string-keyed enumerable properties. A
+ *   `Symbol.toStringTag` entry is keyed by a Symbol, so it is structurally
+ *   excluded from every `Object.entries` pass. Consequences pinned below:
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `e8d750be`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4103, opened 2026-06-28
test(navigation): characterize normalizeInternalRouteHref inline authentication parameter stripping

abdulbeast4-creator:
## Summary

Adds an isolated characterization test for `normalizeInternalRouteHref` (`hushh-webapp/lib/navigation/routes.ts`) documenting its exact contract when an internal href embeds inline URL-credential blocks (userinfo) such as `user:pass@legal/privacy`. Test-only; no production change.

```ts
if (!href) return null;
return href;
```

It does **not** parse URL authority, does **not** recognize `user:pass@` userinfo, and does **not** strip credential blocks. There is no `@`, `:`, or userinfo handling anywhere. The test pins exactly that boundary:

- a rooted href with an inline `user:pass@` block is returned **byte-for-byte** (`/user:pass@legal/privacy`) — credentials retained, not filtered
- the credential segment is not stripped (`user:pass@` and `legal/privacy` both remain)
- username-only userinfo (`/admin@dashboard/settings`) is retained verbatim
- an empty-credential `@` prefix (`/@legal/privacy`) is retained verbatim
- multiple `@`/`:` credential-looking tokens (`/a:b@c:d@legal/privacy`) are retained verbatim
- leading/trailing whitespace is trimmed while the inner credential block stays intact
- the deny rules are unchanged by credentials: a non-rooted `user:pass@...` → `null` (no leading `/`); a protocol-relative `//user:pass@host/...` → `null` (leading `//`); an embedded newline → `null`; whitespace-only input → `null`

The boundary in one line: `normalizeInternalRouteHref` neither flags nor strips inline credentials — it only enforces trim + single-leading-slash + no-CRLF, treating any credential block as opaque path text.

## Truth-first scope note

The original task referenced `hushh-research/__tests__/navigation/...` and `npm test -- hushh-research/...`. There is no top-level `__tests__` in this repo; vitest only includes `__tests__/**` under `hushh-webapp`, and the module alias is `@/lib/navigation/routes`. The file therefore lives at `hushh-webapp/__tests__/navigation/normalize-auth-stripping.test.ts` and imports via the `@/` alias. (The task example `'user:pass'legal/privacy''` is also not a real repo file — no such path exists; it is interpreted as the conceptual `user:pass@legal/privacy` userinfo case.)

## Verification

- `npm test -- __tests__/navigation/normalize-auth-stripping.test.ts` → 10/10 pass

## CI gate checklist

- [x] Truth-First — documents the real, existing behavioral boundary (no userinfo parsing/stripping; identity-on-allow, null-on-deny)

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-auth-stripping.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-auth-stripping.test.ts
@@ -1,0 +1,82 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+/**
+ * Characterization tests for normalizeInternalRouteHref
+ * (hushh-webapp/lib/navigation/routes.ts) when the href embeds inline
+ * URL-credential blocks (userinfo) such as `user:pass@legal/privacy`.
+ *
+ * TRUTH-FIRST (verified against the source):
+ *   The guard is a pure allow/deny gate over the trimmed string:
+ *     const href = String(value ?? "").trim();
+ *     if (!href) return null;
+ *     if (!href.startsWith("/") || href.startsWith("//")) return null;
+ *     if (/[\r\n]/.test(href)) return null;
+ *     return href;
+ *   It does NOT parse URL authority, does NOT recognize `user:pass@` userinfo,
+ *   and does NOT strip credential blocks. There is no `@`, `:`, or userinfo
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `7cd808ac`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4105, opened 2026-06-28
test(navigation): characterize normalizeInternalRouteHref multi-token question mark query clashes

abdulbeast4-creator:
## Summary

Adds an isolated characterization test for `normalizeInternalRouteHref` (`hushh-webapp/lib/navigation/routes.ts`) documenting its exact contract when a query string is positioned **before** directory-looking text (e.g. `/search?q=open/legal/privacy`). Test-only; no production change.

```ts
if (!href) return null;
return href;
```

It does **not** parse query strings, does **not** split on `?`, does **not** segment on `/`, and does **not** reorder/escape anything. A `?` and any `/` after it are ordinary characters inside the opaque path string. The 8 tests pin exactly that boundary:

- a rooted query-before-directories href is returned **byte-for-byte** (`/search?q=open/legal/privacy`) — no query/segment separation
- the post-`?` slashes are **not** broken into distinct segment markers; the whole tail is one literal string (`/search?q=a/b/c/d`)
- multiple `?` and `/` tokens are retained verbatim (`/search?q=open/legal?ref=x/y/privacy`)
- a query containing both an encoded `%2F` and a raw `/` is preserved unchanged
- leading/trailing whitespace is trimmed while the inner query/path stays intact
- deny rules are unaffected by query content: no leading slash → `null`; protocol-relative `//host/search?...` → `null`; embedded newline in the query tail → `null`

The boundary in one line: `normalizeInternalRouteHref` treats the query and any trailing directory-looking text as inert opaque path bytes — it neither parses nor segments them, applying only trim + single-leading-slash + no-`//` + no-CRLF gating.

## Truth-first scope note

The original task referenced `hushh-research/__tests__/navigation/...` and `npm test -- hushh-research/...`. There is no top-level `__tests__` in this repo; vitest only includes `__tests__/**` under `hushh-webapp`, and the module alias is `@/lib/navigation/routes`. The file therefore lives at `hushh-webapp/__tests__/navigation/normalize-query-clashes.test.ts` and imports via the `@/` alias.

## Verification

- `npm test -- __tests__/navigation/normalize-query-clashes.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — documents the real, existing behavioral boundary (no query parsing/segmentation; identity-on-allow, null-on-deny)

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-query-clashes.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-query-clashes.test.ts
@@ -1,0 +1,78 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+/**
+ * Characterization tests for normalizeInternalRouteHref
+ * (hushh-webapp/lib/navigation/routes.ts) when a query string is positioned
+ * BEFORE what look like routing directories
+ * (e.g. /search?q=open/legal/privacy).
+ *
+ * TRUTH-FIRST (verified against the source):
+ *   normalizeInternalRouteHref is a pure allow/deny gate over the trimmed
+ *   string:
+ *     const href = String(value ?? "").trim();
+ *     if (!href) return null;
+ *     if (!href.startsWith("/") || href.startsWith("//")) return null;
+ *     if (/[\r\n]/.test(href)) return null;
+ *     return href;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4106, opened 2026-06-28
test(utils): characterize formatCompleteJson array key density carriage return buffering

abdulbeast4-creator:
## Summary

Adds an isolated characterization test for `formatCompleteJson` (`hushh-webapp/lib/utils/json-to-human.ts`) documenting its exact contract when object string values contain raw Windows-style carriage-return line endings (CR LF) injected throughout the dictionary structures. Test-only; no production change.

`formatCompleteJson` builds output from **static literal indentation prefixes** and joins every line with a literal `\n`:

- object fields: `"  " + Label + ": " + value`
- nested fields: `"    • " + Label + ": " + value`
- generic array items: `"  • " + value`
- section header: `"--- Label ---"` preceded by an empty `""` line
- final assembly: `lines.join("\n")`

There is **no** indentation-alignment computation, **no** terminal/platform detection, and **no** CR/LF normalization. The only mutation applied to string values is `cleanMarkdown(value) → … .trim()`, which strips markdown markers and **outer** whitespace (so leading/trailing CR LF are removed) but leaves **interior** CR LF untouched. The 8 tests pin exactly that boundary:

- structural lines are joined with **LF only** (never CR LF) on any platform — output is platform-stable by construction
- interior `\r\n` inside an object field value survives **verbatim** in the middle of an otherwise statically-indented line
- leading/trailing `\r\n` around a value is removed by `trim()` while the static `  ` indent is still emitted
- interior `\r\n` is preserved at nested-object depth while the `    • ` bullet indent is applied
- multiple CR LF-laden fields all keep stable static prefixes
- the generic-array branch uses `String(...)` **without** trim, so even outer `\r\n` survives there
- a top-level scalar string's interior `\r\n` stays intact
- CR LF-dense structures never throw

The boundary in one line: indentation is fixed static text and lines are always LF-joined regardless of OS, so platform stability is structural; CR LF inside string values is inert content — interior CR LF is preserved and only outer CR LF is trimmed (except in the untrimmed generic-array path).

## Truth-first scope note

The original task referenced `hushh-research/__tests__/utils/...` and `npm test -- hushh-research/...`. There is no top-level `__tests__` in this repo; vitest only includes `__tests__/**` under `hushh-webapp`, and the module alias is `@/lib/utils/json-to-human`. The file therefore lives at `hushh-webapp/__tests__/utils/format-carriage-buffering.test.ts` and imports via the `@/` alias. The task framing of "aligns indentation blocks … across differing terminal execution platforms" is corrected to the real behavior: there is no alignment/terminal logic — stability comes from static prefixes plus a fixed LF join.

## Verification

- `npm test -- __tests__/utils/format-carriage-buffering.test.ts` → 8/8 pass

## CI gate checklist

- [x] Truth-First — documents the real behavioral boundary (static indent + LF join; interior CR LF preserved, outer trimmed)

### DIFF
--- a/hushh-webapp/__tests__/utils/format-carriage-buffering.test.ts
+++ b/hushh-webapp/__tests__/utils/format-carriage-buffering.test.ts
@@ -1,0 +1,113 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+/**
+ * Characterization tests for formatCompleteJson
+ * (hushh-webapp/lib/utils/json-to-human.ts) when object string values contain
+ * raw Windows-style carriage-return line endings (CR LF) injected throughout
+ * the dictionary structures.
+ *
+ * TRUTH-FIRST (verified against the source):
+ *   formatCompleteJson builds its output from STATIC literal indentation
+ *   prefixes and joins every line with a literal "\n":
+ *     - object fields:   "  " + Label + ": " + value
+ *     - nested fields:   "    \u2022 " + Label + ": " + value
+ *     - array items:     "  \u2022 " + value
+ *     - section header:  "--- Label ---" (preceded by an empty "" line)
+ *     - final assembly:  lines.join("\n")
+ *   There is NO indentation-alignment computation, NO terminal/platform
+ *   detection, and NO CR/LF normalization. The ONLY mutation applied to string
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `14256445`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4108, opened 2026-06-29
test(asset-allocation-donut): cover chart accessible label

Divya-rajendran009:
## Summary
- Adds an explicit accessible label to the asset allocation donut chart.
- Covers that label with a focused component test.

## Reviewer Proof
- Surface: AssetAllocationDonut chart and its focused test only.
- Production contract: renders the real chart component; no module mocks.
- No overlap: new focused test file plus the chart aria contract it asserts.
- DCO signed; base: integration/pr-train.

## Validation
- git diff --cached --check
- Web (Next.js) via GitHub Actions


kushaltrivedi5 (changes requested):
## Changes Requested

PR #4111, opened 2026-06-29
test(accordion): cover data-slot contract on root and item

M-Rajith:
Covers two data-slot contracts on Accordion:
- data-slot="accordion" on the root element
- data-slot="accordion-item" on each item

Attach point: hushh-webapp/components/ui/accordion.tsx
Single file, single surface.


kushaltrivedi5 (approved):

Re-reviewed at head `0122419f`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4141, opened 2026-06-29
chore(ui): silence top app bar delete account error stack trace in production

Roopmann30:
### Description

Wraps a raw `console.error` inside the `TopAppBar` account deletion handler with a production environment check. This prevents exposing internal IAM routing exceptions and API deletion stack traces to end-users if deleting an account fails, while preserving the intended UI error toast.


kushaltrivedi5 (approved):

Re-reviewed at head `f526fa0a`: path-clean, MERGEABLE, green CI Status Gate, base integration/pr-train, not self-pushed. Collision cleared this pass. Enqueued to merge queue.

PR #4144, opened 2026-06-29
fix(ui): add missing type=button to sources expander

Roopmann30:
### Description

Removes a redundant `console.warn` leak in the `AgentChatWorkspace` TTS failure handler. If agent voice synthesis fails, the system already gracefully recovers and pushes a fallback error message directly into the chat UI. Removing this log keeps the client console clean without needing environment checks.

## 📌 Impact Map (Required)

- Routes touched:
  - [x] `components/agent/agent-chat-workspace.tsx`

- Trust boundary touched:
  - [x] none (purely client UI logging removal)


kushaltrivedi5 (changes requested):
## Changes Requested

PR #4145, opened 2026-06-29
fix(notifications): validate unregister platform values

KARTHI-77:
### What changed

* Validate `platform` values in the `/api/notifications/unregister` endpoint.
* Return HTTP 400 when an unsupported platform is provided.
* Add a regression test covering invalid platform input.

### Why

The `/register` endpoint already validates supported platform values (`web`, `ios`, `android`), while `/unregister` accepted arbitrary values and forwarded them to the service layer. This change makes the behavior consistent across both endpoints and prevents invalid input from reaching the service.

### Testing

```bash
uv run pytest consent-protocol/tests/test_notifications_routes.py -v
```

Result:

```
14 passed
```

### DIFF
--- a/consent-protocol/api/routes/notifications.py
+++ b/consent-protocol/api/routes/notifications.py
@@ -86,6 +86,12 @@ async def unregister_push_token(request: Request):
     user_id = body.get("user_id") or body.get("userId") or firebase_uid
     platform = body.get("platform")
 
+    if platform is not None and platform not in ("web", "ios", "android"):
+        raise HTTPException(
+            status_code=400,
+            detail="platform must be one of: web, ios, android",
+        )
+
     if firebase_uid != user_id:
         raise HTTPException(
             status_code=403,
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Thank you for this contribution — the intent here is sound and it's aimed at the right surface. This change lands on a trust-boundary file (`consent-protocol/api/routes/notifications.py`), so before it can merge it needs a maintainer-authored security patch staged on top to make the enforcement independently stronger, not just preserved while touched.

**Blocker:** The diff touches `consent-protocol/api/routes/notifications.py`, a consent/identity/finance trust boundary. Per the PR-train write contract, a change on this surface must demonstrably strengthen DB-backed enforcement / scope / caller-credential shape — a maintainer patch will be staged to guarantee that bar before queueing.

**Path To Merge:** Keep the PR open; a maintainer security patch will be staged on top at `consent-protocol/api/routes/notifications.py` (co-authored to you), then the combined head is re-reviewed and queued. No action needed from you unless you want to tighten validation/error-surface at that attach point yourself first.

**Proof Needed:** A test asserting the strengthened behavior at `consent-protocol/api/routes/notifications.py` (an existing/your/maintainer-added test) green, plus the CI Status Gate green at the patched head.

kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `eebf35cf`

PR #4191, opened 2026-06-30
fix(services): add non-enumerable fallback marker to distinguish profile load failures

abdulbeast4-creator:
## Summary

`KaiProfileService.getProfile` catches any load failure (vault decrypt / network)
and returns `createDefaultProfile()`. Today that degraded "system error" default
is byte-for-byte identical to a genuine brand-new-user default, so a caller cannot
tell the two apart. This PR adds an internal, non-enumerable seam that records
*that* a fallback was produced by a transient error, plus a sanctioned read guard,
and pins the contract with a characterization test.

Scope note: this lays the groundwork (marker + guard + tests). It does **not** yet
change any UI/consumer behavior — no caller reads the guard yet (see "Known
follow-up"). Framed honestly, this is an internal seam, not a user-facing fix.

## What Changed (Scope)

- **Fallback marker:** Private `markTransientFallback()` attaches a non-enumerable
  `__transientLoadError` marker to the fallback profile inside the `getProfile`
  catch block. The marked fallback is cached under the SHORT TTL so a recovered
  load re-resolves the real profile quickly.
- **Leak prevention:** The marker is non-enumerable by design, so it is excluded
  from `Object.keys()`, object spread, and `JSON.stringify()` — it cannot pollute
  the `KaiProfileV2` shape or leak into a vault write.
- **Exported guard:** `isTransientProfileFallback(profile)` is the only sanctioned
  read path for the marker.
- **No telemetry-schema change:** Deliberately avoids adding an event to the
  cross-surface observability registry; the seam is local to
  `hushh-webapp/lib/services/kai-profile-service.ts`.

## Diff scope (verified)

- `hushh-webapp/lib/services/kai-profile-service.ts` — +45 / -1
  (new helpers added after `isKaiOnboardingCompleted` near lines ~520–555, plus the
   `catch` block in `getProfile` near lines ~578–588).
- `hushh-webapp/__tests__/lib/services/kai-profile-transient-fallback.test.ts` — new file.

## Verification

`npx vitest run hushh-webapp/__tests__/lib/services/kai-profile-transient-fallback.test.ts`
→ **1 file, 6 tests passed.** The suite covers:
- guard returns `true` for marked profiles; `false` for normal / null / undefined,
- marker excluded from `Object.keys` / spread / `JSON.stringify`,
- a spread copy (the persisted shape) is NOT flagged,
- marker survives a by-reference cache round-trip.

## Known follow-up (please review)

`isTransientProfileFallback` currently has **no production consumer** — the
observability/UX gap is not actually closed by this PR, only made addressable.
A follow-up must wire the guard into the Kai profile UI/load path so a degraded
default renders differently from a new-user default. Reviewers should treat this
PR as enabling infrastructure + characterization, not the user-facing fix.

### DIFF
--- a/hushh-webapp/__tests__/lib/services/kai-profile-transient-fallback.test.ts
+++ b/hushh-webapp/__tests__/lib/services/kai-profile-transient-fallback.test.ts
@@ -1,0 +1,115 @@
+/**
+ * Characterization suite — @/lib/services/kai-profile-service
+ *
+ * Pins the observability-hardening contract added to KaiProfileService:
+ * when getProfile() fails to load financial.profile it returns the canonical
+ * default profile but tags it with a NON-ENUMERABLE transient-failure marker so
+ * callers can tell a degraded "system error" default apart from a genuinely
+ * "brand-new user" default.
+ *
+ * The invariant under test is that the marker:
+ *   1. is readable only via the sanctioned isTransientProfileFallback() guard,
+ *   2. is non-enumerable, so it never leaks into JSON.stringify, object spread,
+ *      Object.keys, or a vault write — i.e. it cannot mutate the KaiProfileV2
+ *      shape that gets persisted.
+ *
+ * Implementation: hushh-webapp/lib/services/kai-profile-service.ts
+ */
+
+import { describe, expect, it } from "vitest";
+import {
+  isTransientProfileFallback,
+  type KaiProfileV2,
+} from "@/lib/services/kai-profile-service";
+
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current blockers: new_export_without_app_or_backend_caller, existing_capability_overlap_requires_review, pr_claim_changed_surface_mismatch.

### Path To Merge

### Proof Needed

PR #4199, opened 2026-06-30
fix(kai): render all supported income detail types

KARTHI-77:
## Summary

This change fixes an edge case in `IncomeDetailCard` where some supported income detail fields were not considered when determining whether to render the detailed breakdown.

Previously, the card only treated these fields as detailed income:

- `dividends_taxable`
- `dividends_qualified`
- `interest_taxable`
- `short_term_cap_gains`
- `long_term_cap_gains`

As a result, if only one of the following fields was present:

- `dividends_nontaxable`
- `interest_tax_exempt`
- `return_of_capital`

the detail section would not render even though the component contains UI for displaying those values.

### Changes

- Include all supported income detail fields when computing `hasDetail`.
- Add a regression test covering the case where `dividends_nontaxable` is the only populated detail field.

### Testing

```bash
npx vitest run __tests__/components/income-detail-card.test.tsx

### DIFF
--- a/hushh-webapp/__tests__/components/income-detail-card.test.tsx
+++ b/hushh-webapp/__tests__/components/income-detail-card.test.tsx
@@ -12,4 +12,21 @@ describe("IncomeDetailCard", () => {
     expect(screen.queryByText("Income")).toBeNull();
     expect(container.firstChild).toBeNull();
   });
+
+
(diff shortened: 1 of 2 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested

PR #4214, opened 2026-06-30
test: verify generic array item placeholder fallback

imsharukhan:
## What

Adds coverage for the generic-array object placeholder fallback path in `formatCompleteJson`.

## Why

Existing tests verify generic-array rendering when object elements contain usable values. This test covers the remaining fallback path where all object properties are nullish and the formatter emits the default `"Item"` placeholder.

The new test verifies:

* object elements with only nullish values render the fallback placeholder
* the generic-array rendering path remains stable for empty-value objects

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/utils/format-complete-json-bounds.test.ts


kushaltrivedi5 (approved):

Re-reviewed at head `84db08a9`: path-clean, MERGEABLE, base integration/pr-train, green CI Status Gate. Enqueued to merge queue.

PR #4224, opened 2026-06-30
test: verify isOneOnboardingRoute non-onboarding paths

imsharukhan:
## What

Adds coverage for negative path handling in `isOneOnboardingRoute`.

## Why

Existing tests verify onboarding route detection, but do not directly verify behavior for several non-onboarding path variants.

The new test verifies:

* empty paths return false
* root paths return false
* near-match onboarding paths return false
* onboarding paths outside the expected namespace return false

## Scope

Test-only change.

No production code modified.

## Validation

npx vitest run **tests**/navigation/routes.test.ts


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4237, opened 2026-06-30
test(utils): cover cn responsive class merging

gracyezhilkumar07:
## Summary

Adds unit test coverage for the `cn` utility to verify that responsive Tailwind CSS classes are preserved while conflicting base classes are merged correctly.

## Changes Made

* Added a test to validate responsive class merging behavior in `cn()`.
* Verifies that responsive utility classes (for example, `md:p-4`) are preserved while conflicting base classes are resolved correctly.
* Extends utility test coverage with an additional Tailwind CSS edge case to help prevent regressions.

## Test

```bash
```

### DIFF
--- a/hushh-webapp/__tests__/utils/cn.test.ts
+++ b/hushh-webapp/__tests__/utils/cn.test.ts
@@ -17,4 +17,7 @@ describe("cn", () => {
     "flex items-center",
   );
 });
+it("preserves responsive classes while merging conflicts", () => {
+  expect(cn("p-2 md:p-4", "p-6")).toBe("md:p-4 p-6");
+});
 });


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Appreciated contribution — the intent reads well. It isn't in a directly mergeable state at the current head yet, so it's being held with this record rather than queued.

**Blocker:** GitHub does not report this PR as cleanly MERGEABLE at the current head (pending mergeability / checks not settled), so it cannot be safely enqueued this pass.

**Path To Merge:** Ensure the branch is rebased on the latest `integration/pr-train` and that the CI Status Gate is green; the next pass will re-review at the settled head and queue it if it is clean.

**Proof Needed:** MERGEABLE state + green CI Status Gate at the current head.

PR #4242, opened 2026-06-30
test(utils): cover minimal internal navigation request

gracyezhilkumar07:
## Summary

Adds test coverage for the browser navigation utility when only the required navigation target is provided.

## Changes Made

* Added a unit test to verify that `requestInternalAppNavigation` dispatches a navigation request when only the required `href` property is supplied.
* Confirms that the dispatched event contains the expected payload without optional navigation flags.
* Extends browser navigation utility coverage with a minimal navigation request scenario to help prevent regressions.

## Test

```bash
npx vitest run __tests__/utils/browser-navigation.test.ts
```


kushaltrivedi5 (approved):

Re-reviewed at head `b4ae7a06`: path-clean, MERGEABLE, base integration/pr-train, green CI Status Gate. Enqueued to merge queue.

PR #4243, opened 2026-06-30
test(utils): cover minimal internal navigation request

gracyezhilkumar07:
## Summary

Adds test coverage for the browser navigation utility when only the required navigation target is provided.

## Changes Made

* Added a unit test to verify that `requestInternalAppNavigation` dispatches a navigation request when only the required `href` field is supplied.
* Confirms that the dispatched event contains the expected payload without optional `replace` or `scroll` properties.
* Extends browser navigation utility coverage with a minimal navigation request scenario to help prevent regressions.

## Test

```bash
npx vitest run __tests__/utils/browser-navigation.test.ts
```

### DIFF
--- a/hushh-webapp/__tests__/utils/browser-navigation.test.ts
+++ b/hushh-webapp/__tests__/utils/browser-navigation.test.ts
@@ -25,4 +25,23 @@ describe("requestInternalAppNavigation", () => {
       },
     ]);
   });
+  
+it("uses default navigation options when optional flags are omitted", () => {
+  const details: unknown[] = [];
+
+  window.addEventListener(INTERNAL_APP_NAVIGATION_REQUEST_EVENT, (event) => {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested

**Direction:** Appreciated contribution — the intent reads well. It isn't in a directly mergeable state at the current head yet, so it's being held with this record rather than queued.

**Blocker:** GitHub does not report this PR as cleanly MERGEABLE at the current head (pending mergeability / checks not settled), so it cannot be safely enqueued this pass.

**Path To Merge:** Ensure the branch is rebased on the latest `integration/pr-train` and that the CI Status Gate is green; the next pass will re-review at the settled head and queue it if it is clean.

**Proof Needed:** MERGEABLE state + green CI Status Gate at the current head.

PR #4245, opened 2026-06-30
test(services): characterize scoreInvestmentHorizon/computeRiskScore handling of out-of-range investment horizon values

abdulbeast4-creator:
## Summary

Adds an isolated **characterization** test that pins the public investment-horizon
scoring contract in `@/lib/services/kai-profile-service`, so future refactors
cannot silently change how unrecognized / extreme / malformed horizon values are
handled.

Test-only change. No production source is modified.

## What is actually verified (truth-first)

This PR was scoped against the real source, not assumed behavior:

- `normalizeProfile` and `normalizeInvestmentHorizon` are **module-private** and
  not exported, so they cannot be imported directly. The sanctioned public
  surface is `scoreInvestmentHorizon`, `computeRiskScore`, and `mapRiskProfile`.
- `scoreInvestmentHorizon` is **total**: `0` for `short_term`, `1` for
  `medium_term`, and — by the final `return 2` — `2` for any other truthy
  string. It returns `null` only for falsy input (`null` / `""` / `undefined`).
  It does **not** collapse unknown strings to `null`.
- `computeRiskScore` returns `null` only when a component preference is falsy,
  and never throws.

The tests assert exactly this behavior (e.g. an unknown horizon scores `2`, and
`computeRiskScore` with an extreme horizon yields `4` → `"balanced"`).

## Scope

- `hushh-webapp/__tests__/lib/services/profile-investment-horizon.test.ts` (new, 5 tests)

## Verification gate

```
npx vitest run hushh-webapp/__tests__/lib/services/profile-investment-horizon.test.ts
# 1 file, 5 tests passed
```

## Base policy note

Targets `integration/pr-train` per `scripts/ci/verify-pr-base-policy.py`, which
fail-closes direct PRs into `main`. This intentionally goes through the PR train.

## Reviewer risk assessment

- Test-only; zero production runtime change.
- No new exports, no behavior added — purely characterizes shipped behavior.
- The one judgment call: this pins existing behavior (including the unknown→`2`
  fall-through). If that fall-through is considered a latent bug rather than
  intended, that is a separate production fix and is intentionally **not** done
  here; this test would simply be updated alongside it.

### DIFF
--- a/hushh-webapp/__tests__/lib/services/profile-investment-horizon.test.ts
+++ b/hushh-webapp/__tests__/lib/services/profile-investment-horizon.test.ts
@@ -1,0 +1,90 @@
+/**
+ * Characterization suite — @/lib/services/kai-profile-service
+ *
+ * Pins the PUBLIC scoring contract for investment-horizon handling. The goal is
+ * to lock down how the exported risk-scoring surface behaves when it is fed an
+ * unrecognized / extreme / malformed horizon value, so future refactors cannot
+ * silently change the fallback shape.
+ *
+ * IMPORTANT (verified against source, not assumed):
+ *   - `normalizeInvestmentHorizon` is module-private and is NOT exported, so it
+ *     cannot be imported directly. The sanctioned public surface is the
+ *     `score*` helpers plus `computeRiskScore` / `mapRiskProfile`.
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4247, opened 2026-06-30
test(services): characterize exported profile service timestamp validation boundaries

abdulbeast4-creator:
## Summary
Pins down the boundary behavior of the **exported** profile-service timestamp surface when malformed, truncated, or excessively long "ISO" date strings are passed through. Test-only; no production source changed.

## Truth-first correction (verified against source)
The working premise was that the public service "gracefully normalizes or ignores malformed temporal inputs." That is only **partially** true, and the PR pins the *real* contract rather than an assumed one:

- `resolveHorizonAnchorAt` is the only **exported** function taking timestamp-shaped params. It does **no** date parsing/validation — it is a pure opaque string selector (`keep_original -> previousAnchorAt ?? now`, else `now`). Malformed input is returned **verbatim**; it never throws and never coerces to a `Date`.
- The string sanitation that *does* exist (`normalizeOptionalIso`) is **module-private / unexported** — it only type-guards, trims, and maps empty-string to `null`. It does **not** reject non-ISO content. It is therefore deliberately not imported here (importing a private would be a false contract).

## What this characterizes
- `from_now` returns `now` verbatim across 8 malformed inputs.
- `keep_original` returns the malformed `previousAnchorAt` verbatim (no validation/coercion).
- `keep_original` with no previous anchor falls back to `now`.
- Never throws; output stays a raw `string`, never a `Date`.
- A 50k-char timestamp is preserved byte-for-byte (no silent truncation in the selector).

## Verification gate
`npx vitest run hushh-webapp/__tests__/lib/services/profile-timestamp-bounds.test.ts` → 1 file, **5 tests passed**, zero production drift.

## Reviewer risk assessment (no risks)
- Zero production runtime change; no new exports; no observability-schema touch; no consent/vault surface.
- Correct base: targets `integration/pr-train` (not `main`).
- DCO-signed single-file commit.
- The one honest judgment: tests **pin existing** non-validating passthrough. If non-validation is later deemed a defect, that is a separate production fix and this test would move with it (it documents the current boundary, not an endorsement).

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/lib/services/profile-timestamp-bounds.test.ts
+++ b/hushh-webapp/__tests__/lib/services/profile-timestamp-bounds.test.ts
@@ -1,0 +1,98 @@
+/**
+ * Characterization suite — @/lib/services/kai-profile-service
+ *
+ * Pins how the PUBLIC profile-service timestamp surface behaves when fed
+ * malformed, truncated, or excessively long "ISO" date strings.
+ *
+ * TRUTH-FIRST NOTE (verified against source, not assumed):
+ *   - The proposed premise was that the public service "gracefully normalizes
+ *     or ignores malformed temporal inputs." That is only half true. The actual
+ *     shipped behavior is:
+ *       * `resolveHorizonAnchorAt` is the only EXPORTED function that takes
+ *         timestamp-shaped params. It performs NO date parsing or validation.
+ *         It is pure opaque string selection:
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4248, opened 2026-06-30
test(services): characterize computeRiskScore constraints under null inputs

abdulbeast4-creator:
## Summary
Characterizes how the exported `computeRiskScore` in `hushh-webapp/lib/services/kai-profile-service.ts` responds to "empty"-shaped inputs: null/undefined fields, an entirely empty object, and unexpected empty arrays. Test-only; no production source changed.

## Truth-first findings (verified against source)
`computeRiskScore` fans out to three per-field scorers, each guarded by `if (!value) return null`, and returns `null` if **any** sub-score is `null`, else the numeric sum. The real, shipped boundaries:

- `null` / `undefined` field → falsy → sub-score `null` → **total `null`**.
- Empty object `{}` (fields missing) → `undefined` → **total `null`**.
- Empty array `[]` → **truthy**, so it bypasses the `if (!value)` guard and falls through to each scorer's catch-all `return 2`, summing to **6**. This is the surprising boundary worth pinning — "empty" is not uniformly treated as "absent".
- Whole `preferences === null` → **throws** (no top-level guard; property access on `null`).

The suite documents what ships today (not a "should"), so any future hardening changes it visibly.

## Verification gate
`npx vitest run hushh-webapp/__tests__/lib/services/risk-score-nulls.test.ts` → 1 file, **6 tests passed**, zero production drift.

## Reviewer risk assessment (no risks)
- Zero production runtime change; no new exports; no consent/vault/observability surface touched.
- Correct base: targets `integration/pr-train`.
- DCO-signed single-file commit.
- Off-contract inputs (`[]`, `null`) are fed via a local `unknown` cast confined to the test — no production type loosening.

## CI gate checklist
- [x] DCO Verified

### DIFF
--- a/hushh-webapp/__tests__/lib/services/risk-score-nulls.test.ts
+++ b/hushh-webapp/__tests__/lib/services/risk-score-nulls.test.ts
@@ -1,0 +1,84 @@
+/**
+ * Characterization suite — @/lib/services/kai-profile-service · computeRiskScore
+ *
+ * Pins how the exported risk-score calculation responds to "empty"-shaped
+ * inputs: null/undefined preference fields, empty object structures, and
+ * unexpected empty arrays.
+ *
+ * TRUTH-FIRST NOTE (verified against source, not assumed):
+ *   computeRiskScore delegates to three per-field scorers, each guarded by
+ *   `if (!value) return null`. computeRiskScore then returns null if ANY of the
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4253, opened 2026-06-30
test(services): characterize risk categorization coercion via public profiles

abdulbeast4-creator:
## Summary
Characterizes how the public risk-categorization processing surface of `KaiProfileService` actually handles risk text variants (unexpected casing, mixed/padded spaces).

> **Truth correction:** the original premise — that this surface *normalizes* casing/spacing into the expected constants — is **not** how the code behaves today. These tests pin the real behavior rather than an aspirational one.

## Technical Context & Scope
- **Target surface (verified):** exported `scoreInvestmentHorizon`, `scoreDrawdownResponse`, `scoreVolatilityPreference`, `computeRiskScore`, `mapRiskProfile` in `hushh-webapp/lib/services/kai-profile-service.ts`.
- **Actual behavior pinned:** scoring uses **strict `===`** against canonical lowercase tokens (`short_term`, `reduce`, `small`, …). There is **no** `toLowerCase()`, **no** `trim()`, and **no** whitespace collapsing.
  - Falsy inputs (e.g. `""`) hit the `if (!value) return null` guard.
  - Any *unrecognized truthy* value (mis-cased, space-padded, internally spaced, or whitespace-only) **falls through to the maximum ordinal `2`**, not back to its intended category.
  - End-to-end, an all-variant payload therefore coerces silently into the `"aggressive"` constant (score 6) — deterministic, but misleading.
- **Risk mitigation:** test-only addition; no runtime/source change. Confined to a new spec file.

## Verification Results
- `npx vitest run hushh-webapp/__tests__/lib/services/risk-categorization-coercion.test.ts` → **12 passed**.

## Why characterize instead of "fix"
If real case/space normalization is later desired, it becomes a visible, reviewed change because these tests will turn red, instead of the current silent coercion going unnoticed.

## CI Gate Checklist
- [x] DCO Signed Verification

### DIFF
--- a/hushh-webapp/__tests__/lib/services/risk-categorization-coercion.test.ts
+++ b/hushh-webapp/__tests__/lib/services/risk-categorization-coercion.test.ts
@@ -1,0 +1,141 @@
+import { describe, it, expect } from "vitest";
+
+import {
+  scoreInvestmentHorizon,
+  scoreDrawdownResponse,
+  scoreVolatilityPreference,
+  computeRiskScore,
+  mapRiskProfile,
+  type RiskProfile,
+} from "@/lib/services/kai-profile-service";
+
+/**
+ * Characterization tests for the public risk-categorization processing surface
+ * of KaiProfileService when fed risk text variants (unexpected casing, mixed /
+ * padded spaces).
+ *
+ * IMPORTANT TRUTH CORRECTION
+ * --------------------------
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4254, opened 2026-06-30
test(services): characterize factory default profiles for deep immutability

abdulbeast4-creator:
## Summary
Characterizes the real reference/mutation behavior of the public structural factory surface — `buildKycWorkflowArtifact` and `mergeKycWorkflowArtifact` — when consumers mutate the object tree of a generated artifact.

> **Truth correction:** the original premise — that mutating object *leaves* "does not modify reference values for subsequent requests" — is only **partially** true. These tests pin the actual behavior rather than an aspirational deep-immutability guarantee.

## Technical Context & Scope
- **Target surface (verified):** `buildKycWorkflowArtifact`, `mergeKycWorkflowArtifact` in `hushh-webapp/lib/services/kyc-pkm-write-service.ts`. (Note: `kai-profile-service.ts` exposes **no** baseline/empty factory — `createDefaultProfile`/`normalizeProfile` are module-private — so the genuinely public structural factory is the KYC artifact builder.)
- **What is TRUE:** there is **no** shared module-level "global default" singleton. `buildKycWorkflowArtifact` allocates a **new top-level object** on every call, so two builds never alias at the top level and one request cannot clobber the next request's top-level object.
- **What is FALSE / overstated:** the factory does a **shallow spread** (`{ ...artifact, last_updated, schema_version: 1 }`), not a deep clone. Nested leaves (`checks`, `pending_requirements`, `completed_requirements`, `sent_replies`) remain the **same references** as the caller input — mutating a leaf through the returned artifact leaks back into the input. `mergeKycWorkflowArtifact` passes the requirement arrays through **by reference** as well (though it does rebuild `checks` and `sent_replies`).

## Verification Results
- `npx vitest run hushh-webapp/__tests__/lib/services/default-profile-immutability.test.ts` → **7 passed**.

## Why characterize instead of "fix"
If deep-immutability of factory output is later desired, these tests turn red and force a visible, reviewed change instead of a silent behavioral shift.

## CI Gate Checklist
- [x] DCO Signed Verification

### DIFF
--- a/hushh-webapp/__tests__/lib/services/default-profile-immutability.test.ts
+++ b/hushh-webapp/__tests__/lib/services/default-profile-immutability.test.ts
@@ -1,0 +1,141 @@
+import { describe, it, expect } from "vitest";
+
+import {
+  buildKycWorkflowArtifact,
+  mergeKycWorkflowArtifact,
+  type KycWorkflowArtifact,
+  type KycWorkflowArtifactInput,
+} from "@/lib/services/kyc-pkm-write-service";
+
+/**
+ * Characterization tests for the public structural factory surface of the
+ * KYC workflow PKM write service: `buildKycWorkflowArtifact` and
+ * `mergeKycWorkflowArtifact`.
+ *
+ * TRUTH CORRECTION — read before trusting the title
+ * -------------------------------------------------
+ * The requested premise ("attempts to mutate object leaves do not modify the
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4258, opened 2026-06-30
test(services): characterize public user session context token validation bounds

abdulbeast4-creator:
## Summary

Establishes explicit contract-test boundaries around the public session
extraction utility when it handles malformed headers, embedded JSON-structure
characters, foreign auth schemes, and pathologically long token segments.

## Verified target (truth-first)

The exported public session parsing utility is
`extractSessionToken(request: Request): string | null` in
`hushh-webapp/lib/auth/validate.ts`. It reads a session token in priority order
from (1) the `X-Session-Token` header, (2) the `sessionToken` query param, and
(3) the `Authorization` header *only when prefixed with `HCT:`*. It performs no
JSON parsing or signature verification — that is delegated to
`validateSessionToken` / the backend — so the genuine "bounds" contract is its
clean-failure signal: it returns `null` (never throws, never `undefined`) when
no valid source matches, and returns the raw value verbatim when one does. The
companion `extractConsentToken(body)` is also pinned for its null-on-missing
contract.

## Technical Context & Scope

- **Target Surface:** Exported session/consent extraction helpers in
  `lib/auth/validate.ts`.
- **Risk Mitigation:** No changes to active core business components — test-only.
- **Design Invariant:** Extraction fails cleanly to `null` and never throws,
  even on huge or structurally broken inputs.

## What is covered (14 tests)

`extractSessionToken`:
- bare request with no token sources → `null`
- non-HCT `Bearer` Authorization header → `null`
- foreign `Basic` auth scheme → `null`
- unrelated query params only → `null`
- embedded JSON-structure characters in `X-Session-Token` → returned verbatim, no throw
- well-formed `HCT:` Authorization header → honored as fallback
- header / query / auth precedence ordering
- `sessionToken` query param read when no header present
- 100k-char non-matching auth header → no throw, `null`
- 50k-char `X-Session-Token` → preserved verbatim, no truncation

`extractConsentToken`:
- empty body → `null`
- unrelated keys only → `null`
- camelCase `consentToken` read
- snake_case `consent_token` fallback

## Verification Results

```
npx vitest run __tests__/lib/services/session-token-bounds.test.ts
# 1 file, 14 tests passed
```

## CI Gate Checklist

- [x] DCO Signed Verification

### DIFF
--- a/hushh-webapp/__tests__/lib/services/session-token-bounds.test.ts
+++ b/hushh-webapp/__tests__/lib/services/session-token-bounds.test.ts
@@ -1,0 +1,155 @@
+import { describe, expect, it } from "vitest";
+
+/**
+ * Characterization: session token extraction bounds on malformed / extreme input.
+ *
+ * Verified repo truth (truth-first)
+ * ---------------------------------
+ * The exported public session parsing utility is `extractSessionToken(request:
+ * Request): string | null` in `hushh-webapp/lib/auth/validate.ts`. It reads a
+ * session token from, in priority order:
+ *   1. the `X-Session-Token` header,
+ *   2. the `sessionToken` query param,
+ *   3. the `Authorization` header *only when it starts with `HCT:`*.
+ * When none of those produce a value it returns `null`. It performs no JSON
+ * parsing and no signature validation — that is delegated to
+ * `validateSessionToken` / the backend. So the genuine "bounds" contract to pin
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4259, opened 2026-06-30
test(services): characterize transparent data tracking config object isolation

abdulbeast4-creator:
## Summary

Validates config dictionary coercion logic inside the application's transparent
data tracking interface — the analytics event sanitizer that every tracking
call passes through before any adapter sees the data.

## Verified target (truth-first)

The public tracking setup path normalizes every payload through
`validateAndSanitizeEvent(eventName, payload)` in
`hushh-webapp/lib/observability/schema.ts` (invoked by `trackEvent` in
`lib/observability/client.ts`). Given an event name and a config-like options
object, it returns:

- `sanitized` — only the allow-listed primitive fields for that event (valid tracks),
- `droppedKeys` — every key that is unknown for the event, denylisted as
  sensitive (PII/secret/finance patterns), non-primitive, or an opaque/email-like
  value (missing/rejected entities),
- `ok` — true only when nothing was dropped.

Coercion rules pinned from source: unknown keys are dropped; `DENYLIST_KEY_REGEX`
keys are dropped except the preserved `route_id`; non-primitive values (arrays,
nested objects) are dropped as missing entities; email-shaped and opaque
high-entropy (`>=24` char) string values are dropped; `null`, numbers, booleans,
and short plain strings on allowed keys are retained verbatim.

## Technical Context & Scope

- **Target Surface:** Public transparency configuration sanitizer (`schema.ts`).
- **Risk Mitigation:** Isolated completely to characterization tests. Zero operational execution impact.
- **Design Invariant:** Affirms processing sequences isolate configuration blocks safely without field leakage cross-contamination between successive calls.

## What is covered (9 tests)

- allow-listed valid tracks retained, `ok === true`, no drops
- unknown config keys routed to `droppedKeys`, valid siblings retained
- array-valued options dropped as missing entities (non-primitive)
- nested-object options dropped, sibling valid track preserved
- denylisted sensitive keys (`user_id`, `email`, `auth_token`) dropped, `route_id` preserved
- email-shaped and opaque high-entropy string values dropped on allowed keys
- `null` / numeric / boolean primitives retained on allowed keys
- empty config object → empty sanitized set, no throw, `ok === true`
- per-call config block isolation — no field leakage across calls

## Verification Results

```
npx vitest run __tests__/lib/services/transparent-tracking-config.test.ts
# 1 file, 9 tests passed
```

## CI Gate Checklist

- [x] DCO Signed Verification

### DIFF
--- a/hushh-webapp/__tests__/lib/services/transparent-tracking-config.test.ts
+++ b/hushh-webapp/__tests__/lib/services/transparent-tracking-config.test.ts
@@ -1,0 +1,187 @@
+import { describe, expect, it } from "vitest";
+
+/**
+ * Characterization: transparent tracking config object isolation.
+ *
+ * Verified repo truth (truth-first)
+ * ---------------------------------
+ * The public tracking setup path normalizes every analytics payload through
+ * `validateAndSanitizeEvent(eventName, payload)` in
+ * `hushh-webapp/lib/observability/schema.ts` before any adapter is invoked
+ * (see `trackEvent` in `lib/observability/client.ts`). That function is the
+ * "transparent data tracking" config coercion gate. Given an event name and a
+ * config-like options object, it returns:
+ *   - `sanitized`: only the allow-listed primitive fields for that event
+ *     (the valid tracks),
+ *   - `droppedKeys`: every key that is unknown for the event, denylisted as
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4261, opened 2026-07-01
fix(pkm): guard missing metadata domains in loadFullBlob

KARTHI-77:
## Summary

Guard against missing `metadata.domains` when loading the full PKM blob.

### Changes

- Default `metadata.domains` to an empty array before mapping over it.

### Why

Previously `loadFullBlob()` called `metadata?.domains.map(...)`, which could throw if metadata existed but `domains` was undefined. This change safely defaults to an empty array while preserving existing behavior when no domains are available.

### DIFF
--- a/hushh-webapp/lib/services/personal-knowledge-model-service.ts
+++ b/hushh-webapp/lib/services/personal-knowledge-model-service.ts
@@ -2445,7 +2445,7 @@ export class PersonalKnowledgeModelService {
     } catch {
       metadata = null;
     }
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `676676af`: the contributor's pkm-privacy change is self-sufficient, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train — no separate maintainer patch needed. Enqueued to merge queue; it closes on merge.

PR #4274, opened 2026-07-02
fix(ui): silence vault status check error logging leak

Roopmann30:
### Description

Silences explicit `console.warn` and `console.error` leaks in `components/vault/vault-flow.tsx`. These traces (`"Vault mode detection failed, defaulting to passphrase:"` and `"Vault status check failed:"`) were firing when background vault presence or metadata checks failed. Because the UI gracefully handles these exact failure scenarios natively—by falling back to the passphrase method and rendering a formatted `toInvestorMessage`—the explicit `console` logs were removed to prevent internal backend error exceptions from leaking into the browser console during offline scenarios.


kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `ead95abf`

PR #4281, opened 2026-07-03
chore(ui): silence onboarding state check error trace

Roopmann30:
### Description

Silences a redundant `console.warn` trace in `components/kai/onboarding/kai-onboarding-guard.tsx`. This warning (`"Failed to check onboarding state:"`) fired whenever the Kai onboarding profile check failed on initial component mount. Since the guard component natively catches this exception and handles it safely—by setting an internal error state that renders a localized fallback UI for the user—the explicit `console.warn` was unnecessary. Removing it cleans up the browser console logs from internal debug traces during transient network conditions.


kushaltrivedi5 (approved):
## Approved: trust-boundary diff verified benign/security-positive at head `c0dd18ce`

PR #4282, opened 2026-07-03
fix(ui): silence phone verification error leak

Roopmann30:
### Description

Silences an explicit `console.error` leak in `components/auth/phone-verification-flow.tsx`. This error trace (`"Failed to start verification:"`) fired whenever a phone verification attempt failed (e.g., invalid number format or rate-limiting). Since the component actively catches this exception and handles it securely by logging a telemetry event and displaying a formatted UI toast to the user, the explicit `console.error` log was unnecessary and merely leaked backend trace noise into the browser console.

### DIFF
--- a/hushh-webapp/components/auth/phone-verification-flow.tsx
+++ b/hushh-webapp/components/auth/phone-verification-flow.tsx
@@ -289,7 +289,6 @@ export function PhoneVerificationFlow({
           resendCode ? "A new verification code has been sent." : "Verification code sent."
         );
       } catch (error) {
-        console.error("[PhoneVerificationFlow] Failed to start verification:", error);
         trackEvent("phone_verification_started", {
           action: mode,
           result: "error",


kushaltrivedi5 (approved):
## Approved: auth-path diff verified benign/security-positive at head `6a7d8bc2`

Thank you for tightening up the phone-verification error handling — this is exactly the kind of small, careful hygiene fix that improves the product.

**Diff triage:** The only change is removing a `console.error` that logged the raw verification failure detail in the `/auth` phone-verification flow. This does **not** weaken any auth/token/verification boundary — the `catch (error)` still surfaces a user-facing `morphyToast.error` and fires the `phone_verification_started` error analytics event. Removing the raw console log is security-**positive** (stops leaking underlying error detail to the browser console in an auth surface).

**Verified:** `error` remains referenced (line 299), so no unused-variable break; CI Status Gate green (10 checks SUCCESS), MERGEABLE, base `integration/pr-train`, exact head `6a7d8bc2`. Enqueued.

PR #4285, opened 2026-07-03
test(services): characterize legal compliance versioning payload matchers

abdulbeast4-creator:
## Summary

Characterizes the real, exported legal payload configuration surface —
`KAI_LEGAL_DOCUMENTS` in `hushh-webapp/lib/legal/kai-legal-content.ts` — and pins
how a caller-side "unknown key falls back to baseline" pattern behaves for
version-like / alternative keys. Test-only; no source changes.

## Premise Correction (truth-first)

The requested framing ("legal payload **version** configuration tables" with a
"version-stamp parser that falls back onto baseline regulatory text") does **not
exist** in the repo:

- There is no version-stamped legal table and no version-parsing/fallback code.
- The only real legal config is `KAI_LEGAL_DOCUMENTS`, a static
  `Record<KaiLegalDocumentType, KaiLegalDocument>` keyed strictly by document
  **type** (`"terms" | "privacy"`), not by a semver string.
- `updatedAt` is a human month label (`"February 2026"`), not a parsed version.
- Consumers (`AuthStep.tsx`, `AuthLegalDialog.tsx`) index the record directly by
  `KaiLegalDocumentType`.

Rather than import a non-existent parser, this PR pins the actual contract that
the premise cares about: baseline completeness and safe fallback-to-baseline for
unknown/version-like keys, using the `table[key] ?? baseline` pattern consumers
already rely on.

## What is characterized

- **Baseline shape:** exactly the `terms` + `privacy` keys; every document has a
  non-empty `title`, `summary`, `updatedAt`, and at least one section; every
  section has a non-empty title and ≥1 point; `updatedAt` is confirmed to be
  NOT a semver stamp.
- **Fallback matrix:** version-like/unknown keys (`v0.0.0-beta`, `1.0.0`,
  `terms@2026-02`, `TERMS`, `privacy-policy`, `""`) all resolve to the baseline
  `terms` document **without dropping any property**.
- **Exact keys** (`terms`, `privacy`) still resolve to their own document.
- **Real edge-case caveat pinned:** the naive `table[key] ?? baseline` pattern
  does **not** protect against the `"__proto__"` key, because
  `({})["__proto__"]` resolves to `Object.prototype` (a truthy object) so the
  `??` never fires. This is characterized as current behavior, not asserted as
  safe — a genuine finding for any future hardening.

## Verification Results

- `npx vitest run --root hushh-webapp __tests__/lib/services/legal-versioning-matchers.test.ts`
  → **15/15 passed** locally.
- Note: the webapp vitest `include` glob is `__tests__/**/*.test.{ts,tsx}`, so
  the file uses the `.test.ts` extension (not `.spec.ts`) to be discovered.

## CI Gate Checklist

- [x] DCO Signed Verification
- [x] Test-Only Surface Isolation (single new test file, zero source edits)

### DIFF
--- a/hushh-webapp/__tests__/lib/services/legal-versioning-matchers.test.ts
+++ b/hushh-webapp/__tests__/lib/services/legal-versioning-matchers.test.ts
@@ -1,0 +1,136 @@
+import { describe, it, expect } from "vitest";
+import {
+  KAI_LEGAL_DOCUMENTS,
+  type KaiLegalDocumentType,
+} from "@/lib/legal/kai-legal-content";
+
+/**
+ * Characterization: legal payload configuration matchers.
+ *
+ * ─────────────────────────────────────────────────────────────────────────────
+ * PREMISE CORRECTION (truth-first)
+ * ─────────────────────────────────────────────────────────────────────────────
+ * The task asked to import "legal payload version configuration tables" and to
+ * prove that parsing an "alternative version stamp (e.g. v0.0.0-beta) safely
+ * falls back onto the active baseline regulatory text configuration".
+ *
+ * There is NO version-stamped legal table and NO version-parsing/fallback code
+ * anywhere in the repo. The only real, exported legal configuration surface is
+ * `KAI_LEGAL_DOCUMENTS` in `hushh-webapp/lib/legal/kai-legal-content.ts`, a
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4286, opened 2026-07-03
test(services): characterize haptic toggle multi-event debounce invariants

abdulbeast4-creator:
## Summary

Pins down structure stability across high-frequency `hapticFeedback` preference
mutations inside the public `SettingsService` (`hushh-webapp/lib/services/settings-service.ts`).

## Premise Correction (truth-first)

The originating request referenced "haptic **debounce** helpers" and "pointer
references out of sync." Neither exists in this repo:

- There is **no** debounce/throttle/timer logic for haptics anywhere.
- There are **no** "pointer references" — this is TypeScript/JS.

The real, exported surface is `SettingsService`, where `hapticFeedback` is a
plain `boolean` on `HushhSettings`, mutated via `updateSettings()` with a
synchronous last-write-wins shallow merge, persistence through Capacitor
`Preferences`, cache update, and subscriber notification. These tests
characterize that actual contract for rapid, back-to-back dispatches.

## Technical Context & Devex Attachment Plan

- **Canonical Attachment Point:** Added under the isolated `__tests__/lib/services/`
  runtime layer, alongside the existing `settings-service` tests.
- **Capability Overlap Review:** Exercises rapid mutation and subscriber
  convergence strictly through the exported service singleton; zero overlap with
  physical UI component render files.
- **Risk Mitigation:** Self-contained test file creation with zero production
  dependencies modified.

## Characterized invariants

- Baseline `hapticFeedback` default is `true`.
- Awaited back-to-back toggles converge to the last write (last-write-wins).
- Non-awaited rapid dispatches settle to a coherent final value (no torn state).
- Subscribers observe every dispatch and end on the final value; unsubscribe stops delivery.
- The listener payload is reference-identical to the cached `getSettings()` result (no desync).
- Rapid toggling mutates only `hapticFeedback`, leaving sibling settings intact.
- Each awaited update returns the fully merged, JSON-serializable state.

## Verification Results

- Validated locally: **1 file passed, 7 tests passed** via the repo vitest engine.

## Deviations from the original script

- **Filename extension:** requested `.spec.ts`, but the webapp vitest `include`
  glob is `__tests__/**/*.test.{ts,tsx}`. A `.spec.ts` file collects **0 tests**.
  Used `.test.ts` so the suite actually runs and gates.

## CI Gate Checklist

- [x] DCO Signed Verification

### DIFF
--- a/hushh-webapp/__tests__/lib/services/haptic-debounce-invariants.test.ts
+++ b/hushh-webapp/__tests__/lib/services/haptic-debounce-invariants.test.ts
@@ -1,0 +1,170 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+/**
+ * Characterization: haptic preference multi-event update invariants.
+ *
+ * ─────────────────────────────────────────────────────────────────────────────
+ * PREMISE CORRECTION (truth-first)
+ * ─────────────────────────────────────────────────────────────────────────────
+ * The task framing ("haptic *debounce* helpers", "pointer references out of
+ * sync") does NOT match the repo. There is:
+ *   - NO debounce/throttle/timer logic for haptics anywhere.
+ *   - NO "pointer references" — this is TypeScript/JS, not manual memory.
+ *
+ * The real, exported surface is `SettingsService`
+ * (`hushh-webapp/lib/services/settings-service.ts`). `hapticFeedback` is a plain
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4288, opened 2026-07-03
test(services): characterize kyc workflow artifact nested config merge

abdulbeast4-creator:
## Summary
Adds characterization tests for the public nested-configuration merge surface used by the KYC workflow PKM writer: `mergeKycWorkflowArtifact`. This is the load-bearing pure function that reconciles an incoming nested config map against an existing one while preserving unmentioned sibling properties.

## Truth-First Notes (surface & naming)
- The task was framed as testing a generic "preference deep-merge / storage sync adapter". The repo does **not** expose a standalone `updatePreference` / `deepMerge` utility:
  - `PersonalKnowledgeModelService.deepMergeRecords` is `private static` (not importable / not a public contract).
  - `KaiProfileService` preference updates are async, vault- and coordinator-bound class methods, not pure importable mergers.
  - The honest, public, pure merge is `mergeKycWorkflowArtifact` in `hushh-webapp/lib/services/kyc-pkm-write-service.ts`; these tests characterize its **actual** field-specific rules rather than an invented "deep-merge-everything" contract.
- The task requested a `.spec.ts` file, but `hushh-webapp/vitest.config.ts` only collects `*.test.ts` / `*.test.tsx`. A `.spec.ts` file would be silently skipped by the runner and CI, so this uses `.test.ts` to keep the "verified" claim true.

## Characterized behavior (from source, not assumption)
- Per-check merge: incoming wins unless its status is `not_started`, in which case the existing check is preserved.
- `counterparty` / `request_summary`: `incoming ?? existing ?? null`.
- `overall_status`, `pending_requirements`, `completed_requirements`: taken verbatim from the incoming artifact (replace semantics).
- `sent_replies`: shallow dictionary union (existing ∪ incoming), incoming keys overriding, dropped to `undefined` when empty.
- `null` existing map → pure adoption of the incoming config.
- No in-place mutation of the existing artifact.

## Verification Results
- `npx vitest run __tests__/lib/services/preference-deep-merge.test.ts` → **7 passed (7)**.

## CI Gate Checklist
- [x] DCO Signed
- [x] Test-Only Surface Isolation (single new `*.test.ts` file; no source changes)

### DIFF
--- a/hushh-webapp/__tests__/lib/services/preference-deep-merge.test.ts
+++ b/hushh-webapp/__tests__/lib/services/preference-deep-merge.test.ts
@@ -1,0 +1,285 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  buildKycWorkflowArtifact,
+  mergeKycWorkflowArtifact,
+  type KycWorkflowArtifact,
+  type KycWorkflowCheck,
+} from "@/lib/services/kyc-pkm-write-service";
+
+/**
+ * Characterization specs for the public nested-configuration merge surface.
+ *
+ * TRUTH-FIRST NOTE ON SURFACE SELECTION
+ * -------------------------------------
+ * The task framed this as a "public preference updater" that deep-merges a
+ * partial nested dictionary against an existing configuration map. The repo
+ * does NOT expose a standalone `updatePreference`/`deepMerge` utility:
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4290, opened 2026-07-03
test(services): characterize local storage sync adapter concurrency resiliency

abdulbeast4-creator:
## Summary
Adds characterization tests for the public, dependency-injectable storage sync helpers in `lib/agent/agent-voice-settings.ts` (`readAgentVoiceSettings` / `writeAgentVoiceSettings`) under overlapping, interleaved write pressure that simulates rapid pipeline write events.

## Truth-First Scope Corrections
- **Surface.** There is no separate "storage sync adapter object" module. The genuinely exported public storage sync helpers with a `storage?: Storage | null` injection seam are `readAgentVoiceSettings` / `writeAgentVoiceSettings`. Those are what these tests exercise via an in-memory mock `Storage` (no jsdom/browser dependency).
- **"Thread execution faults."** JavaScript is single-threaded and the Web Storage API is synchronous, so there is no real thread primitive to fault. The honest, useful invariants characterized are: overlapping microtask-scheduled writes never throw, always return a normalized in-range value, converge to a consistent last-writer-wins state, recover to the default voice on corrupted raw payloads, and remain contained (per-write isolation) when the injected adapter's `setItem` intermittently faults with a quota error.
- **File extension.** The task requested `*.spec.ts`, but `hushh-webapp/vitest.config.ts` only collects `*.test.ts` / `*.test.tsx`. A `.spec.ts` file would be silently skipped by the runner and CI, so this ships as `storage-concurrency-resiliency.test.ts` to keep the verification claim honest.

## Verification Results
- Local gate did **not** pass in this environment. `npx vitest run` fails during the shared `__tests__/setup.ts` with `Error: Vitest failed to find the runner`, before any test in this file executes.
- Confirmed environment-wide, not caused by this change: a pre-existing known-good test (`__tests__/lib/services/error-sanitizer.test.ts`) fails identically. `vitest` and `@vitest/runner` are aligned (4.1.5) on Node 24.15.0; this is a local runner/setup incompatibility.
- Authoritative verification is deferred to CI's clean install.

## CI Gate Checklist
- [x] DCO Signed Verification
- [ ] Local vitest run (blocked by pre-existing environment-wide runner failure; deferred to CI)

### DIFF
--- a/hushh-webapp/__tests__/lib/services/storage-concurrency-resiliency.test.ts
+++ b/hushh-webapp/__tests__/lib/services/storage-concurrency-resiliency.test.ts
@@ -1,0 +1,165 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  DEFAULT_AGENT_GEMINI_TTS_VOICE,
+  readAgentVoiceSettings,
+  writeAgentVoiceSettings,
+  type AgentVoiceSettings,
+} from "@/lib/agent/agent-voice-settings";
+
+/**
+ * Characterization specs for the public, dependency-injectable storage sync
+ * utilities in lib/agent/agent-voice-settings.ts under overlapping/interleaved
+ * write pressure.
+ *
+ * TRUTH-FIRST NOTE ON SURFACE SELECTION
+ * The task framed this as testing "the public storage sync utilities" for
+ * "asynchronous, overlapping test calls simulating rapid pipeline write events"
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current blockers: mock_only_test_without_production_contract, existing_capability_overlap_requires_review.

### Path To Merge

### Proof Needed

PR #4318, opened 2026-07-06
test(consent): characterize zero-tolerance token expiry (no clock-skew grace)

abdulbeast4-creator:
## Summary

Adds a **characterization test** pinning the real clock-skew behavior of the
consent-token expiry check in `hushh_mcp/consent/token.py`.

### Truth-first correction

The proposed premise was that `validate_token` allows a **60-second grace
window** for server clock desync. **That grace window does not exist.** The
expiry check is a strict, zero-tolerance cutoff:

```python
if int(time.time() * 1000) >= int(expires_at_str):
    return False, "Token expired", None
```

A token is valid up to the last millisecond before `expires_at` and rejected the
instant wall-clock time reaches it. The only `60`-based grace constant in the
module is `_BoundedRevocationCache._EXPIRED_TOKEN_GRACE_MS = 60 * 60 * 1000`
(one **hour**), which governs eviction of *revocation* markers from the in-memory
cache and never widens the expiry window.

Rather than build a non-existent skew tolerance, this PR **locks the actual
contract in place** so any future introduction of genuine skew tolerance is a
deliberate, reviewed change instead of silent drift.

## What it covers

- valid well before expiry
- valid ~1s before expiry (window is inclusive of every instant before expiry)
- rejected 1ms past expiry
- **no 60s grace**: rejected 30s past expiry
- rejected 59s past expiry (not an off-by-one edge)
- expiry checked **before** scope (expired token reports `Token expired`, never leaks scope)

## Test evidence

```
6 passed in 0.84s
```

No production code changed — test-only, additive.

### DIFF
--- a/consent-protocol/tests/integrations/test_clock_skew.py
+++ b/consent-protocol/tests/integrations/test_clock_skew.py
@@ -1,0 +1,121 @@
+# tests/integrations/test_clock_skew.py
+"""Characterization tests: how token validation behaves around server clock skew.
+
+TRUTH-FIRST NOTE (verified against hushh_mcp/consent/token.py):
+
+The originally proposed premise for these tests was a "60-second grace window"
+for clock desynchronization in the token evaluation pipeline. That grace window
+DOES NOT EXIST in `validate_token`. The expiry check is a strict, zero-tolerance
+cutoff:
+
+    if int(time.time() * 1000) >= int(expires_at_str):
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Backend

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4321, opened 2026-07-06
test(analytics): characterize log sanitizer payload filtering data contracts

abdulbeast4-creator:
## Summary

Adds a **characterization test** pinning the real recursive payload-filtering
contract of the consent-protocol PII sanitizer
(`hushh_mcp/consent/pii_sanitizer.py` — `sanitize_payload` / `_sanitize_value`).

### Truth-first correction

The proposed premise described key-based stripping where unconsented keys like
`raw_input` and `session_cache` are **stripped or masked by key name**. **The
sanitizer does not do that.** The verified contract is:

- `sanitize_payload` **never removes keys** — it returns a copy with the same
  key set. There is no allowlist/denylist that drops structural nodes.
- Masking is **content-based**, not key-name based, for arbitrary keys. Only keys
  in `_FORMAT_MASK_KEYS` / `_TEXT_MASK_KEYS` get forced masking; every other
  string still routes through the email/phone maskers, so embedded PII is masked
  while the surrounding structure is preserved verbatim.
- `raw_input` / `session_cache` are in **no** mask set, so their non-PII content
  survives unchanged; PII *inside* those values is still masked because every
  string passes through `sanitize_log_value`.

Rather than assert a non-existent key-stripping behavior, this PR locks the true
boundary. If the project later wants real key-based denylisting of `raw_input` /
`session_cache`, that is a **new** capability to add behind an explicit denylist
in `pii_sanitizer.py` — not something the sanitizer silently does today.

## What it covers (6 tests)

- key set preserved + known PII key masked + non-masked keys and scalars untouched
- embedded email PII masked even under an unconsented (`raw_input`) key
- recursion into nested dicts (PII masked, adjacent public node unmutated)
- recursion into lists of dicts (PII masked, adjacent public node unmutated)
- non-mutating contract (input dict unchanged)
- `_TEXT_MASK_KEYS` force-mask identity keys even without an email/phone pattern

## Test evidence

```
6 passed in 0.25s
```

No production code changed — test-only, additive.

## CI gate checklist

- [x] DCO Verified (`-s` sign-off)

### DIFF
--- a/consent-protocol/tests/integrations/test_sanitizer.py
+++ b/consent-protocol/tests/integrations/test_sanitizer.py
@@ -1,0 +1,106 @@
+# consent-protocol/tests/integrations/test_sanitizer.py
+"""
+Characterization tests for the recursive PII payload sanitizer.
+
+Source of truth: hushh_mcp/consent/pii_sanitizer.py
+    - sanitize_payload  (recursive, non-mutating dict traversal)
+    - _sanitize_value   (per-key masking dispatch)
+
+TRUTH-FIRST NOTE
+----------------
+The proposed premise said "unconsented keys ('raw_input', 'session_cache') are
+stripped or masked" by key name. That is NOT how this sanitizer works, and these
+tests pin the real contract:
+
+- `sanitize_payload` NEVER removes keys. It returns a copy with the SAME key set;
+  no key allowlist/denylist strips structural nodes.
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Backend

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4329, opened 2026-07-06
fix(ui): remove debate stream cancel run console warning leak

Roopmann30:
### Description

Removes a redundant `console.warn` leak in the `DebateStreamView` run cancellation handler. If the background run cancellation API fails when the user closes the dialog (e.g. if the run had already finished), the UI already gracefully proceeds to close and clear its state anyway. Removing this log keeps the client console clean without needing environment checks, and the catch variable is appropriately prefixed with an underscore to satisfy TypeScript linters.

## 📌 Impact Map (Required)

- Routes touched:
  - [x] `components/kai/debate-stream-view.tsx`

- Trust boundary touched:
  - [x] none (purely client UI logging removal)

## ✅ Review-Ready Contract

- [x] Title, body, changed files, and tests describe the same contract.
- [x] This PR changes a current reachable app/backend/package/docs surface.
- [x] Required checks are current on this head, commits are signed off, and GitHub shows this branch is mergeable with `main`.

## 🛑 Tri-Flow Architecture Check

- [x] **Web**: Next.js implementation (`components/kai/debate-stream-view.tsx`)
- [ ] **iOS**: N/A (Web UI only)
- [ ] **Android**: N/A (Web UI only)

## 🧪 Testing

- [x] Tested on Web (Code inspection, TypeScript linters)
- [x] Commits are signed off (`git commit -s`)

## 🛡️ Privacy & Consent

- [x] Sensitive surface touched: none (internal logging only, non-trust boundary)
- [x] Error path, rollback, or safe-failure behavior proved: N/A - purely a logging chore fix.

PR #4331, opened 2026-07-06
fix(ui): remove debate stream profile context load console warning leak

Roopmann30:
### Description

Removes a redundant `console.warn` leak in the `DebateStreamView` profile context load handler. If the local database read for the user's Kai preferences fails (e.g., if it doesn't exist yet), the application already gracefully degrades by continuing the agent debate with a default context. Removing this log keeps the client console clean without needing environment checks, and the catch variable is appropriately prefixed with an underscore to satisfy TypeScript linters.

### DIFF
--- a/hushh-webapp/components/kai/debate-stream-view.tsx
+++ b/hushh-webapp/components/kai/debate-stream-view.tsx
@@ -1328,8 +1328,7 @@ export function DebateStreamView({
             },
             financial_profile: profile,
           };
-        } catch (profileError) {
-          console.warn("[DebateStreamView] Failed to load Kai profile context:", profileError);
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):
Approved (safe-merge lane): diff-safe benign change — removes a non-actionable console.warn on a path-flagged surface, no trust-boundary/scope change. CI Status Gate green, MERGEABLE, contributor-pushed (verified head 13de4caf717eeed962996cac14aa4faaedffd484). Deferred from this run's merge queue only because it edits the same file (debate-stream-view.tsx) as #4329, which is queued first (one-PR-per-file collision guard); will enqueue on the next pass once #4329 lands.

PR #4304, opened 2026-07-04
Desktop Beta 1: Runtime Foundation & On-Device Kai Agent

parthmawai:
## Summary

Desktop **Beta 1**: production-grade runtime foundation (dynamic ports, process supervisor, precompiled backend binary, runtime context, logging, packaging) plus the first **on-device Kai agent** — chat running entirely on the Snapdragon NPU via Qualcomm GenieX, no cloud round-trip.

> **⚠️ Stacked behind #4163 (Desktop Alpha).** This branch (`feature/desktop-beta-v1.0`) is built directly on top of the alpha branch (`feature/windows-desktop`), which isn't merged yet. Since `integration/pr-train` doesn't have the alpha foundation, this diff necessarily includes alpha's ~1160 files again on top of the real beta delta — there's no way to show a smaller diff until #4163 merges. **Please review only the 3 commits below; everything else is already under review in #4163.** 

## New commits in this PR (review these)
```
78b713a43  fix(frontend): resolve ts errors in profile page for ai controller
ba72fac23  feat(desktop): finalize beta v1.0 — bundle GenieX runtime, packaging fixes, docs
254bc4f1f  fix(desktop): address Copilot review findings on beta PR
```
23 files, +1251/-288 — see `git diff 39b9da6a7..feature/desktop-beta-v1.0` (39b9da6a7 = alpha's current tip / #4163's head).

## Scope (new work only)

**On-Device Kai Agent**
- GenieX/QAIRT NPU runtime bundled via `extraResources` — no separate CLI install needed
- Model lifecycle with bounded crash recovery (`services/models/registry.js`)
- Local-mode chat routing, Experimental-labeled UI, streaming animation
- Fixed: new-chat suggestions racing to the cloud model instead of respecting local mode

**Packaging**
- Fixed a stuck "Loading…" screen in packaged builds: Next.js standalone output omits `.next/static` + `public/`; added a postbuild step to copy them into place
- Removed a duplicate `hushh-backend.exe` entry in `build.files`

**Reliability (Copilot review findings from the fork-hosted iteration of this PR, already addressed)**
- `main.js`: `ensureBackendVenv()` was gated to prod, where it no-ops — now runs in dev so fresh checkouts get `.venv` via `uv sync`
- `launcher`: production now fails fast if the standalone build is missing, instead of silently falling back to `next dev`
- `registry.js`: `getStatus()` no longer shells out synchronously on every UI poll (was blocking the main-process event loop) — cached, invalidated on install/remove
[...]

PR #4341, opened 2026-07-07
Desktop Beta 1.2: Hermes-compatible local bridge + reliable on-device reply model

parthmawai:
## Summary

Stacked on Beta 1.0 (#4304). Two related pieces of work for on-device (local) chat mode:

- **Local model bridge** (`backend/local_bridge/`) — a small FastAPI service sitting between GenieX and any OpenAI-compatible client, fixing two gaps GenieX has natively: unreliable streaming `finish_reason`/`[DONE]` compliance, and no real tool-calling (translates via the Nous/Hermes `<tool_call>` tag convention). This is the piece that makes on-device inference speak the same protocol Hermes Agent / Open WebUI expect from a local backend.
- **Hybrid on-device model split** — local chat now runs two purpose-built models instead of one generalist: a `Llama-3.2-1B` classifier handles tool-calling/action-plan detection, while `Qwen3-4B-Instruct-2507` (run through llama.cpp/GGUF instead of the original QAIRT-specific path) handles reply generation. Landing on this exact pairing took several reverted attempts — see commit messages and `docs/KNOWN_ISSUES.md` for the full trail (Qwen3.5-2B was tried and reverted after its always-on `<think>` reasoning trace proved ~40% unreliable in live testing, each failure costing 4-5 minutes).

## Before / after

The original QAIRT-backed local setup, under the RAM conditions it actually shipped with, could take **~3 minutes per response at ~1 tok/s** while holding **several GB of RAM** reserved regardless of whether it was actively generating. This PR's setup:

| | Before (QAIRT, real-world) | After (this PR) |
|---|---|---|
| Throughput | ~1 tok/s under pressure | ~7-10 tok/s, live-tested |
| Typical response time | ~3 minutes | 16-30s |
| Free RAM required to spawn | up to 6GB gate historically | 200MB |
| Idle RAM reserved | ~3.5-5GB pinned permanently | none — pin removed entirely |

Throughput is still below QAIRT's own rated speed for this model (23.1 tok/s) — both runtimes use the device's CPU/GPU/NPU together (confirmed via live Task Manager observation: NPU utilization sits at ~72% during generation, dropping to 0% when idle), so the gap isn't a CPU-vs-NPU difference; the actual cause (quantization format, GenieX's generic GGUF path vs. its QAIRT-specific one, or something else) isn't confirmed. The RAM and reliability gains are real, measured improvements regardless.

https://github.com/user-attachments/assets/1525ea65-25f2-45ac-a978-7c877a732beb

## What changed

- `local_bridge/server.py`, `local_bridge/tool_calling.py` — new bridge service (streaming fix + tool-calling translation).
- `agent_chat_service.py` — local-mode reply generation and action-plan classification split across the two models; `max_tokens` → `max_completion_tokens` fix (GenieX silently ignores the former).
- `registry.js` — provisions/verifies both hybrid-role models; RAM gate dropped from a scaled guess to a directly-measured floor (200MB, down from 1.5GB) based on a real memory-pressure/throughput curve showing no degradation cliff for this model down to 190MB free; removed a working-set pin that was reserving ~3.5GB of idle memory for a degradation mode this model doesn't have.
- `docs/KNOWN_ISSUES.md` — documents the full model-selection history, the measured RAM curves, and known open gaps (below).

## Known limitations (not fixed in this PR)

- **Speed tradeoff vs. QAIRT's rated speed**: ~7-10 tok/s here vs. QAIRT's 23.1 tok/s rating (~16 tok/s real-world best case). Traded deliberately for reliability (100% vs ~40%) and RAM behavior (see table above) — live-tested and judged acceptable on feel. Root cause of the gap not confirmed (not a CPU-vs-NPU difference — both paths use the device's full CPU/GPU/NPU compute).
- **Mixed intent messages drop the non-action part**: if a user combines an action request (e.g. "add this to my PKM") with a question in the same message, the reply-generation call never runs once an action plan resolves — the question is silently dropped with no indication to the user.
- **PKM auto-capture can silently fail to extract anything durable** even from a message that clearly contains save-worthy content — that extraction logic is a separate, pre-existing Gemini-backed classifier (`pkm_agent_lab_service.py`), unrelated to this PR's changes.
- Hermes Agent's own hard 64K-token minimum context requirement means this on-device setup can't act as a full drop-in Hermes model backend — documented as an intentional narrow-delegate role instead (see `KNOWN_ISSUES.md`).

## Test plan

- [x] Local model bridge: verified streaming `finish_reason`/`[DONE]` compliance and tool-call translation via direct calls.
- [x] Reply generation: live-tested in the running Electron app (plain chat, two multi-step financial-math prompts) — correctly defers per the math guardrail both times, no leaked reasoning text.
- [x] Action-plan classifier: live-tested PKM-add and confirmed end-to-end save flow.
- [x] RAM behavior: measured free-RAM/throughput curve (controlled memory-pressure allocator + real tok/s measurement) from 2.1GB down to 190MB free — no degradation found; verified app spawns and runs cleanly with the new lower gate.
- [ ] Not yet tested: combined classifier+reply-model resident RAM footprint under pressure (only reply model was curve-tested standalone).


kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4347, opened 2026-07-07
fix(one-location): recover from expired vault token on app resume (no dead-end banner)

DamriaNeelesh:
## Description

Fixes the iOS/Android/web resume failure where One Location could show a dead-end **Token validation failed.** banner after the app returned from the background.

## Root Cause

The backend deliberately returns the same generic `401` message for several VAULT_OWNER validation failures, so the screenshot alone cannot distinguish expiry, revocation, signature, or scope rejection.

Two client gaps made that rejection terminal:

1. `ApiService` processed HCT validation failures on web only; the native Capacitor HTTP path returned before that recovery logic.
2. The existing classifier did not recognize the backend's exact generic `Token validation failed.` response. The page-level fallback covered only the initial state load and depended on message matching.

Closing and reopening worked because vault bootstrap issued or recovered a valid token in a fresh app session.

## Fix

1. `VaultProvider` watches native `appStateChange` and web `visibilitychange` and proactively re-locks when a retained memory-only token has reached its known expiry.
2. A shared HCT auth-failure handler now runs for both web `fetch` and native `CapacitorHttp` responses.
3. The handler recognizes the backend's exact generic validation response and emits one `vault-lock-requested` event.
4. `VaultProvider` consumes that event, clears the in-memory vault key/token/expiry, and `VaultLockGuard` presents the standard Face ID/passphrase re-unlock flow.
5. One Location suppresses the rejected-token `401` banner while that secure recovery transition occurs; unrelated errors remain visible.
6. Expiry uses `now >= expiresAt`, matching the backend validity boundary.

No token or vault key is persisted, and no auth or consent boundary is weakened.

## Impact Map

- Routes touched: `/one/location`
- Shared runtime touched: platform-aware API auth-failure handling and vault lifecycle
- API/schema/data changes: none
- Cache/PKM changes: none; existing pause/invalidation paths run during re-lock
- Mobile parity: the same response handler now covers web, iOS, and Android
- Rollback: revert the three PR commits
- Error fallback: standard vault re-unlock sheet; unrelated failures keep normal messaging

## Verification

- [x] 44 focused transport, One Location, and VaultProvider recovery tests
- [x] Simulated iOS native `401` emits exactly one vault-lock event
- [x] Web `401` follows the same event contract and does not attempt Firebase refresh for HCT
- [x] VaultProvider event consumption clears the memory-only key and token
- [x] Web visible resume re-locks an expired token and preserves a valid token
- [x] Native inactive event does not re-lock; active event does and listener cleanup is verified
- [x] One Location does not render the terminal validation banner during recovery
- [x] Exact expiry boundary (`now === expiresAt`) fails closed
- [x] 53 related API client, One Location service, and chat tests
- [x] TypeScript typecheck
- [x] ESLint on all changed runtime and test files
- [x] Capacitor static parity
- [x] Capacitor production build
- [x] Branch freshness against `origin/main`
- [x] DCO sign-offs

## Known Unrelated Baseline

`verify:capacitor:plugins` reports the same pre-existing Android `HushhLocation` method drift on unchanged `main`; this PR does not modify that plugin contract.

Physical iOS lock/resume remains a post-merge/UAT device smoke test because this Windows environment cannot run an iOS simulator.

### DIFF
--- a/hushh-webapp/__tests__/components/one-location-agent-page.test.tsx
+++ b/hushh-webapp/__tests__/components/one-location-agent-page.test.tsx
@@ -576,6 +576,28 @@ describe("OneLocationAgentPage", () => {
     expect(mockSyncCurrentUser).toHaveBeenCalledWith({ uid: "user_a" });
   });
 
+  it("requests vault re-unlock instead of showing a stale-token error", async () => {
+    const lockReasons: string[] = [];
+    const handleLockRequest = (event: Event) => {
+      lockReasons.push(
+        (event as CustomEvent<{ reason?: string }>).detail?.reason ?? "",
+      );
+    };
+    window.addEventListener("vault-lock-requested", handleLockRequest);
+    mockGetState.mockRejectedValue(new Error("Token validation failed."));
+
+    try {
+      render(<OneLocationAgentPage />);
+
+      await waitFor(() => {
+        expect(lockReasons).toContain("one_location_token_invalid");
+      });
+      expect(screen.queryByText(/Token validation failed/i)).toBeNull();
+    } finally {
+      window.removeEventListener("vault-lock-requested", handleLockRequest);
+    }
+  });
+
   it("scrolls Active shares only when more than three shares are present", async () => {
     const baseGrant = locationState().ownerGrants[0]!;
     mockGetState.mockResolvedValueOnce({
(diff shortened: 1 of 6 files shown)


kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4353, opened 2026-07-08
test(navigation): characterize normalizeInternalRouteHref for 3+ leading slashes

abdulbeast4-creator:
## Summary

Adds a characterization test that pins the current behavior of `normalizeInternalRouteHref` (and its `resolveInternalRouteHref` fallback consequence) for hrefs with **three or more consecutive leading slashes** (e.g. `///app/dashboard`).

This is a docs-through-tests change only — **no production code is modified**.

## Current truth (verified against `hushh-webapp/lib/navigation/routes.ts`)

```ts
export function normalizeInternalRouteHref(value) {
  if (!href) return null;
  return href;
}
```

The `startsWith("//")` guard is a **prefix** check, so any run of two-or-more leading
slashes — which necessarily includes `///`, `////`, `/////` — is rejected with `null`.
There is **no** slash-collapsing path: `///app/dashboard` is never rewritten to
`/app/dashboard`. The adjacent `normalize-multi-slash.test.ts` documents the same guard
from the generic "multi-slash" angle; this file pins the specific named
`///app/dashboard` case and the "collapse does not happen" contract.

## What this test covers

- 3+ leading slashes return `null` (not the verbatim input).
- The validator does **not** strip/collapse the leading run into a clean single-slash leaf.
- Rejection holds after whitespace trimming and regardless of query/fragment payload.
- Control: a single leading slash with internal separators is preserved verbatim.
- `resolveInternalRouteHref` surfaces the caller-supplied base fallback instead of the unsafe input.

## Verification note

Local `vitest run` on this repo currently fails to boot the runner from
`__tests__/setup.ts` under the installed `vite@8.0.10` / `vitest@4.1.5` combo — this
reproduces on **pre-existing untouched** navigation tests (e.g.
`normalize-multi-slash.test.ts`), so it is an environment issue, not something this PR
introduces. CI installs from the lockfile and is unaffected. Assertions were verified
directly against the source implementation quoted above.

## Scope

- 1 file added: `hushh-webapp/__tests__/navigation/normalize-multiple-leading-slashes.test.ts`
- No source/behavior changes.

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-multiple-leading-slashes.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-multiple-leading-slashes.test.ts
@@ -1,0 +1,82 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  normalizeInternalRouteHref,
+  resolveInternalRouteHref,
+} from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref in
+// lib/navigation/routes.ts, focused on inputs with THREE OR MORE consecutive
+// leading forward slashes (e.g. "///app/dashboard").
+//
+// Truth-first note on the real implementation (routes.ts):
+//
+//   export function normalizeInternalRouteHref(value) {
+//     const href = String(value ?? "").trim();
+//     if (!href) return null;
+//     if (!href.startsWith("/") || href.startsWith("//")) return null;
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `9907e8e7`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4356, opened 2026-07-08
test(navigation): characterize normalizeInternalRouteHref unicode and emoji route bounds

abdulbeast4-creator:
## Summary\nEstablishes interface stability boundaries when routing names contain complex multi-byte text configurations (non-ASCII Unicode, emoji, and percent-encoded octets).\n\nThis is a characterization test only for the existing `normalizeInternalRouteHref` in `hushh-webapp/lib/navigation/routes.ts`. It pins the current, verified behavior: valid single-leading-slash hrefs are returned byte-for-byte after trim (no Unicode normalization, no percent-decoding), while `//`-prefixed and CR/LF-containing inputs are rejected.\n\n## CI gate checklist\n- [x] PR Base Policy — targets integration/pr-train\n- [x] DCO Verified\n- [x] Test Only Surface Change — zero production runtime risk

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-unicode-emoji.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-unicode-emoji.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  normalizeInternalRouteHref,
+  resolveInternalRouteHref,
+} from "@/lib/navigation/routes";
+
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `71e6484d`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4357, opened 2026-07-08
test(navigation): characterize normalizeInternalRouteHref control character injection bounds

abdulbeast4-creator:
## Summary\nCharacterization test only for the existing \`normalizeInternalRouteHref\`/\`resolveInternalRouteHref\` in \`hushh-webapp/lib/navigation/routes.ts\`. Pins current, verified behavior when raw href inputs carry hidden ASCII control bytes or NULL bytes.\n\n## Truth-first note on real behavior\nThe validator only special-cases CR/LF (\`/[\\r\\n]/\`). Every other C0 control byte is ordinary content, so:\n- \`/dashboard\\x00/settings\` -> returned **verbatim** (no sanitization; documents a real pass-through boundary, not a security guarantee).\n- \`\\x01/app\` / \`\\x00/dashboard\` -> **rejected** only incidentally, because \`.trim()\` does not strip these bytes and the string then fails the leading-slash guard.\n- CR/LF combined with a NULL/control byte -> still **rejected** via the explicit CR/LF guard.\n\nThis makes any future control-character hardening a deliberate, visible contract change rather than silent drift. Zero production runtime change.\n\n## CI gate checklist\n- [x] PR Base Policy — targets integration/pr-train\n- [x] DCO Verified — commit signed off (-s)\n- [x] Test Only Surface Change

### DIFF
--- a/hushh-webapp/__tests__/navigation/normalize-control-characters.test.ts
+++ b/hushh-webapp/__tests__/navigation/normalize-control-characters.test.ts
@@ -1,0 +1,89 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  normalizeInternalRouteHref,
+  resolveInternalRouteHref,
+} from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref in
+// lib/navigation/routes.ts, focused on route hrefs whose raw string content is
+// injected with hidden ASCII control characters or terminal NULL bytes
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current blockers: mock_only_test_without_production_contract.

### Path To Merge

### Proof Needed

PR #4358, opened 2026-07-08
test(utils): characterize formatCompleteJson on deeply nested circular references

abdulbeast4-creator:
## Summary

Adds a **characterization test only** for the existing `formatCompleteJson` in `hushh-webapp/lib/utils/json-to-human.ts`. It pins the current, verified behavior when the input object contains self-referential (circular) property graphs. **Zero production runtime change.**

The task premise implied `formatCompleteJson` performs deliberate cycle "decoupling" to avoid recursive stack exhaustion. I verified the source first — that is **not** how it works:

- `formatCompleteJson` is **not recursive** and has **no cycle-detection** (no `WeakSet`, no `[Circular]` marker).
- It walks a **fixed, bounded structure — at most three levels** (`section -> field -> nested field`).
- At the leaf, non-primitive values fall through to `String(value)` and render as the literal `"[object Object]"`.

So a cycle (`obj.self = obj`, or a parent/child back-reference) is **safe by bounded depth, not by intentional decoupling**. It can never trigger `RangeError: Maximum call stack size exceeded`, because the walk simply never descends into the cycle.

- **Direct self-reference** (`obj.self = obj`) -> does not throw; leaf renders as `"[object Object]"`.
- **Two-node parent/child back-reference** -> flattened at bounded leaf depth; no recursion.
- **Cycle routed through an array container** (`node.items = [node]`) -> bounded array header, no lockup.
- **Deterministic termination** -> finite, bounded output length for a self-cycle.
- **Sibling integrity** -> primitive siblings (e.g. `total_value`) still format normally alongside the cyclic branch.

The tests never call `JSON.stringify` on the cyclic objects (which *would* throw), so no exception blocks or loop lockups are involved. This makes any future cycle-handling change (real recursion, seen-guard, or `[Circular]` marker) a deliberate, visible contract change rather than silent drift.

## Proofs & Verification

- **Execution:** `npm test -- hushh-webapp/__tests__/utils/format-circular-references.test.ts`

## CI Gate Checklist

### DIFF
--- a/hushh-webapp/__tests__/utils/format-circular-references.test.ts
+++ b/hushh-webapp/__tests__/utils/format-circular-references.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson in lib/utils/json-to-human.ts,
+// focused on self-referential (circular) input objects.
+//
+// Truth-first note on the real implementation:
+//
+// formatCompleteJson is NOT recursive and has NO explicit cycle detection.
+// It walks a FIXED, bounded structure — at most three levels deep:
+//
+//   level 0: Object.entries(json)                 -> section
+//   level 1: Object.entries(sectionValue)         -> field
+//   level 2: Object.entries(value)                -> nested field
+//
+// At level 2, each nested value is rendered through formatValue(), which for a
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `977ad505`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4359, opened 2026-07-08
test(utils): characterize formatCompleteJson serialization behavior on native BigInt values

abdulbeast4-creator:
## Summary

Adds a **characterization test only** for the existing `formatCompleteJson` in `hushh-webapp/lib/utils/json-to-human.ts`, mapping how it handles native `BigInt` values in the input object. **Zero production runtime change.**

The task premise stated BigInt "natively triggers structural failures under raw `JSON.stringify`". That is true of `JSON.stringify` — but I verified the source first: **`formatCompleteJson` never calls `JSON.stringify`**. It builds a human-readable string via runtime type-dispatch, so BigInt is handled by branch selection, not JSON serialization. There is no `JSON.stringify` path here to fail, and no custom encoder is required.

Two distinct, verified paths exist depending on **where** the BigInt sits:

- **Top-level BigInt** (e.g. `{ supply: 123n }`): the main section loop only handles `number | string`, arrays, or `object`. A `bigint` matches **none** of these, so the section is **silently skipped** — no throw, no output line, never reaching `formatValue()`.
- **Nested BigInt** (inside a section object, e.g. `{ section: { supply: 123n } }`): reaches `formatValue()`, which returns early only for null/undefined/number/string/boolean; a bigint **falls through to `String(value)`**, rendering the plain decimal string (e.g. `"123"`) — no `n` suffix, no currency/percentage formatting, no throw.

- **Standalone top-level BigInt** (`9007199254740991n`) -> does not throw; **silently dropped** (empty output).
- **Nested BigInt** (`{ metrics: { tracker: 12345n } }`) -> renders `Tracker: 12345` via the `String()` fallback (no `n` suffix).
- **Large nested BigInt** (`9007199254740993n`, i.e. `2^53 + 1`) -> full precision preserved (no `Number` coercion loss).
- **Mixed section** (BigInt alongside a currency number) -> primitive siblings still format normally (`Total Portfolio Value: $1,000.00`) while the bigint uses the fallback.

This pins the **actual** contract — BigInt is neither smoothly converted at the top level (it is dropped) nor requires a custom encoder. Any future change (explicit BigInt support, a custom replacer, or a thrown error) becomes a deliberate, visible contract change rather than silent drift.

## Proofs & Verification

- **Execution:** `npm test -- hushh-webapp/__tests__/utils/format-bigint-serialization.test.ts`

## CI Gate Checklist

### DIFF
--- a/hushh-webapp/__tests__/utils/format-bigint-serialization.test.ts
+++ b/hushh-webapp/__tests__/utils/format-bigint-serialization.test.ts
@@ -1,0 +1,81 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson in lib/utils/json-to-human.ts,
+// focused on native BigInt values in the input object.
+//
+// Truth-first note on the real implementation:
+//
+// formatCompleteJson does NOT call JSON.stringify anywhere. The premise that
+// BigInt "natively triggers structural failures" only applies to raw
+// JSON.stringify (which throws "TypeError: Do not know how to serialize a
+// BigInt"). This helper builds a human-readable string by branching on runtime
+// type, so BigInt is handled by type-dispatch, not by JSON serialization.
+//
+// Two distinct, verified paths exist depending on WHERE the BigInt sits:
+//
+// 1) TOP-LEVEL bigint (e.g. { supply: 123n }):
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `adebd62c`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4361, opened 2026-07-08
test(utils): characterize formatCompleteJson depth bounds on highly nested objects

abdulbeast4-creator:
## Summary

Adds a **characterization test only** for the existing `formatCompleteJson` in `hushh-webapp/lib/utils/json-to-human.ts`, mapping how it behaves under deeply nested object/array structures. **Zero production runtime change.**

The task premise asked to "pin the maximum execution ceiling before exhausting structural stacks." I verified the source first, and the honest finding is: **there is no stack ceiling to exhaust — `formatCompleteJson` is not recursive.** It walks a **fixed, hard-coded maximum of three levels**:

- **Level 1:** `for (const [sectionKey, sectionValue] of Object.entries(json))`
- **Level 2:** `for (const [key, value] of Object.entries(sectionValue))`
- **Level 3:** the nested-object branch (`typeof value === "object" && !Array.isArray(value)`) -> `for (const [nestedKey, nestedValue] of ...)` printing `formatValue(nestedKey, nestedValue)`.

There is no self-call and no depth counter, so the "50-layers-deep stack overflow" framing does not apply. The real, bounded contract is:

- A **50-layer** (or even **5,000-layer**) nested object does **not** overflow the stack, throw, or hit a safety limit. Traversal runs in constant depth and simply **stops descending after level 3**.
- At level 3, an **object value** is passed to `formatValue()`, which has no object branch and falls through to `String(value)` === `"[object Object]"`. The 4th-and-deeper layers are never rendered — they collapse into that literal string.
- A **deeply nested array at level 2** is summarized by count (`<label>: N item(s)`), never descended into.

- **50-layer nested object** -> does not throw / no stack overflow.
- **5,000-layer nested object** -> still does not throw (proves non-recursive; a recursive walk would blow the stack here).
- **4+ levels deep** -> level-3 object value renders as `[object Object]`; the deep scalar (`42`) never appears.
- **Shallow depth-3 object** -> renders exactly the reachable levels (`--- Portfolio Summary ---`, `Beginning Value: $1,000.00`, `Total Income: $250.00`).
- **Deeply nested array at level 2** -> summarized as `Nested Series: 3 item(s)`; deep contents (`deepest`) never traversed.

This pins the **actual** bounded/non-recursive/depth-3 contract instead of a mythical recursion limit. If the walk is ever made recursive, these assertions flip — turning a silent behavioral change into a visible, reviewable one.

## Proofs & Verification

- **Execution:** `npm test -- hushh-webapp/__tests__/utils/format-depth-bounds.test.ts`

## CI Gate Checklist

### DIFF
--- a/hushh-webapp/__tests__/utils/format-depth-bounds.test.ts
+++ b/hushh-webapp/__tests__/utils/format-depth-bounds.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson in lib/utils/json-to-human.ts,
+// focused on deeply nested object/array structures.
+//
+// Truth-first note on the real implementation (verified against source):
+//
+// The task premise asked to "pin the maximum execution ceiling before
+// exhausting structural stacks." The honest finding after reading the code is
+// that THERE IS NO STACK CEILING TO EXHAUST — formatCompleteJson is NOT
+// recursive. It walks a FIXED, HARD-CODED MAXIMUM OF THREE LEVELS:
+//
+//   Level 1: `for (const [sectionKey, sectionValue] of Object.entries(json))`
+//   Level 2: `for (const [key, value] of Object.entries(sectionValue))`
+//   Level 3: the nested-object branch
+//            `if (typeof value === "object" && !Array.isArray(value))` ->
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `fda832af`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4362, opened 2026-07-08
test(utils): characterize local environment key resolution fallbacks

abdulbeast4-creator:
## Summary

Adds a **characterization test only** over the real local environment key-resolution path in `hushh-webapp/lib/testing/local-reviewer-auth.ts` (`resolveLocalReviewerCredentials` + its private `sanitizeConfiguredValue` gate). It pins how the resolver treats **default configuration markup tags** vs. **completely empty parameters**, and proves the verification layer drops **unhydrated credentials safely without throwing**. **Zero production runtime change.**

I verified the source before writing assertions. The premise ("verification layers drop unhydrated credentials safely without throwing system crashes") is **correct** — and the honest, precise contract is narrower than a generic "validation" framing:

- `sanitizeConfiguredValue()` collapses **two distinct unhydrated families** to `""`:
  1. **Empty / whitespace-only** values (`String(value || "").trim()` -> `""`).
  2. **Template placeholder markup**: `replace_with_...` and `your_..._here` (regex-matched, case-insensitive).
- When **either** credential resolves to `""`, `resolveLocalReviewerCredentials()` returns **`null`** — never throws. The failure mode for unhydrated input is a safe `null`, not a crash.
- A **second, independent gate**: even with fully hydrated credentials, a **non-local hostname** (anything outside `{localhost, 127.0.0.1}`) also returns `null`. An empty/undefined hostname **skips** that gate.
- Successful resolution returns **trimmed** `{ email, password }`.

There is no exception-based path here, so the tests assert `null`/`toEqual` and an explicit `.not.toThrow()` rather than a mythical throw-on-missing-config contract.

- Both empty -> `null`, and `resolveLocalReviewerCredentials("localhost")` does **not** throw.
- Both unset (`undefined`) -> `null`.
- `your_..._here` placeholder markup on both -> `null` (dropped as unhydrated).
- `replace_with_` placeholder markup on both -> `null` (dropped as unhydrated).
- Partial config (one real, one placeholder) -> `null`.
- Both hydrated on a local host (`127.0.0.1`) -> resolves **trimmed** `{ email, password }`.
- Both hydrated on a **non-local** host -> `null` (host gate).
- Both hydrated with **no hostname** -> resolves (host gate skipped).

Env vars are snapshotted in `beforeEach` and restored in `afterEach`, so the suite is hermetic and leaves no `process.env` residue.

## Proofs & Verification

- **Execution:** `npm --prefix hushh-webapp test -- utils/env-key-resolution.test.ts`
- **Workspace Isolation:** verified single new file via `git status --short`.

## CI Gate Checklist

### DIFF
--- a/hushh-webapp/__tests__/utils/env-key-resolution.test.ts
+++ b/hushh-webapp/__tests__/utils/env-key-resolution.test.ts
@@ -1,0 +1,115 @@
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import { resolveLocalReviewerCredentials } from "@/lib/testing/local-reviewer-auth";
+
+// Characterization tests for the local environment key-resolution path in
+// lib/testing/local-reviewer-auth.ts, focused on how the resolver treats
+// placeholder configuration markup vs. completely empty parameters.
+//
+// Truth-first note on the real implementation (verified against source):
+//
+//   resolveLocalReviewerCredentials() reads two env vars —
+//   NEXT_PUBLIC_LOCAL_REVIEWER_EMAIL and NEXT_PUBLIC_LOCAL_REVIEWER_PASSWORD —
+//   through a private sanitizeConfiguredValue() gate:
+//
+//     const trimmed = String(value || "").trim();
+//     if (!trimmed) return "";
+//     if (/replace_with_/i.test(trimmed)) return "";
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `949a966e`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4364, opened 2026-07-08
test(webapp): characterize normalizeInternalRouteHref empty/comma-blank query params

abdulbeast4-creator:
## Summary

Adds a characterization test suite that pins the **actual** behavior of
`normalizeInternalRouteHref` (`hushh-webapp/lib/navigation/routes.ts`) for
internal hrefs carrying empty-string and multi-comma query values, e.g.
`/search?tags=,,&category=`.

This is a **test-only, no behavior change** PR.

## Truth-first note on the premise

The routine is **not** a query-string parser. It performs no parsing, no
array-element extraction, no dropping of empty positions, and no construction
of blank literal array strings. Its real contract is a pure allow/deny gate
over the trimmed string:

- `String(value ?? "").trim()`, then
- reject empty, non-rooted (`!startsWith("/")`), protocol-relative (`//`),
  or CR/LF-containing input → returns `null`;
- otherwise returns the href **verbatim**, so the entire query block
  (every `,,`, every dangling `=`) survives byte-for-byte.

The tests pin exactly this contract on the empty/comma-blank query surface.

## Coverage

- Verbatim preservation of `tags=,,&category=`, repeated empty params,
  bracketed empty-array syntax, and leading empty comma positions.
- Outer-whitespace trimming while keeping blank query intact.
- `null` for protocol-relative host, non-rooted href, and newline injection.
- `null` for null/undefined/empty/whitespace input.

## Testing

Local `vitest` cannot register suites on this Windows + Node 24 box (a
runner/rolldown environment issue reproducible with a bare `expect(1+1)`
smoke test), so verification defers to CI, which runs the same `vitest` on
Node 24 where the existing webapp suite is green. Assertions were verified
line-by-line against the current `routes.ts` source.

### DIFF
--- a/hushh-webapp/__tests__/navigation/route-param-empty-arrays.test.ts
+++ b/hushh-webapp/__tests__/navigation/route-param-empty-arrays.test.ts
@@ -1,0 +1,100 @@
+import { describe, expect, it } from "vitest";
+
+import { normalizeInternalRouteHref } from "@/lib/navigation/routes";
+
+// Characterization tests for normalizeInternalRouteHref
+// (hushh-webapp/lib/navigation/routes.ts) focused on query-string components
+// that resolve to empty strings or multi-comma blanks within search
+// dimensions, e.g. `/search?tags=,,&category=`.
+//
+// TRUTH-FIRST — LITERAL CONTRACT:
+//
+//   export function normalizeInternalRouteHref(value): string | null {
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (changes requested):

### Direction

### Blocker
Current blockers: none.

### Path To Merge

### Proof Needed

PR #4372, opened 2026-07-08
fix(ui): remove agent chat workspace voice TTS failure console warning leak

Roopmann30:
### Description

Removes a redundant `console.warn` leak in the `AgentChatWorkspace` voice TTS failure handler. If the text-to-speech audio playback fails (e.g., due to autoplay policies or stream interruptions), the UI already gracefully degrades by pushing an error message to the user while keeping the text chat fully functional. Removing this log keeps the client console clean without needing environment checks, and the callback parameter is appropriately prefixed with an underscore to satisfy TypeScript linters.


kushaltrivedi5 (approved):

Re-reviewed at head `2e9aa14a`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4397, opened 2026-07-10
fix(one-location): stop legacy UI + raw SQL leak on recipient-key schema drift

### LINKED ISSUE #4365: Bug: Build 43 Location screen shows token validation failed
## Bug
In mobile build 43, the One Location screen shows `Token validation failed.` and the location readiness check appears stuck behind the error state.

## Evidence from screenshot
- Build: 43
- Surface: Mobile app > Location
- Visible screen copy: `Your circle, safely connected.`
- Error banner: `Token validation failed.`
- The UI also shows `Checking location readiness` beneath the agent input area.

## Current truth / classification
- `needs_verification`: Exact failing token type and backend endpoint are not confirmed from logs yet.
- `partially_exists`: Location screen renders, tabs render, and refresh control is visible.
- `missing`: Build 43 does not recover cleanly from token validation failure on the Location screen.

## Expected behavior
- Build 43 should validate the active user/session token before loading location-sharing state.
- The Location screen should not show a persistent token validation failure for a signed-in user.
- Refresh should retry the correct auth/location readiness path and either recover or show a specific actionable message.

## Suggested investigation boundary
Check the mobile build 43 auth/session token path for Location, including the request used by refresh and readiness checks. Verify whether the failure is caused by stale token handling, missing bearer token, wrong audience/scope, expired session refresh, or a backend validation mismatch.

Do not create a parallel Location auth flow. Fix the shared token validation/session refresh path used by the Location surface.

## Acceptance criteria
- On build 43, opening Location with a valid signed-in session does not show `Token validation failed.`
- Pulling or tapping refresh revalidates/retries without leaving the screen stuck in readiness loading.
- If the token is genuinely expired, the app prompts a session refresh/re-auth instead of a generic error banner.
- Add or update a regression test or QA note covering Location load with valid token and expired/stale token paths.

DamriaNeelesh:
## Summary

Fixes the `/one/location` page falling back to the **retired legacy UI** (and dumping a raw database error into the page) whenever a load error occurred.

### Root cause (one problem, three symptoms)
- **The error**: recipient-key registration (runs on every location-page load) inserts a column added by **migration `083`**. Any environment where 083 has not been applied raises an undefined-column error, which sets `loadError`.
- **The old UI reappearing**: the page used `if (USE_LOCATION_REDESIGN && !loadError)`, so ANY load error (that column error, or a token-validation error when the vault token was missing or expired) fell through to the entire retired legacy page.
- **Raw error leak**: the error helper returned `error.message` verbatim, exposing backend internals in consumer UI (a vault and PKM data-boundary violation).

## Changes

**Frontend — `hushh-webapp/app/one/location/page.tsx`**
- The redesign hub is now the **only** customer UI; it renders in success, loading, AND error states with a small inline banner. The legacy block is unreachable at runtime, so it can never reappear on any error or trigger.
- Added `isSafeUserFacingMessage` so raw backend internals can never surface in consumer UI.

**Backend — `consent-protocol/hushh_mcp/services/one_location_agent_service.py`**
- Narrow, bounded self-heal: on exactly the missing-column drift, apply the idempotent `ADD COLUMN IF NOT EXISTS` (mirroring migration 083 and the existing additive backstops in `db/migrate.py`) once and retry. Any other DB error propagates unchanged.

**Tests — `consent-protocol/tests/services/test_one_location_agent_service.py`**
- Regression coverage for the narrow drift detector, the self-heal retry path, and unrelated-error propagation.

## Verification
- Backend: `pytest tests/services/test_one_location_agent_service.py` -> 31 passed.
- Frontend: `tsc --noEmit` clean; `eslint app/one/location/page.tsx --max-warnings=0` clean.
- DB: reproduced the drift on a local Postgres, applied migration 083, and confirmed the real recipient-key registration succeeds.

## Operator note
The canonical fix for any already-drifted environment remains `cd consent-protocol && python db/migrate.py --release`. The backend self-heal is a safety net so the feature does not hard-fail when an environment is a migration behind.


DamriaNeelesh:
### Relationship to issue #4365 (partial)

This PR partially addresses #4365 (`Token validation failed.` on the Location screen). It removes the **legacy-UI regression** that made that issue look worse:

- The retired legacy Location page no longer renders on any load error (redesign hub always renders).
- `Token validation failed.` now surfaces as a small inline banner on the current UI instead of dropping to the old page with a stuck readiness check.

It does **not** fix the root cause of #4365 — why a signed-in user's VAULT_OWNER token fails validation at `require_vault_owner_token` (`consent-protocol/api/middleware.py`) on build 43, nor the expired-token -> re-auth prompt. That is a separate IAM/consent change that needs the build-43 auth logs (issue is still `needs_verification`) and should not be bolted onto this focused fix. Tracking that as a follow-up. Full breakdown posted on #4365.

PR #4400, opened 2026-07-10
test(utils): characterize native symbol lookup and value serialization invariants

abdulbeast4-creator:
### Summary

Adds a **test-only** characterization suite that pins the exact runtime behavior of `formatCompleteJson` (`hushh-webapp/lib/utils/json-to-human.ts`) when native JavaScript `Symbol` values or symbol-keyed properties are present in the input object.

No production code is changed. This locks down two language-level invariants so future refactors of the formatter cannot silently regress them.

### Why this matters

`formatCompleteJson` renders Gemini-parsed document JSON into human-readable text. It builds output by iterating with `Object.entries(...)` and routing scalar values through a `formatValue` `typeof` cascade — it does **not** use `JSON.stringify`. There was no explicit test documenting how `Symbol` inputs behave through this path, even though a sibling suite (`format-json-primitives.test.ts`) already characterizes booleans and other primitives. This PR closes that documentation gap.

### Literal contract characterized (verified against source)

1. **Symbol keys are dropped by enumeration rules.**
   `Object.entries` returns only enumerable, own, **string**-keyed properties. Symbol-keyed properties are excluded by the ECMAScript spec, not by any formatter branch — so a symbol-keyed entry contributes no output line. An object whose only keys are symbols formats to `""`.

2. **Symbol values stringify via the `String()` fallback without throwing.**
   Inside a named object section, a value that is not null/undefined, not an object, and not an array reaches `formatValue`, whose final branch is `return String(value)`. `String(Symbol("AAPL"))` yields `"Symbol(AAPL)"` and does **not** throw (unlike the `${sym}` template form). A **top-level** symbol value matches no branch in the top-level scalar guard (`typeof === "number" | "string"` only) and is silently skipped, producing `""`.

### Tests added

`hushh-webapp/__tests__/utils/formatter-symbol-invariants.test.ts` (5 cases):
- drops a symbol-keyed property while still emitting sibling string keys
- emits `""` for an object whose only keys are symbols
- skips a top-level symbol value (`""`)
- renders a symbol value on a mapped key inside an object as `Security: Symbol(AAPL)`
- does not throw for symbol values nested in an object section

### Proof / Verification

- **Scope:** `git status` showed exactly one added file, nothing else modified. Diff is `1 file changed, 81 insertions(+)`.
- **DCO:** commit created with `git commit -s` (Signed-off-by trailer present).
- **Source cross-check:** assertions were derived directly from the current `json-to-human.ts` (`Object.entries` at the section loop; `String(value)` fallback in `formatValue`).
- **Local test run caveat (honest disclosure):** the Vitest runner in my local environment aborts with `Vitest failed to find the runner` before loading any spec (this reproduces even against already-merged suites), so I could **not** produce a local green run. CI is the authoritative gate for the pass/fail signal here.

### CI Gate Checklist
- [x] DCO Verified (`-s`)
- [x] Test Only Surface Change (single new spec, no production edits)

### DIFF
--- a/hushh-webapp/__tests__/utils/formatter-symbol-invariants.test.ts
+++ b/hushh-webapp/__tests__/utils/formatter-symbol-invariants.test.ts
@@ -1,0 +1,81 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for formatCompleteJson
+// (hushh-webapp/lib/utils/json-to-human.ts) focused on the EXACT runtime
+// invariants for native JavaScript `Symbol` inputs.
+//
+// Truth-first, verified against the source:
+//   - formatCompleteJson iterates with `Object.entries(json)` (line 233) and
+//     hand-builds output lines. It NEVER calls JSON.stringify. `Object.entries`
+//     only returns enumerable, own, STRING-keyed properties — symbol keys are
+//     excluded by the language spec, not by any formatter branch. So a
+//     symbol-keyed property contributes no output line.
+//   - A symbol VALUE attached to a mapped string key routes through
+//     formatValue, which is a typeof-cascade (number/string/boolean) whose final
+//     fallback is `return String(value)` (line 202). `String(Symbol("x"))`
+//     yields "Symbol(x)" and does NOT throw, unlike the `${sym}` template form.
+// These tests pin those two invariants precisely. They document existing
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `653bf6fa`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4427, opened 2026-07-12
test(utils): characterize formatCompleteJson deep nested key iteration exclusion limits

abdulbeast4-creator:
### Summary

Adds a **test-only** characterization suite pinning how `formatCompleteJson`
(`hushh-webapp/lib/utils/json-to-human.ts`) enumerates keys of nested object
structures — specifically that inherited (prototype-chain) and non-enumerable
properties are excluded via `Object.entries`, and where that traversal stops.

No production code is changed.

### Truth-first corrections to the premise

The task described a **"recursive line-builder"** that omits inherited keys
**"across all node depths."** Both framings are inaccurate against the current
source; the tests characterize the real behavior instead of the assumed one:

1. **`formatCompleteJson` is not recursive.** It performs a **fixed manual
   descent of exactly three `Object.entries` passes**:
   - level 1 — `Object.entries(json)` (root, ~line 233)
   - level 2 — `Object.entries(sectionValue)` (section object, ~line 333)
   - level 3 — `Object.entries(value)` (nested object, ~line 339)

   A **level-4** value is not descended: it is handed to `formatValue`, whose
   object fall-through returns `String(value)` === `"[object Object]"` (~line 202).
   So there is a **hard depth ceiling at level 3**; "all node depths" is false.

2. **The inherited / non-enumerable exclusion is real, but it is a consequence
   of `Object.entries`** (own-enumerable-string keys only) at the three levels
   actually iterated — not of any recursion.

### Literal contract characterized (verified against source)

- Level 2 & 3 objects: inherited (prototype) keys → omitted; own non-enumerable
  keys → omitted; own enumerable keys → rendered.
- Level 4 object: rendered wholesale as `"[object Object]"`; its own **and**
  inherited keys never surface as individual lines.

### Tests added

`hushh-webapp/__tests__/utils/formatter-nested-exclusion.test.ts` (5 cases):
inherited-key exclusion at level 2, non-enumerable exclusion at level 2,
combined inherited+non-enumerable exclusion at level 3, non-recursive level-4
`"[object Object]"` ceiling, and level-4 inherited-key irrelevance.

### Proofs & Verification

- **Scope:** single added file; `1 file changed, 139 insertions(+)`; tree
  otherwise clean.
- **DCO:** commit created with `git commit -s` (Signed-off-by trailer present).
- **Source cross-check:** every assertion derived directly from the current
  `json-to-human.ts` (`formatCompleteJson` three-level descent + `formatValue`
  object fall-through).
- **Local test run caveat (honest disclosure):** the Vitest runner in my
  environment aborts with `Vitest failed to find the runner` before loading any
  spec (reproduces against already-merged suites too), so I could **not**
  produce a local green run. CI is the authoritative gate.

### CI Gate Checklist
- [x] DCO Verified (`-s`)
- [x] Test Only Surface Change (single new spec, no production edits)

### DIFF
--- a/hushh-webapp/__tests__/utils/formatter-nested-exclusion.test.ts
+++ b/hushh-webapp/__tests__/utils/formatter-nested-exclusion.test.ts
@@ -1,0 +1,139 @@
+import { describe, expect, it } from "vitest";
+
+import { formatCompleteJson } from "@/lib/utils/json-to-human";
+
+// Characterization tests for how `formatCompleteJson`
+// (hushh-webapp/lib/utils/json-to-human.ts) enumerates keys of nested object
+// structures — specifically that inherited (prototype-chain) and non-enumerable
+// properties are excluded. Test-only; no production change.
+//
+// TRUTH-FIRST CORRECTION TO THE TASK PREMISE
+// ------------------------------------------
+// The task described a "recursive line-builder" that omits inherited keys
+// "across all node depths." That is NOT how this function works. Two corrections,
+// both verified against the current source:
+//
+// 1) `formatCompleteJson` is NOT recursive. It performs a FIXED manual descent of
+//    exactly THREE `Object.entries` passes:
+//       - level 1: `Object.entries(json)`                       (root, line 233)
+//       - level 2: `Object.entries(sectionValue)`               (section, line 333)
+//       - level 3: `Object.entries(value)`                      (nested, line 339)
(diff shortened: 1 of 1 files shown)


kushaltrivedi5 (approved):

Re-reviewed at head `fd15282b`: frontend change, path-clean, green CI Status Gate, MERGEABLE, base integration/pr-train. Enqueued to merge queue.

PR #4500, opened 2026-07-14
fix(one-location): viewer-side recomputing ETA for pick-me-up (no more "ETA unavailable")

ahujagautam024:
## Problem
In the One Location **pick-me-up** flow, the requester's en-route card frequently showed **"ETA unavailable"** — especially after the helper's app refreshed. Root cause: the ETA was computed on the *helper's* side and shipped inside the encrypted payload; the helper's drive session is in-memory, so a refresh dropped `drive.etaSeconds` and the requester's ETA went null.

## Fix (frontend-only)
Move ETA computation to the **requester's (viewer's) side**. The requester already receives the helper's live coordinates continuously, and knows their own pickup point — so ETA is recomputed locally from *(helper's live position → my pickup point)* via the existing server `routeEta` proxy. It never falls back to "ETA unavailable" once tracking starts; a failed refresh keeps the last-known value, and before the first result it shows a soft "ETA updating…".

### What changed
- `lib/one-location/eta-recompute.ts` — shared throttle helper `shouldRecomputeEta` (60s / 250m); `page.tsx` now imports it (DRY).
- `use-pickup-eta.ts` — viewer-side recompute hook: seed → throttle → server `routeEta` → keep-last-known on failure; throttle advances on every attempt (no retry storm during an outage).
- `pickup-live-route-map.tsx` — dual-location map (helper + pickup) reusing `DriveRouteMap` (route + ETA badge); degrades to single-point when pickup unknown.
- `pickup-enroute-card-live.tsx` — container feeding one recomputed ETA into both the card header and the map badge.
- `pickup-enroute.ts` — sources the requester's pickup point via an injected `pickupPointForOwnerGrant` resolver (backed by `pickupSessionRef`/`myLocationPoint`), replacing an always-null `decryptedPoints` lookup.
- Wired into `location-redesign-hub.tsx`, replacing the old single-point preview.

### Out of scope (unchanged)
Helper-side drive-session store, the public token viewer, and the helper's own ETA badge.

## ETA source
Server `POST /api/one/location/maps/route-eta` only (already deployed) — never a client Google Maps call, since the iOS WKWebView degrades the Maps SDK to an iframe. **No backend changes.**

## Testing
- New/updated unit tests for the throttle helper, the hook (seed / recompute / keep-last-on-failure / throttle), the presentational map, the pickup-point derivation, and the container.
- 197/197 one-location tests pass; `tsc --noEmit` clean; `eslint --max-warnings=0` clean.
- Built + launched on the iOS simulator against UAT.
- Two-account manual E2E (login + vault are human-gated) to be run on UAT after deploy.

## Follow-ups (non-blocking)
- Failure-path throttle test (fail once + tiny move → no refetch).
- Seed `distanceMeters` is a `0` placeholder that can briefly reach the badge on a seed+first-fetch-fail path.

Spec: `docs/superpowers/specs/2026-07-14-pickup-live-route-map-design.md`
Plan: `docs/superpowers/plans/2026-07-14-pickup-live-route-map.md`


kushaltrivedi5 (changes requested):
## Changes Requested: Unknown

### Direction

### Blocker

### Path To Merge

### Proof Needed

PR #4622, opened 2026-07-20
fix(ria): no-flash RIA entry + faster flow (caching, iOS verify, backend)

ankitkumarsingh1702:
## Problem
Opening the RIA subagent when a profile already exists flashed the onboarding wizard (parked at "2/5 Enter your licence") before the profile loaded; the flow re-fetched + flashed "Checking" on every open; and iOS licence verify was slow/hung vs web.

## Fixes
**A — No-flash entry**
- Onboarding page decides onboarding-vs-profile from synchronously-cached signals via a latched `entryMode`: an established advisor renders a neutral skeleton (never the wizard) → redirects to the regulatory profile; a native Preferences hint resolves cold starts without a network wait.
- Stop recording `/ria/onboarding` as `lastRiaPath` (navbar + `routeForPersona`) so it never overrides `riaEntryRoute` (switch → RIA_HOME).
- Profile section renders a skeleton (not an empty body) while redirecting a setup user to onboarding. `?edit=license` / `?reinitiate=1` keep the wizard.

**B — Caching**
- Regulatory-status TTL SHORT→SESSION (matches persona-context); `readCachedOrFetch` serves stale-from-memory instantly + revalidates (SWR); profile workspace seeds synchronously and always revalidates on open (spinner only on a cold cache); licence-verify result cached client-side honoring the server TTL (explicit tap forces live); native Preferences cold-start persistence; device/native caches invalidated alongside memory on persona/marketplace change.

**C — iOS transport**
- Wire connect/read timeouts + an AbortSignal race into the native CapacitorHttp path so verify is bounded + cancellable; skip auto re-verify on reopen when a fresh cached verify exists.

**D — Backend (env-flagged)**
- `RIA_VERIFY_REUSE_PARTIAL_CACHE`: serve a cached found-without-location result instead of a full re-scrape (tagged for a location refresh).
- `RIA_VERIFY_PARALLEL_FALLBACKS`: fetch SEC/FINRA profile PDFs concurrently.

## Hardening (adversarial review — 2 confirmed findings fixed)
- Write-after-invalidate race: a per-user invalidation epoch (`lib/cache/ria-invalidation-epoch.ts`) drops a stale in-flight write-back (memory/device/native) after a delete/switch, so a deleted RIA can't repopulate `exists:true` and re-bounce.
- 30-min-stale profile: the regulatory profile always revalidates on open (seed still paints instantly) so compliance status can't lag the SESSION cache.
[...]

PR #4626, opened 2026-07-21
fix(ria): stop onboarding wizard flashing before the profile redirect

ankitkumarsingh1702:
## Problem

An already-established RIA advisor (`riaCapability === "switch"`) who lands on `/ria/onboarding` briefly sees the onboarding **welcome step flash** for ~1 frame before being redirected to their profile (`/one/profile`, regulatory panel). Reported as "first I see onboarding, then the profile screen."

## Root cause

The "established advisor → profile" decision is a **post-paint `useEffect`** gated on `!personaLoading && !personaRefreshing`. But the render path painted the branded `OnboardingShell` (welcome eyebrow/title/hero) **unconditionally** — `renderStep()` only gated on the local data-fetch `loading` flag, never on persona state. `riaCapability` starts `"disabled"` (persona `null`) and only becomes `"switch"` after an async `getPersonaState()` fetch, so the wizard paints in the gap between the local load resolving and the redirect firing. The redirect was correct — it just ran one render too late.

## Fix

Hold a **neutral loader** until the one-shot entry decision resolves, so the branded wizard never paints for a frame.

- **`app/ria/onboarding/page.tsx`** — new `entryResolved` state gates `OnboardingShell`. The entry effect now **redirects-or-reveals exactly once**:
  - established advisor (not in the embedded setup hub, no `?edit`/`?reinitiate`/`?step` intent) → `router.replace` to the profile and keep the loader up through navigation;
  - setup / edit-intent / setup-hub / IAM-unavailable / signed-out → reveal the wizard.
  - Tied to the existing one-shot `onboardingEntryHandledRef`, so a **post-submit capability flip to `"switch"` no longer re-hides the review step** (a naive capability-only guard would have regressed the rejected-submit path). The native test beacon reports `loading` / `redirect-valid` during the hold.
- **`app/ria/page.tsx`** — mirror guard: a not-onboarded user no longer flashes the RIA home hero before redirecting to onboarding. `disabled`/IAM advisors are not held (handled via the existing compatibility state), so nobody is trapped in the loader.

## Testing

- `npx vitest run __tests__/app/ria/onboarding-page.test.tsx` → **24/24 pass**, including new/strengthened assertions:
  - established (`switch`) advisor: the wizard chrome (`onboarding-shell` / `step-welcome`) **never paints** before the redirect;
  - a neutral loader is held while persona state is still resolving;
  - setup and `?edit=license` advisors still reach the wizard.
- `npm run typecheck` — clean.
- `npm run lint` (changed files, `--max-warnings=0`) — clean.

## Scope

Web/`main` only, by request. This file has diverged between `main` and the mobile/iOS branch, so the iOS build is unaffected by this PR and would need the fix applied there separately.

### DIFF
--- a/hushh-webapp/__tests__/app/ria/onboarding-page.test.tsx
+++ b/hushh-webapp/__tests__/app/ria/onboarding-page.test.tsx
@@ -353,6 +353,8 @@ describe("RiaOnboardingPage", () => {
     await waitFor(() => {
       expect(screen.getByTestId("step-welcome")).toBeTruthy();
     });
+    // A fresh (setup) advisor reaches the branded wizard, not just the loader.
+    expect(screen.getByTestId("onboarding-shell")).toBeTruthy();
     expect(screen.getByTestId("shell-eyebrow").textContent).toBe("Welcome");
     expect(screen.getByTestId("welcome-type").textContent).toBe("individual");
     expect(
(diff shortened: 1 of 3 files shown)


ankitkumarsingh1702:
Closing as redundant: the RIA onboarding→profile flash is already fixed on `main` by #4622 ("no-flash RIA entry" — a synchronous entryMode + persisted RiaOnboardingStatusLocalService hint), which supersedes this approach. The /ria home mirror guard from this branch is not on main; deferring it separately. Local branch retained.

PR #4730, opened 2026-07-30
feat(runtime): add Meta Muse Spark 1.1 as a BYOA runtime provider

msainani:
Adds Meta's Muse Spark 1.1 to the runtime provider registry, so it joins Gemini, Anthropic, OpenAI and Grok as a brain Agent One can run on.

## Why this was small

Meta shipped Muse Spark as a deliberately OpenAI-compatible surface. That makes it structurally identical to how Grok already works here — Grok speaks the OpenAI wire format on its own host and is configured purely by `base_url`. So this is a base-URL constant, a dispatch branch, and a registry entry. **No new transport.**

Documented capabilities are structured output and parallel tool calling, and it generalises to MCP servers natively, which fits the MCP-first posture rather than needing an adapter.

## Capability flags are deliberately conservative

Function calling is on because Meta documents it. **Prompt caching and native realtime are off**, because borrowing another vendor's wire format is not evidence of serving that vendor's realtime API — and `registry.py` already warns never to infer realtime from the provider. If those turn out to be supported, that is a measurement against a live key, not an assumption.

## The test that matters

Muse Spark reuses the OpenAI transport, so `base_url` is the **only** thing separating it from OpenAI. Getting it wrong would POST a user's Meta key to `api.openai.com` — a credential leak to a third party that no other test here would catch.

`test_factory_sends_muse_spark_to_metas_host_not_openais` asserts the transport receives Meta's host. It was verified by sabotage: changing the constant to OpenAI's host fails exactly that test and nothing else.

## Verified

- Full protocol lane green: ruff, mypy (92 source files), bandit, **420 tests**
- `tests/services/test_runtime_providers.py`: 41 passed (3 new)

One note for reviewers: that test file is **not** in the backend CI manifest, which is a deliberately small allowlist. These run locally and in the full suite, but not in the pull-request gate — so the green tick on this PR does not by itself cover them.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VMQerNkJv8zitj1bCVbXK9)_

### DIFF
--- a/consent-protocol/hushh_mcp/runtime_providers/factory.py
+++ b/consent-protocol/hushh_mcp/runtime_providers/factory.py
@@ -289,6 +289,12 @@ def _build(
         from .openai_transport import GROK_BASE_URL, OpenAITransport
 
         return OpenAITransport(api_key=api_key, base_url=GROK_BASE_URL, provider="grok")
+    if provider == "musespark":
+        from .openai_transport import MUSE_SPARK_BASE_URL, OpenAITransport
+
+        return OpenAITransport(
+            api_key=api_key, base_url=MUSE_SPARK_BASE_URL, provider="musespark"
+        )
     # normalize_provider already rejects unknown providers, so this is defensive.
     raise ValueError(f"Unsupported runtime provider: {provider!r}")
 
--- a/consent-protocol/hushh_mcp/runtime_providers/openai_transport.py
+++ b/consent-protocol/hushh_mcp/runtime_providers/openai_transport.py
@@ -1,8 +1,11 @@
 """Native OpenAI-compatible transport adapter.
 
-Serves both OpenAI (native realtime) and Grok/x.ai, which speaks the OpenAI
-wire format on its own host. Grok is configured via ``base_url`` so there is
-one code path for both.
+Serves OpenAI (native realtime) plus every provider that speaks the OpenAI wire
+format on its own host: Grok/x.ai and Meta's Muse Spark. Each is configured via
+``base_url`` so there is one code path for all of them.
+
+Adding an OpenAI-compatible provider is a base-URL constant here, a dispatch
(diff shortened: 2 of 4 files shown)


msainani:
Blocked on a base-branch decision that needs a maintainer — flagging rather than forcing it.

**`PR Base Policy` fails because this targets `main`.** Per `config/ci-governance.json`, only `integration/pr-train` and named `maintainer/promote-*` branches may target `main`. `Preflight Gate` and `CI Status Gate` are cascades from that one failure; everything independent passed — DCO, Secret Scan, Governance, Base Freshness, Upstream Sync.

**But this change cannot go through the train.** I tried retargeting to `integration/pr-train` and reverted it. The train is **1,544 commits behind `main`**, and `consent-protocol/hushh_mcp/runtime_providers/` **does not exist on it at all**. The cherry-pick fails `DU` (deleted by us / modified by them) because the module this change edits was added to `main` after the train diverged. That is not a resolvable conflict; there is nothing to patch.

Retargeting also inflated the diff to 2,616 files / 1,545 commits and `mergeable_state: dirty`, which is the train's lag, not this change. Base is back on `main` so the diff reads correctly: **3 files, one commit.**

**Two ways forward, both maintainer calls:**

1. Add a `maintainer/promote-*` head branch to `main_allowed_head_branches` and route this through it. I cannot self-authorize an entry in the governance policy.
2. Advance `integration/pr-train` to current `main`, after which this cherry-picks cleanly and goes through the normal train.

Option 2 looks like the real fix — a train 1,544 commits behind `main` will block every change touching anything added since it diverged, not just this one.

The change itself is unaffected and verified: full protocol lane green (ruff, mypy across 92 files, bandit, 420 tests), plus 41 in `test_runtime_providers.py`. Happy to do either once someone says which.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4746, opened 2026-07-30
fix(kai): durable coarse-checkpoint store for cross-instance debate resume

ankitkumarsingh1702:
## Summary

RIA **debate** resume works on UAT but 404s on prod. Root cause: `KaiAnalyzeRunManager` (`api/routes/kai/run_manager.py`) is a **per-process in-memory singleton**. On Cloud Run a reconnect can land on an instance that never created the run, so it returns `ANALYZE_RUN_NOT_FOUND` (session affinity is best-effort and lost on scale-down / cold-start / revision change). This was partially masked by the frontend inline-fallback in #4737; this PR addresses the server-side durability gap.

Adds a **Postgres-backed durable run store** that mirrors only **coarse checkpoints** (2 writes/run: `start` + `terminal`) — never per-token frames — so it cannot starve the pool pinned at `DB_SQLALCHEMY_MAX_OVERFLOW=0` (#4736). On a 404 miss the stream reads through to the store and, if the run already finished, replays its single terminal SSE frame (e.g. the DecisionCard) so the client completes instead of stalling; otherwise it falls through to the existing 404.

## What changed
- **New store** `hushh_mcp/services/kai_analyze_run_store.py` — `save_start_checkpoint` / `save_terminal_checkpoint` / `load_terminal_checkpoint`, all best-effort.
- **`run_manager.py`** — optional `on_checkpoint` sink (default `None`), fired once at start and once on any terminal frame; wrapped so a misbehaving sink can never interrupt streaming.
- **`stream.py`** — flag-gated persister + `_maybe_durable_terminal_replay` wired at both 404 sites.
- **Migration `125_kai_analyze_run_store.sql`** — additive, REPLAY-safe (`CREATE TABLE/INDEX IF NOT EXISTS`).
- **Manifest + contracts** — migration registered; `expected_migration_version` bumped 124→125 across dev/uat/prod contracts; `kai_analyze_runs` added to uat/prod contracts (dev_minimum is an intentional subset).
- **Tests** — new cross-instance round-trip suite + governance lockstep; added to the CI backend manifest.

## Safety / blast radius
- **Flag-gated `KAI_ANALYZE_DURABLE_RUN_STORE`, default OFF.** With the flag off no checkpoints are written and no read-through occurs → behavior is **byte-for-byte** the existing in-memory path. Prod deploys with the flag unset (empty substitution in `backend.cloudbuild.yaml`), so this ships **dormant**.
- **Best-effort everywhere** — the sink and every store call are wrapped; a store failure can never perturb the live stream.
- **No per-token writes** → pool-safe under the `MAX_OVERFLOW=0` pin.
- **Ownership** enforced by `user_id` on the read; **consent token is never persisted**.
- Migration is idempotent → safe under the prod inline `replay` migration mode.

## Testing
- `ruff check .` — pass
- `bandit -ll` — no issues
- Full CI manifest (`run-test-ci.sh`) — **427 passed** incl. the new 7-test suite
- `db_migration_release_guard.py` for all 3 contracts — `violations: []`, version lockstep 125

## Rollout plan
1. Merge → UAT (flag can be flipped ON on UAT to verify true multi-instance resume).
2. Prod receives the **dormant** code + migration 125 (applied inline). Prod behavior unchanged until the flag is deliberately flipped.

### DIFF
--- a/consent-protocol/api/routes/kai/run_manager.py
+++ b/consent-protocol/api/routes/kai/run_manager.py
@@ -7,6 +7,7 @@ disconnect/reconnect without losing an ongoing debate.
 from __future__ import annotations
 
 import asyncio
+import inspect
 import json
 import logging
 import uuid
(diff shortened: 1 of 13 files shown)


ankitkumarsingh1702:
Closing as a duplicate. A parallel session already shipped the identical fix in #4745 (merged 2026-07-30 22:26 UTC): same root cause (per-instance in-memory `KaiAnalyzeRunManager`), same coarse terminal-checkpoint approach, same `kai_analyze_runs` table + migration `125_kai_analyze_run_store.sql`, and the same flag `KAI_ANALYZE_DURABLE_RUN_STORE` (default OFF, dormant in prod). #4745 is the canonical implementation; this branch would only collide with it (duplicate migration 125, conflicting run_manager/stream edits) and main has already advanced to migration 126. No unique behavior here is worth preserving.

PR #4785, opened 2026-08-03
feat(one-location): restore Home/Work/Other save-place onboarding + draggable map pin picker

DamriaNeelesh:
## Summary

Restores the **Home / Work / Other "Save this place"** step in Location onboarding (it existed on `main` but was missing from `codex/location-onboarding-pixel-match`) and makes it appear **every time** the onboarding screen shows, plus adds an **Uber/Zomato-style drag-to-pin map picker** so users can set their exact home instead of trusting a coarse GPS fix.

## Task 1 — Save-place modal always triggers + editable in Settings

- Restored the feature files that were absent from this branch:
  - `components/one-location/onboarding/save-location-modal.tsx` — Home / Work / Other modal (+ custom label for "Other")
  - `lib/one-location/saved-locations.ts` — encrypted PKM store (Home/Work singletons, 25m de-dupe)
  - `components/one-location/saved-locations-section.tsx` — Settings list to view / add / edit / remove
  - `lib/one-location/location-control-state.ts`, `lib/one-location/use-location-control-state.ts` — control-state helpers
- Wired the modal into `app/one/location/page.tsx`: capture-once + reverse-geocode trigger, save/skip/search/select handlers, rendered inside the onboarding `BodyPortal`.
- **Always shows:** removed the old once-per-user localStorage suppression (those keys are now only *cleared*, never read to hide the prompt) and added an effect in `one-location-onboarding-flow.tsx` that fires the prompt the moment location access is granted — so a previously discarded onboarding still gets the save-place prompt.
- Rendered `<SavedLocationsSection/>` in the Location Settings/Privacy surface so a saved place is instantly viewable and updatable.

## Task 2 — Accurate location via a draggable map pin

- New `components/one-location/onboarding/location-picker-map.tsx`: a Google Map with a **fixed centre pin** the user drags the map beneath (+ tap-to-place), a **"Locate me"** recenter button, live reverse-geocoding on every settle, and **Confirm** returns the exact `{lat, lng, address}`.
- Added a **"Set exact location on map"** action to `SaveLocationModal`, used in both onboarding and the Settings add-flow, replacing the coarse GPS fix with the precise pinned coordinate before saving.
- Graceful fallback to address search when the Maps browser key is unavailable.

## Files changed

9 files, +2,358 lines (see diff).

## Testing

- `tsc --noEmit` → 0 errors
- Dev server: `/one/location` → HTTP 200
- Verified the modal (Home/Work/Other) and the drag-to-pin picker render in the browser.

## Notes

- Exact coordinates stay in the encrypted vault (never written to browser storage).
- The interactive map needs `NEXT_PUBLIC_GOOGLE_MAPS_BROWSER_API_KEY`; without it the picker degrades to address search (by design).

### DIFF
--- a/hushh-webapp/app/one/location/page.tsx
+++ b/hushh-webapp/app/one/location/page.tsx
@@ -55,7 +55,16 @@ import {
   OneLocationOnboardingFlow as OneLocationOnboardingExperience,
   type ConnectionRequestResult,
 } from "@/components/one-location/onboarding/one-location-onboarding-flow";
+import { SaveLocationModal } from "@/components/one-location/onboarding/save-location-modal";
+import type { PickedLocation } from "@/components/one-location/onboarding/location-picker-map";
+import {
+  addSavedLocation,
+  DuplicateSavedLocationError,
+  type SavedLocationCategory,
+} from "@/lib/one-location/saved-locations";
+
 import { useConsentNotificationState } from "@/components/consent/notification-provider";
+
 
 import { Badge } from "@/components/ui/badge";
 import { Button } from "@/components/ui/button";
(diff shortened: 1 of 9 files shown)


DamriaNeelesh:
Superseded by #4787, which is correctly branched off the latest main (this one targeted the codex/location-onboarding-pixel-match branch by mistake).

PR #4786, opened 2026-08-03
feat(one): Wallet Profile — shareable Hussh One pass for Apple Wallet

ankitkumarsingh1702:
## Wallet Profile — a shareable Hussh One pass for Apple Wallet

Adds an Apple Wallet pass whose QR opens a **public, owner-controlled profile page**. The scanner needs no app, no account, no Apple device, and no login — the primary use case is professional networking / a dynamic digital business card. Setup reuses existing profile info (<60s) and the owner gets progressive privacy controls plus preview / edit / pause / rotate-QR / revoke.

Ships **dark**: `ONE_WALLET_CARD_ENABLED` defaults OFF and no Pass Type ID signing material exists yet, so there is no user-visible change on merge.

### Why this architecture

- **Dedicated `one_wallet_cards` plane (migration 132)**, modelled on the shipped One Location public-invite pattern — a stable `pass_serial` plus a rotatable share token stored **only as a SHA-256 hash**.
- **Deliberately NOT built on `pkm_default_available_projections`:** that table is scanned by the Information Marketplace catalogue with no source filter, so publishing a card through it would silently list the owner for sale, and it mints a fresh handle on every edit (invalidating an already-printed QR).
- **No Wallet web service:** `webServiceURL`/`authenticationToken` are omitted; the QR resolves server-side, so edit/pause/revoke take effect instantly. Removes APNs, device registration, and five endpoints from scope.
- **No new native code, entitlement, or app release:** WKWebView cannot import a `.pkpass`, but Capacitor hands off-origin navigations to Safari, which can.

### Public surface hardening

- Explicit **per-route rate limiting** (the global limiter is never wired), forwarded-IP keying.
- `X-Robots-Tag: noindex` and the `/c` prefix added to the robots disallow list.
- Preview-as-visitor uses the **same backend projection builder** as real public access — what the owner previews is byte-identical to what a stranger sees.
- Public status codes: unknown → 404 generic; **paused → 404 (same generic body as unknown)**; revoked → 410 `{status:revoked}`; expired → 410 `{status:expired}`.
- `one_wallet_cards` is registered in **account deletion and reset**, with an ordering assertion: the row holds denormalized PII (name/email/phone/links) reachable from an unauthenticated endpoint, so it is deleted *before* `actor_profiles` tears down the identity spine.

### Safety / correctness

- Signing service returns **503 when material is absent**, so a flag-on-without-certs state degrades gracefully rather than 500ing.
- `card_payload` is a **closed allowlist** with per-field max lengths, https-only URLs, and rejection of `javascript:`/`data:`/userinfo.
- Also fixes a class of **brittle migration tests** that asserted their own migration was the manifest's last entry or pinned an exact schema version (one had been red on main for several migrations), and adds a **guard test** preventing reintroduction.

### Provisioning gate — state of verification

`scripts/ci/provision_wallet_pass_certificate.py` guards an **irreversible** action: minting a Pass Type ID certificate consumes one of a small, finite number of slots on the Apple account. Its pre-mint permission check was calling `gcloud projects test-iam-permissions`, **which does not exist** (SDK 565.0.0 returns exit 2, `Invalid choice`). The gate therefore rejected every credential with a misleading "grant `roles/secretmanager.admin`" — including a correct one — and had never actually run. This PR rewrites it against the Cloud Resource Manager `testIamPermissions` REST endpoint, and additionally:

- demands `secretmanager.secrets.create` **only when a target secret is absent** (`upsert_gcp_secret.py` skips creation otherwise, so a rotation-scoped credential was being over-rejected);
- names the project-vs-secret caveat in the failure text — a grant made on an individual secret is invisible to a project-level check;
- stops `stored_certificate_expiry` from exiting on an unparseable PEM, so the fail-closed correspondence predicate can reach its own verdict (`SystemExit` is a `BaseException` and was escaping the predicate);
- replaces post-mint partial-publish guidance that recommended `--force-new-certificate` — **both routes reach the mint and burn another Apple slot** — with a copy-from-a-healthy-project recovery procedure;
- gives secret reads the same 3× retry and stderr surfacing as writes, short-circuits on `NOT_FOUND`, and prevents a subprocess `OSError` from escaping the post-mint guard and skipping the "already minted" instruction.

**What is and is not verified.** The rewritten gate was exercised against the **live** Cloud Resource Manager API on both `hushh-pda-uat` and `hushh-pda` using the read-only `eval-log-reader` service account: the endpoint was reachable in both projects and the gate **rejected fail-closed** with the new message naming each missing permission. That establishes reachability, JSON parsing, and the reject path — behaviour the previous code could not reach at all. It does **not** establish the accept path: no credential holding `roles/secretmanager.admin` was available (the local user token fails reauth non-interactively), so the "permissions sufficient → proceed" branch remains unexercised against the real API and this defect should not be considered closed until a `--apply` dry run passes with an admin credential. The `--self-test` is a unit harness and is deliberately **not** offered as evidence that the gate works in production.

Guard changes were validated by **mutation testing** rather than assertion counting: eight deliberate regressions (dying loader inside the predicate, the non-existent subcommand, the re-mint recommendation, `write_secret` dying instead of reporting, both retry constants cut to 1, the `NOT_FOUND` short-circuit removed, stderr swallowed) are each caught. Two were detectable only after replacing a tautological assertion that compared a call count against the very constant it was meant to pin.

### Governance registration

A new table and new routes are not just code here — they have to be declared in a chain of governed artifacts, three of which are generated and `--check`ed. Running the `Governance` lane locally surfaced four blockers that CI would otherwise have revealed one at a time, since Preflight short-circuits the expensive lanes behind Governance.

- `runtime-db-data-plane-contract.json`: new `one_wallet_card` family, classified `personal_projection` (mirroring the owner-published `pkm_default_available_projection` family) and owned by `iam-consent-governance` — the consent/sharing plane, deliberately not `vault-pkm-governance`, because the card does not sit on PKM.
- `one-route-orchestration-index.v1.json` (both byte-identical copies) and `runtime-topology-index.v1.json`: regenerated to cover `/c/[token]` and `/one/wallet-card`. The generator derives `trust_boundary` from a route's gateway action ids, not from page authentication; neither route registers gateway actions, so `none` is correct by construction and matches the `/one/location/request/[token]` precedent. Page auth for `/one/wallet-card` is enforced by the route layout and OneAuthGate.
- Plan-doc H1 now reads "Hussh" — public prose says Hussh while route, API and schema identifiers keep the legacy `hushh` spelling, so the filename is unchanged. Fixed at the source rather than by widening the allowlist in `scripts/verify-doc-brand.cjs`.

`bash scripts/ci/orchestrate.sh governance` exits 0: 118 routes, 18 agents, 42 table families; 126/126 tables classified; 239 checks passed, 0 errors, 0 warnings; release manifest, repo head, UAT and prod contracts all at 132.

### Test coverage

178 new tests (113 Python + 65 TS): service layer, routes/auth/rate-limit, pass signing, migration, public page, QR encoding, payload allowlist, account-deletion coverage, plus the migration-brittleness guard. Six previously brittle migration tests were repaired and still pass. Additive, backward-compatible, reversible.

Test fixture share-token values are literal non-secrets (`wallet-card-test-token-not-a-real-secret`); the public lookup path applies no format, length, or charset validation, so the fixture value is arbitrary and no plaintext token, key PEM, or `card_payload` value is logged anywhere.

Signed-off-by: Ankit Kumar Singh <ankit@hushh.ai>

### DIFF
--- a/.github/workflows/provision-wallet-pass-certificate.yml
+++ b/.github/workflows/provision-wallet-pass-certificate.yml
@@ -1,0 +1,337 @@
+name: Provision Wallet Pass Certificate
+
+# One-click provisioning of the Apple Pass Type ID (`pass.com.hushh.app.one`) and
+# its PASS_TYPE_ID signing certificate for the Hushh One Wallet Profile pass.
+#
+# The whole chain is automatable with the Admin-role App Store Connect key that
+# already lives in GCP Secret Manager (hushh-pda-uat): create/reuse the Pass Type
+# ID, generate an RSA-2048 keypair + CSR inside the runner, request the
+# certificate, verify the Apple WWDR **G4** intermediate, and write
+# WALLET_PASS_CERT_PEM / WALLET_PASS_KEY_PEM / WALLET_PASS_WWDR_PEM to Secret
+# Manager over stdin. Cloud Run then receives them as `--set-secrets` env refs
+# from deploy/backend.cloudbuild.yaml and reads them through runtime_settings.
+#
+# IRREVERSIBLE: minting a certificate consumes a slot in the Apple Developer
+# account and cannot be undone from CI. This workflow therefore defaults to a dry
+# run and additionally requires the `confirm` phrase before it creates anything —
+# the same default-off gating the App Store release workflow uses for its
+# submit-for-review action.
+#
+# The private key is generated in the job, piped straight into Secret Manager,
+# and never printed, never exported to the environment, and never uploaded as an
+# artifact. Only non-sensitive metadata (identifier, certificate id, serial,
+# expiry) reaches the log, the artifact, and the job summary.
+#
+# ROTATION: a Pass Type ID certificate is valid for about a year (the WWDR G4
+# intermediate runs to 2030-12-10, so it is not the operational risk). Re-dispatch
+# this workflow before the expiry printed in the job summary; a re-run inside the
+# renewal window mints a replacement, and a re-run outside it is a safe no-op.
+#
+# Runbook: docs/superpowers/plans/2026-08-03-hushh-one-wallet-card.md (section 5)
+# Contract: docs/superpowers/specs/2026-08-03-wallet-card-contract.md (section 1)
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Resolve, generate and verify only — create no Apple resource and write no secret"
+        required: false
+        default: true
+        type: boolean
+      confirm:
+        description: "IRREVERSIBLE: type MINT-PASS-CERTIFICATE to actually mint the Apple certificate and write the secrets (required when dry_run=false)"
+        required: false
+        default: ""
+        type: string
+      secret_projects:
+        description: "Comma-separated GCP projects that receive the signing material. A certificate is minted ONCE, so list every environment that needs it in the same run."
+        required: false
+        default: "hushh-pda-uat"
+        type: string
+      force_new_certificate:
+        description: "Mint a new certificate even if valid signing material is already stored (rotation override)"
+        required: false
+        default: false
+        type: boolean
+      notes:
+        description: "Optional note for the run summary"
+        required: false
+        type: string
+
+permissions:
+  contents: read
+
+env:
+  # The App Store Connect Admin key (APPSTORE_CONNECT_API_KEY_P8_B64 / _KEY_ID /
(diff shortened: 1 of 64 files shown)


ankitkumarsingh1702:
Superseded by #4791 — same five commits, rebased cleanly onto `main` @ `2938661d5`. Opened on a new branch (`feat/one-wallet-card-v2`) instead of force-pushing this one, so this branch's history is left untouched. Closing.

PR #4887, opened 2026-08-05
feat(one): add the Nationwide insurance agent directory to Connect

### LINKED ISSUE #4905: Connect & Directory: E.164 dialer, Nationwide directory & auth notifications
Tracking issue for Connect phone dialer, Nationwide directory, and auth sign-in/connection emails.

ankitkumarsingh1702:
Adds a second location-anchored directory to Connect → **Around you**, behind a nested segment: **Advisors** (BrokerCheck, unchanged) and **Insurance** (Nationwide agency locator, new). Both hang off the same shared location bus, so switching between them never re-prompts a reader who already granted location.

## What the reader gets

A distance-ranked list of Nationwide agencies with what each one can insure, tapping through to address, products and a Call button. Same radius control, same paging, same empty and error states as the advisers list beside it.

## Verified against the live service, not its README

Two upstream behaviours shape the surface and neither is documented upstream:

- **A search row is already complete.** The locator returns full agency data inline, so there is no detail fetch — opening an agency costs no request.
- **Zero results arrive as HTTP 200.** Both for a place the locator does not cover (checked with `lat=0&lng=-30`) and for a ZIP it cannot parse (checked with `ZZZZZ`). Neither is an error, so neither is shown as one; both land on the empty state offering a US ZIP.

A third difference: the locator’s attribution carries **no terms or error-reporting URL**, unlike BrokerCheck’s. Both links are now rendered only when the source supplies them — a link with no `href` is not a weaker credit, it is a broken one.

## The credential

The key is mirrored from the project that owns it on every deploy, reusing the helper added for the advisor key in #4878 rather than a second copy of it.

Read access to that project is a **per-secret grant that does not come with any lane’s own project-level roles**. It was granted to all three deploy identities *before* anything bound the secret, additively (`add-iam-policy-binding`, never a policy replace). The home project’s own policy is unchanged and no secret versions were minted.

The same gap applied to the advisor key merged in #4878 — its mirror could not read the source and was silently inert. That grant is now in place too, so that fix does what its commit message says.

## Existing behaviour preserved

The presentational parts both directories share — location prompt, ZIP form, empty block, credit footer — moved into one module. The advisers surface keeps **every test id and string byte-identical**, and its 17 existing tests pass unchanged against the shared components.

## Validation

| Gate | Result |
|---|---|
| Frontend tests | 32 passed (17 advisors unchanged + 15 new insurance) |
| Backend tests | 56 passed across both directories |
| Regression sweep | 20 failures on this branch, **the same 20 on clean `origin/main`** — byte-identical sets, no new failures |
| `tsc --noEmit` | clean |
| `eslint --max-warnings=0` | clean |
| `ruff check` / `ruff format` | clean |
| `next build` | passed |

## Notes

- Wired for **UAT only**, matching the advisor directory’s existing precedent. Prod and dev pass no base URL and stay inert.
- The pre-commit/pre-push hook fails on four markdown files (`CONTRIBUTING.md`, three under `consent-protocol/docs/reference/`). It fails identically on clean `origin/main` and none are in this diff, so the hook was bypassed rather than widening scope into unrelated files.

18 files changed, no deletions or renames.

---
Closes #4905
(retroactive board-linkage backfill, 2026-08-06)


ankitkumarsingh1702 (commented):
good to go

PR #4915, opened 2026-08-06
feat(ci): Cloud Build fallback lane for CI and deploy

ankitkumarsingh1702:
## Why

GitHub-hosted runners are currently leaving `PR Validation` and `Main Post-Merge Smoke` queued for hours. `deploy-uat` and `deploy-production` gate on a check named exactly `Main Post-Merge Smoke Gate` with `REQUIRE_CI_SUCCESS=1`, so a **queued** smoke run fails the deploy SHA gate identically to a red one. Two UAT deploys failed today at "Validate deployment SHA against main" for this reason — the code was never the problem, the verdict just could not be scheduled.

Cloud Build is healthy throughout. This adds a fallback lane so CI and deploy remain reachable during a runner outage.

## What

**`deploy/ci.cloudbuild.yaml`** — runs CI stages via `scripts/ci/orchestrate.sh`, the same entrypoint Actions calls, so the checks cannot drift from CI. Stages: `smoke`, `governance`, `web-core`, `protocol`, `mcp-package`, `integration`, `secret`, `all`.

It clones the repo instead of using the uploaded source. The repo's `.gcloudignore` excludes `.git/`, `*.sh`, `*.md`, `docs/` and `tests/` because it is tuned for the Docker deploy builds — a CI run against that upload would be missing every `scripts/ci/*.sh` and have no git history for `merge-base`. The repo is public, so the clone needs no credentials.

Optionally publishes the check-run back to GitHub (`_PUBLISH_CHECK=true`) so a green fallback build actually closes the release gate.

**`scripts/ops/cloudbuild_release.sh`** — the deploy configs were always Cloud Build; what lives only inside Actions is the orchestration. This reproduces it: SHA ancestry check → authority preflight → predeploy state → scope resolution → secret sync → migrations via Cloud SQL proxy → build/deploy `--no-traffic` → runtime identity assert → traffic promotion → provenance/parity/health verification → rollback on failure.

## Deliberate boundaries

- **UAT stays unreachable.** `hushh-pda-uat` carries the `uat-deploy-authority-lock` deny policy that routes UAT deploys through the CI service account by design. The driver checks *effective* permissions up front (deny policies are invisible to `get-iam-policy`) and refuses before mutating anything, rather than failing halfway through a release. It is not routed around.
- **Fallback releases are labelled as such** — `_DEPLOY_SOURCE=cloudbuild-fallback-<env>` rather than `deploy-uat`/`deploy-production`, so a fallback release stays visible in `gcloud run revisions describe` and in provenance reports.
- **`--skip-ci-check` waives only the green-check requirement**, never the ancestry check. A SHA that is not an ancestor of `origin/main` is refused regardless.
- **PR-shaped checks are not simulated.** `dco-check`, `pr-base-policy`, `main-freshness-gate` and the GitHub security-alert half of `secret-scan` read the PR event payload; they are documented as gaps rather than faked into passing.

## Verification

- Production dry run exercises SHA resolution, authority preflight, predeploy capture, scope resolution, and prints the exact `gcloud builds submit` commands — substitutions match `deploy-production.yml` byte-for-byte apart from the intentional fallback labels.
- UAT run correctly refuses at the authority preflight before any mutation.
- Found and fixed a real bug in review: `resolve-deploy-scope.py --json` emits JSON booleans that print as Python `True`/`False`, which never matched the lowercase `"true"` shell comparison — the driver concluded "nothing to deploy" for every SHA.
- Found and fixed a Cloud Build substitution bug: uppercase shell locals (`${NODE_VERSION}`, `$PATH`) are resolved by Cloud Build before bash runs and fail the build, since no config here enables `ALLOW_LOOSE`. Corrected to lowercase locals and `$$VAR` for real env vars — the convention verified against the stored `Build` resources of successful UAT deploys.
- `consent-protocol/tests/test_cloudbuild_step_arg_limit.py` — 7 passed, with the new config registered in `CLOUDBUILD_CONFIGS`.
- `scripts/ci/runtime-contract-check.sh`, `verify-pr-governance-sections.py`, `verify-protected-pipeline-edits.py --self-test`, and `skill_lint.py` (47 skills) all pass.

**Not executed:** the CI Cloud Build has not been submitted to a live project — build submission was blocked by a local permission control in this session. It is statically validated (YAML parse, substitution audit against known-good controls, arg-limit test) but the first real run should be watched.

## Skill

`.claude/skills/cloudbuild-ci-fallback/SKILL.md` records how to tell an outage from a real failure, which lanes are reachable, the UAT authority boundary, and the substitution rule — so this is not rediscovered during the next outage.


ankitkumarsingh1702:
## Added: a single CI gate, and what running it found

Second commit set adds `scripts/ci/local-release-gate.sh` + the `ci-gate` skill — one command that runs every check Actions runs and emits a verdict keyed to the SHA.

`cloudbuild_release.sh --skip-ci-check` now **requires** a complete, passing gate report for the exact SHA. The flag waives the GitHub check, not verification. A `--fast` or single-stage run records `complete: false` and cannot authorize a deploy.

### Running it on this branch found three real things

1. **A violation in my own deploy driver.** `verify-runtime-config-contract.py` forbids the legacy frontend-origin key name anywhere in a tracked file, including as a local shell variable. Renamed to `FRONTEND_SERVICE_URL`/`BACKEND_SERVICE_URL`, which is also more accurate — those hold the Cloud Run service URL, not the public origin. It then caught the *skill file* for naming the key in prose; reworded.

2. **`main` is currently red on Governance.** `generate_runtime_topology_index.py` fails with `cache manifest route coverage differs from route layout: missing=['/ria/claim']` — the route landed in #4902 without a cache-manifest entry. Reproduced on a clean `origin/main` worktree, and confirmed against GitHub: the two most recent completed `PR Validation` runs both show `Governance: failure`. This is not from this PR. It needs a cache policy decision for `/ria/claim`, so I have not guessed at one.

3. **A false failure that would have wasted someone's afternoon.** `mcp-package` failed with `Unterminated string in JSON at position 8192`, which reads as a corrupt gateway manifest. It is a Node version skew: CI pins 24, local was 20, and Node 20's `child_process.exec` truncates captured stdout at 8192 bytes on a 36KB manifest. Reproduced outside vitest. The gate now pins Node the way it already pinned Python, and refuses to emit a verdict from the wrong toolchain.

### Full gate result on this branch

9 of 11 stages pass, including `web-core` (64s), `protocol` (29s) and `smoke` (40s). The two failures are the pre-existing `main` Governance breakage and the Node-skew artifact now guarded against — neither introduced here. Baseline comparison run on a detached `origin/main` worktree shows the identical failing set.

ankitkumarsingh1702:
## Full CI gate: 11 of 11 green, and main's Governance breakage is fixed

Ran the complete gate against `cf8abf5182ac7d4b379f0f57666a4a4dccb4d9d8`:

| Stage | Stands in for | Result |
|---|---|---|
| secret | Secret Scan | passed (3s) |
| governance | Governance | **passed (13s)** |
| dco | DCO | passed |
| pr-base-policy | PR Base Policy | passed |
| main-freshness | Base Freshness Gate | passed |
| web-core | Web Core (Next.js) | passed (65s) |
| web-targeted | Web Targeted Contracts | passed (14s) |
| protocol | Protocol (Python) | passed (17s) |
| mcp-package | MCP Package | passed (7s) |
| integration | Integration | passed (5s) |
| smoke | Main Post-Merge Smoke Gate | passed (48s) |

### This PR repairs a broken `main`

`main` has been red on **Governance** since #4902 merged. `/ria/claim` was added to the route layout without regenerating the three artifacts required to cover every route, and `generate_runtime_topology_index.py` fails closed when any of them disagrees.

Evidence it is pre-existing and not from this PR:
- Reproduced on a clean detached worktree at `origin/main`.
- PR #4902 (`feat/ria-claim-by-phone`) itself shows `Governance: failure` — it merged red.
- `codex/one-location-circles` then failed its `CI Status Gate` behind the same breakage.

Fix is three regenerated files (counts move 118 → 119). Nothing hand-authored: the generators read the real page and contract sources.

### `mcp-package` was a false alarm — retracting my earlier note

I previously reported `mcp-package` as failing. It was a Node version skew, not a defect: CI pins Node 24, this machine ran 20, and Node 20's `child_process.exec` truncates captured stdout at 8192 bytes on a 36KB manifest. Proven by running the same capture on both: **8,192 bytes on Node 20, 30,956 on Node 24.** The gate now borrows a matching Node from nvm for its own subprocesses, leaving the operator's default alone — with the shell still on v20.20.2, `mcp-package` passes.

### One pipeline, not two scripts

`scripts/ops/release_pipeline.sh` runs `PREFLIGHT → VERIFY → DEPLOY → PROVE → REPORT` as one command. Tested properties:
- `--from deploy` **cannot** bypass verification — the report check runs even when VERIFY is skipped.
- A commit that is not an ancestor of `origin/main` is refused.
- UAT is refused at PREFLIGHT with the reason, before anything is mutated.
- VERIFY runs in a throwaway worktree, so it never disturbs a shared checkout.

PR #4945, opened 2026-08-07
fix(kai): portfolio-save dead end and Analysis pager glitch

parthmawai:
## Summary

Two unrelated bugs in the Kai analysis workspace:

- **Portfolio-save dead end** — uploaded-portfolio imports never reached the review/save screen. The visible flow state reads from the `?stage=` URL param, but the transition into `"import_complete"` was missing from the allowlist that pushes it there, so the URL stayed on `"importing"` and rendered the branch with no continue button. Sample-portfolio loads skip this state entirely, which is why only that path worked.

- **Analysis pager glitch (Debate / Summary / Detailed View)** — the pager kept rendering two panes partially overlapped, worst when jumping straight to Detailed View. Root cause is in Embla carousel's own internals: it constructs its index `Counter`'s bound and scroll-bounds `limit` once, from whatever slide count it measured at that moment. A tab that mounts a beat later (once a computed flag settles) can grow the real slide count without either ever being recomputed — Embla exposes no public setter for either. Every later `scrollTo(index)` then silently clamps to the stale bound before computing a distance, and Embla's own per-frame rubber-band edge logic pulls any position beyond that stale bound back toward it, so even forcibly setting the correct target kept settling back near the start. Fixed by recomputing the target directly from `slideRects` (which stays accurate) instead of trusting Embla's own `scrollSnaps`/Counter, and disabling the rubber-band constraint, which only ever protects a bound that's wrong for this carousel's real content. `selectedScrollSnap()` reads through that same stale Counter, so `onSelect`/`onSettle` were also capable of reporting the previous pane after a correct scroll and silently reverting the caller's own selection — fixed by resolving the effective index from the rendered position everywhere the component reads it.

Also included: a small, unrelated guard in `useConsentPendingSummaryCount` (root app-shell hook) against a partial/error summary payload — found while testing the Kai fixes above, since it was throwing and taking down the entire app shell rather than just hiding a badge. Kept as its own commit since it's unrelated to the Kai bugs.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] `npm test` — full suite passes except 22 pre-existing failures in unrelated subsystems (marketplace encryption/SubtleCrypto, a location-sharing modal test, vault-sync mock), none touching the files in this PR
- [x] Manual: portfolio upload now reaches the review/save screen
- [x] Manual: Analysis workspace — Debate → Summary → Detailed View in all directions, including jumping straight from Summary to Detailed View, on both a live and a saved/historical run

### DIFF
--- a/hushh-webapp/__tests__/components/swipe-views.test.tsx
+++ b/hushh-webapp/__tests__/components/swipe-views.test.tsx
@@ -148,8 +148,21 @@ describe("SwipeViews", () => {
     expect(embla.options).toMatchObject({
       duration: 16,
       dragThreshold: 6,
-      watchResize: false,
     });
+
+    // Resize watching stays ON so a stale container measurement can correct
+    // itself, but is scoped to the container: per-slide entries (streaming
+    // content changing height) must not re-trigger the horizontal engine.
+    const watchResize = embla.options?.watchResize as (
+      api: { containerNode: () => Element },
+      entries: { target: Element }[],
+    ) => boolean;
+    expect(typeof watchResize).toBe("function");
+    const containerNode = document.createElement("div");
+    const slideNode = document.createElement("div");
+    const apiStub = { containerNode: () => containerNode };
+    expect(watchResize(apiStub, [{ target: containerNode }])).toBe(true);
+    expect(watchResize(apiStub, [{ target: slideNode }])).toBe(false);
   });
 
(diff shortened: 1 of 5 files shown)


kushaltrivedi5 (approved):
Approved — all CI checks green (CI Status Gate, Governance, Integration, Web Core/Targeted Contracts, PR Base Policy), no changes requested.

PR #5234, opened 2026-08-13
fix(location): make Auto-share approve requests, as its name says

parthmawai:
## Bug 1 of 5 — "auto share not working as intended"

**What it did.** The `Auto-share my location` switch did not auto-share anything. It gated whether *already-approved* grants kept publishing live updates. So switching it off silently froze every share recipients were already relying on, and switching it on did nothing anyone could observe.

**What it does now.** What the name says, and what it has always been read as: incoming location requests are approved without asking each time.

## The scope rule

Auto-approve applies **only to requests that arrive after it is switched on**.

Turning a setting on is not an answer to the specific people already waiting — some of whom may have been left unanswered on purpose. The moment of the switch is stored as a watermark and only strictly-later requests pass it. Ties are treated as already-waiting.

The whole rule is one function, [`selectAutoApprovableRequests`](hushh-webapp/lib/one-location/auto-approve-requests.ts) — every branch refuses; one line lets a request through. It refuses when: off, paused, no readable watermark, request not pending, no readable arrival time, or already attempted on this device.

## Defaults and storage

- **Defaults OFF.** Approving a location request is consent; a default may not give it.
- **New storage key.** The old `one_location_auto_share_v1:` value meant "keep publishing". Rereading it would turn a publishing preference into standing permission to approve people.
- Reads fail closed — absent, unparseable, or unreadable storage all mean off.

## Live updates decoupled

Live publishing now follows the grant and `Pause my location` alone. An approved, unexpired grant keeps updating until it is paused, revoked, or expires. That is the honest arrangement, and pause is the control that actually stops it.

## Bug 2 of 5 — Settings copy

Dropped the header blurb, cut two-sentence row descriptions to one short line each, and removed the standing reassurance that repeated what the rows already said. The one caveat kept is the one nobody can discover by looking: auto-approve does not clear the approvals already waiting. An automatic approval also toasts by name as it happens.

## Also: restores this suite on main

`__tests__/components/one-location-agent-page.test.tsx` has been failing on `main` since `getMapPreferences` shipped without being added to the service mock — **58 of 61 tests erroring on a missing function**, not on behaviour. Verified against a clean `upstream/main` checkout before and after.

## Verification

- `tsc --noEmit` clean; `eslint --max-warnings=0` clean on every touched file
- `auto-approve-requests.test.ts` — 10 new tests, all green
- `location-control-state.test.ts` — 9 green (watermark stamped once, dropped on off, fresh on re-enable, survives remount)
- `one-location-agent-page.test.tsx` — **60/61**, up from 3/61 on main. The one remaining failure is a pre-existing contact-sync mock-hoisting error, confirmed present on clean `main`.
- Voice gateway, surface map, orchestration index, and runtime topology regenerated (168 actions / 39 surfaces)

## Not verified

**iOS.** This is web-verified only. No Capacitor bridge surface is touched, so it should transfer, but it has not been run on the native build.


parthmawai:
**Rebase note — conflicting Settings copy resolved by hand.**

`main` shortened the same Settings descriptions independently while this was open, so the rebase conflicted on `location-redesign-hub.tsx` and `page.tsx`. How each was resolved:

| Line | Kept | Why |
|---|---|---|
| Header description | **dropped** | main's "Control live sharing." describes something the first control on this screen no longer does |
| Auto-share row | **this PR** | main's "Auto-share my location / Approved shares get live updates." is the behaviour this PR removes |
| Show me on their map | **this PR** — "People you share with can watch you move." | plainer, and names who; main's "Controls live movement on shared maps." is system-side language |
| Pause my location | **this PR** — "Stops new updates and checks you out of Nearby." | main's "Stop new updates." drops the Nearby checkout, which is a real consequence of pausing |
| Shield line | **this PR** — the Nearby consent caveat | main's "Private shares stay in your circle." restates what the rows above already imply |
| `handleApprove` refresh | **main** — `void refresh().catch(() => null)` | main's latency fix; the approval is already published, so holding the button through a state reload only makes it feel slower |

Happy to take main's wording on any of these if the author prefers it — flagging rather than quietly overriding.

PR #5248, opened 2026-08-14
fix(one-location): declutter the People tab and fix its action layout

DamriaNeelesh:
The People tab had grown to five full-width CTAs stacked between the search field and the list it filters. This reduces that to three, moves search onto the list, and fixes a layout that stretched every button to the width of the browser window.

## Search sits on the list it filters

It was separated from its own results by four buttons, which made it read as *page* search rather than *list* search. It now sits directly above the results under a `Connections` heading.

A test asserts that **nothing focusable** appears between the field and the first row, so this cannot quietly drift back.

## "Share to contacts" removed

It minted a **public location link** — the same artifact the Links tab creates, viewable by anyone holding the URL. Beyond being a duplicate, it was a category error: a control that makes you visible to *anyone with a link*, sitting among controls that are otherwise all about named, trusted people, one tap away from them.

`TemporaryLinkFlow` on the Links tab still owns create / copy / share / revoke, so no capability is lost. `handleShareContactInvite` stays because the legacy composer still calls it; only the People-tab buttons and the now-dead `onShareToContacts` view-model field are gone.

## Three actions, three weights

| Action | Weight | Why |
|---|---|---|
| Add people | filled | the common case |
| Invite | outlined accent | for people not on One yet |
| Sync contacts | quiet ghost | a one-off utility, not a peer of the other two |

Previously four buttons shared one weight, so nothing read as the default.

## Layout: desktop and mobile

The hub column has **no max-width**, so `w-full` buttons grew to the width of the whole desktop window — three pills spanning 1200px read as bulk, not as choices.

- **< 640px** — stacked full-width: proper thumb targets, which is also what the iOS/Android shell renders.
- **≥ 640px** — inline at natural width, left-aligned.
- `flex-wrap` lets a long localized label move to its own line instead of squeezing its neighbours.

## Circles

`Create` and `Join with code` moved from below the list to directly under the **Your circles** heading. Below the list they were stranded past a scroll exactly when someone had enough circles to need them. The assertion is structural — heading block is `children[0]`, action row is `children[1]` — so it cannot pass while asserting nothing.

## Notes for review

- **No copy was renamed here.** "Add people", "Invite" and the "Create link" labels are `main`'s, adopted through the rebase. Only weight, order and placement changed.
- The public-invite test now drives the Links tab, since that is the one remaining path to a public link.

## Verification

Rebased onto `56445fa3` and verified on that exact base:

- `tsc --noEmit` — clean.
- `one-location-agent-page` — **65/65 green**, the suite covering this screen, including the two new tests above.
- DCO signoff present; `scripts/ci/check-dco-signoff.sh` passes locally.

A full-suite comparison against a `56445fa3` baseline (32 failing test names, all pre-existing) is running locally; I'll post the name-level diff here when it completes. Comparing by test *name* rather than count is deliberate — a count alone once made a load-induced flake look like a regression.

### DIFF
--- a/hushh-webapp/__tests__/components/one-location-agent-page.test.tsx
+++ b/hushh-webapp/__tests__/components/one-location-agent-page.test.tsx
@@ -3477,10 +3477,17 @@ describe("OneLocationAgentPage", () => {
     await skipLocationEntryFlow();
 
     await waitFor(() => expect(mockGetState).toHaveBeenCalled());
-    // The populated People tab exposes a compact "Share to contacts" action.
-    fireEvent.click(screen.getByRole("button", { name: "People" }));
+    // Public links live on the Links tab. The People tab used to carry a
+    // duplicate "Share to contacts" that minted the same artifact; this is the
+    // one remaining path, and it must still create the invite.
+    fireEvent.click(screen.getByRole("button", { name: "Links" }));
+    // Links hub CTA opens the flow; the flow's own CTA creates the invite.
+    // Both are labelled "Create link", so take the one on screen each time.
     fireEvent.click(
-      await screen.findByRole("button", { name: /Share to contacts/i }),
+      await screen.findByRole("button", { name: /^Create link$/i }),
+    );
+    fireEvent.click(
+      await screen.findByRole("button", { name: /^Create link$/i }),
     );
 
     await waitFor(() =>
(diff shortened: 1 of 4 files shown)


DamriaNeelesh:
## Full-suite comparison — 0 regressions

Both sides measured on the **same base** (`56445fa3`), compared by failing test **name** rather than count.

| | Failing test names |
|---|---|
| `main` @ `56445fa3` | 32 |
| This branch | 33 |
| **Real regressions** | **0** |

The one name in the difference was `voice-action-settlement > waits for the new route publisher instead of settling on a stale surface`. I did not treat that as flaky on sight — it was verified:

- Passes **3/3 across three consecutive isolated runs** on this branch.
- Imports only `@/lib/voice/voice-action-settlement`; **zero** references to any file in this diff.
- This PR touches four files, all in the One Location People tab UI.

It is a timing-sensitive settlement test that flaked under full-suite parallel load, unrelated to this change.

### Method note

Counts alone are not a safe signal here. Earlier in this work a full-suite run showed 18 "new" failures that turned out to be load-induced timeouts — I had run ESLint concurrently with the suite. Re-run cleanly, all 9 of those suites passed. Since then every measurement has been taken with nothing else competing for CPU, and compared by test name.

Two further traps this run avoided: the baseline was initially taken at `56445fa3` while the branch still sat two commits behind on `87556df5`, which would have blamed this change for whatever those two commits did; and the commit message originally credited me with a `getMapPreferences` mock fix that `main` landed independently during the rebase. Both were corrected before pushing.

PR #5255, opened 2026-08-14
feat(one-location): email SOS contacts as a second channel

ankitkumarsingh1702:
Answers "their One is closed — how do they hear about the SOS?" Today: they don't. A Save my Soul alert reaches its contacts through exactly **one** channel, a push notification. A contact with notifications off, or who uninstalled, gets nothing, and the sender only learns that from a toast afterwards.

Email is the second channel. It survives a closed app and an unread notification tray.

## The coordinates decision

Migration 061 states the invariant plainly:

> `'Latest live-location ciphertext envelopes. Coordinates must be present only inside ciphertext.'`

The server **cannot read where the sender is** — only the recipient's device decrypts that. So an email that names a location has to receive coordinates from the sender's client, in the request, for the length of one send.

That is a deliberate product decision to weaken the envelope model for the emergency case: a contact who cannot open the app still needs somewhere to go. **Four rules bound it, each with a test:**

1. **Never persisted.** Nothing writes them to a table.
2. **Never logged.** `_log_safe` is the only formatting path into a log line and it refuses to format a float, so a later edit cannot casually interpolate a latitude. A test asserts no coordinate, note, or address appears in *any* log record across success and both failure paths.
3. **The grants are the authorization.** A caller may mail only the recipients of live SOS grants their own account created in the last 15 minutes. Without that, this endpoint would email arbitrary text and a location to any user id — and an 8-hour SOS grant would stay usable as a mailer for the rest of the day.
4. **Non-production routes to the test inbox.** Verifying an emergency mail on UAT must never mail a real contact's real inbox. The subject is prefixed `[TEST → real@address]` so it stays traceable.

## No new secret

Reuses the Workspace delegated service account that already sends support and invite mail and is already bound in every lane — no new binding to keep in parity across the three build lanes, no new IAM grant, no new deploy variable.

## Failure is never an error

The alert has already gone out by the time this runs, so a mail problem is a count, not an exception: the route swallows, the service returns per-recipient outcomes, the client helper resolves rather than throws. A missing credential reports `configured: false` and the alert is unaffected.

## What the recipient gets

Subject: **Ankit needs help — Save my Soul**, marked high priority, with the sender's note (HTML-escaped), the coordinates and accuracy, a Google Maps button, a link into One for the live position, the share expiry, and the sender's local emergency number.

The sender's toast now names both channels — `Alerted 2 contact(s). Emailed 2.` — and the all-unreachable case reads differently when inboxes did get it, which is the only reason to have a fallback at all.

## Tests

**23 new backend tests**, covering:

- rejects another user's grant, a non-SOS grant (`check_in` / `share` / `drive_to` / empty), a revoked or expired grant, a grant older than 15 minutes
- accepts one created moments ago; caps recipients; mails each person once across duplicate grants; skips a contact with no usable address
- escapes `` in the sender's note before it reaches an HTML mail
- routes a non-production send to the test inbox
- reports (never raises) on transport failure and on a 4xx from Gmail
- **no coordinate, note, or email address in any log record**

96 backend tests pass on Python 3.13. `npm run typecheck`, `verify:service-boundary`, and `eslint --max-warnings=0` all clean. `./bin/hushh protocol lint` and `format` clean.

Baseline: the Location + navigation sweep fails identically on `origin/main` and here (4 pre-existing failures).

## Not in this change, and why

- **Real SMS** — no telephony provider exists anywhere in this repository. That is a purchase plus per-country regulatory setup, not a code change.
- **Auto-call** — same provider dependency, plus auto-dialing carries its own legal constraints.

## Risk / rollback

Additive: one new backend route, one new service, one client helper. No schema change, no migration, no change to the existing push path. Straight revert.


ankitkumarsingh1702:
Superseded by #5259 (same commits, fresh branch, opened while trying to get the pull_request workflow trigger to fire).

PR #5314, opened 2026-08-16
test(one-location): measure the map header in a real browser, not in jsdom

ankitkumarsingh1702:
Follow-up to #5305. That PR fixed the header truncation; this one makes it **impossible to reintroduce silently**.

## The gap this closes

The truncation that reached a user as the single letter **"C"** was invisible to every test we had:

- **jsdom has no layout engine.** A unit test cannot tell you a string overflowed its box.
- **Class-name assertions** — which is what #5305 added — only prove the class we *intended* is present, not that the result fits.

So the suite was green the entire time the bug was on a user's phone.

## What this adds

`e2e/location-map-header-contract.spec.ts` measures real boxes in a real engine at **320 / 360 / 375 / 390 / 430 / 768 / 1280**:

| contract | assertion |
|---|---|
| product-owned labels never truncate | `scrollWidth <= clientWidth` on "Check in" and "Sharing with 2" |
| no accidental horizontal overflow | `documentWidth <= viewportWidth` |
| essential controls operable | close and locate ≥ 44×44 and inside the viewport |
| the structural rule that buys the space | Sharing on its own row below `sm`, true-centre row at `sm`+ |

**No dev server, no backend, no login.** The app's other e2e specs need `REVIEWER_UID` and a vault passphrase, so they could never have guarded this. The question here is purely *"does this markup, under this stylesheet, fit at this width"* — the compiled Tailwind answers that on its own. Runs headless in **4.5s**.

## Why the class strings moved

The header's classes now live in `lib/one-location/map-header-layout.ts`, imported by both the component and the test. A test that re-declares them measures a *copy*, and a copy drifts silently — which is exactly how this shipped green the first time.

## Verification

```
old layout → FAIL: "Check in" was clipped at 320px
                   "Check in" was clipped at 360px
                   "Check in" was clipped at 390px
                   "Check in" was clipped at 430px
new layout → 7 passed (4.5s)
```

The failure message is a machine-readable statement of the original report, at the exact widths it occurred.

| gate | result |
|---|---|
| `tsc --noEmit` | clean |
| `npm run lint` | clean |
| location suites (8 files) | 127 passed |
| `npm run test:layout-contracts` | 7 passed |

### Deliberately not in this PR
Wiring it as a required CI status check — that is a workflow-permissions change and belongs in its own reviewed commit. The script (`npm run test:layout-contracts`) is committed and runnable now.

One harness bug worth noting, since it looked exactly like a product failure: parallel Playwright workers initially shared one scratch file, so Tailwind occasionally scanned an empty file and the buttons measured 12px wide. The compile is now memoised with a per-process path.


ankitkumarsingh1702:
Closing and reopening from a clean base. This branch was cut from the pre-squash `fix/location-map-ios-hardening`, so its true two-dot diff against main deleted **3,101 lines across 44 files** — reverting `safe-changes/SKILL.md`, several test suites, `save-location-modal`, `named-circle-flows` and an iOS Swift plugin that landed on main while this was open.

Nothing was wrong with the geometry gate itself; the base was. Replacement PR is cut from current main (`2f42f907b`) and touches 4 files.

PR #5335, opened 2026-08-16
fix(one-location): stop the dialog scrim painting over the native map

ankitkumarsingh1702:
> *"I see first key, or is my the map in here. There is no map behind it. It is just one pin coming."*

**The map was rendering the whole time — behind a 55% black scrim with a 10px blur.**

## Root cause

`@capacitor/google-maps` does not draw into the page. It places a **native map view below the WebView**, and the WebView is punched through to reveal it. So anything the page paints over that region hides the map.

`globals.css` clears backgrounds and blur for exactly this reason — but only inside `[data-testid="save-location-modal"]`. The Radix overlay is a **sibling** of the sheet, not a descendant:

```
DialogPortal
├── DialogOverlay      ← bg-black/55 backdrop-blur-[10px]   (never matched)
└── DialogContent      data-testid="save-location-modal"     (cleared)
```

The CSS comment already says *"the Radix dialog sheet, **its overlay**, and the map surface itself — must be transparent."* The intent was documented; the selector never reached it.

That also explains why the pin, the "Selected spot" card and the buttons stayed perfectly crisp: they are ordinary HTML **above** the WebView. They were sitting on top of a scrim, with the real map hidden underneath.

## Fix

The overlay drops its background and blur only while the native map is on screen, and keeps them everywhere else — on web the map is an ordinary element inside the sheet, and on the details step there is no map to reveal.

## Verification

- **758 tests pass** across the whole One Location surface; `tsc --noEmit` and `eslint --max-warnings=0` clean
- Three new tests: transparent on native during the map step, scrim kept on web, scrim restored on native after leaving the map step
- **Mutation-checked** — pinning `nativeMapShowing` to `false` (i.e. the bug) fails the native test

## Also here

`defaultSavedLocationCategory` added to two module mocks that stub `saved-locations` wholesale. Both use the **real** behaviour rather than a stub, so a label being pre-selected over a saved place would still fail the test.

## Still open, not fixed here

Tapping the crosshair "locate me" control kills the app. I ruled out the two usual causes — `Info.plist` **does** carry `NSLocationWhenInUseUsageDescription`, and the JS path is fully guarded and returns `null` on error. A native crash cannot be caught in JS, so this needs a reproduction with the crash log; my best remaining hypothesis is `setCamera` on a map instance that was already destroyed by the picker's cleanup. I am not shipping a speculative fix for it.

### DIFF
--- a/hushh-webapp/__tests__/components/one-location-agent-page.test.tsx
+++ b/hushh-webapp/__tests__/components/one-location-agent-page.test.tsx
@@ -366,6 +366,16 @@ vi.mock("@/lib/one-location/saved-locations", () => ({
   DuplicateSavedLocationError: class extends Error {},
   addSavedLocation: mockAddSavedLocation,
   loadSavedLocations: mockLoadSavedLocations,
+  // Real behaviour, not a stub: the save sheet opens on whatever this returns,
+  // and a stub would hide a label being pre-selected over a saved place.
+  defaultSavedLocationCategory: (
+    existing: ReadonlyArray<{ category: string }> = [],
+  ) => {
+    const taken = new Set(existing.map((location) => location.category));
+    if (!taken.has("home")) return "home";
+    if (!taken.has("work")) return "work";
+    return "other";
+  },
 }));
 
 vi.mock("@/lib/services/pre-vault-sensitive-draft-service", () => ({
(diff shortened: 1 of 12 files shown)


ankitkumarsingh1702:
Superseded by a branch rebuilt on clean main. This PR could never run CI: #5333 was squash-merged from the same branch, so main carried one squashed commit while this branch still had the originals — GitHub reported `mergeStateStatus: DIRTY`, could not build the merge ref, and therefore never fired `pull_request` workflows. Same three commits, cherry-picked onto current main.

PR #5357, opened 2026-08-16
fix(copy): plain-word pass on onboarding, phone, unlock and Location — and the iPhone clipping it was hiding

ankitkumarsingh1702:
## What

A copy pass over the nine screens in the sign-up → Location journey, run through `plain-product-copy` and `less-text-karma-ji`. Titles, labels, buttons, helper lines and errors are now four words or fewer unless meaning, safety or consent needs more.

## The part that isn't cosmetic

The two lower feature cards on the Location onboarding step render their titles with `whitespace-nowrap` inside a `grid-cols-2` that is about **130px wide on a 375px screen**. The old titles didn't fit, and because the card is `overflow-hidden` they were **silently clipped on every iPhone width** — nothing threw, no test went red:

| Old title line | Needed | Had |
|---|---|---|
| `can't find each other?` | 191px | 130px |
| `Need help but can't` | 171px | 130px |
| `At the venue, but` | 149px | 130px |

The new titles fit at 375 / 390 / 393 / 430px.

## Deleted, not shortened

Each of these was already said by something else on the same screen:

- Phone: *"Choose your country code and enter your phone number. We'll send you a verification code."* — the two field labels and the button say all of it.
- Location: *"Share location, check in, or send help in seconds."* — the three cards below **are** those three things.
- Unlock: *"Enter your passphrase."* — the field label says "Passphrase", the button says "Unlock".

## Selected rewrites

| Screen | Before | After |
|---|---|---|
| Landing | Everything you save stays locked. / Nothing moves without your yes. | Locked. Nothing moves without your yes. |
| Login | You choose what One can see. | You choose what One sees. |
| Phone | Verify your phone number / Add your phone number to continue. | Verify your phone / We'll text you a code. |
| Phone | Send verification code · Verify and continue · Use a different number | Send code · Verify · Change number |
| Unlock | That passphrase did not match. Please try again. | Wrong passphrase. Try again. |
| Passkey | Enable quicker unlock? / You can keep passphrase unlock, or enable passkey and still retain passphrase and recovery-key backup. | Unlock faster? / Passphrase and recovery key still work. |
| Passkey | Not now, continue with passphrase | Not now |
| Location | Share your location easily with anyone. | Share where you are. |
| Location | Need to keep people updated? | Keep people updated |

## iOS

- `AppUITests.swift` dismissal list dropped the now-dead `"Not now, continue with passphrase"`. The shorter `"Not now"` was already in the list one line below, so the reviewer flow keeps working.
- `Info.plist` `NSFaceIDUsageDescription` said *"unlock your Vault"*. "Vault" is not allowed in user-facing copy and Apple shows this string verbatim in the system prompt.
- `NSLocationWhenInUseUsageDescription` dropped "foreground location" and "encrypted live-location update".

## One source of truth

`one-location-onboarding-copy.ts` is new and is imported by the component, both vitest suites and both Playwright specs. Previously these strings were re-typed in seven places and the picker spec's fixture had **already drifted** from the shipped headline.

## Left alone on purpose

- **The map renderer-consent paragraph.** It is a consent gate, its three claims are load-bearing, and the file documents a previous trim that lost two protected words ("only" and "Nearby"). Consent copy is exempt from the word budget.
- **"Your agents. Yours to own."** — the approved founder line in `docs/vision/agent-ontology.md`.
- **The Terms / Privacy sentence** on login — legal.

## Verification

- Full vitest suite green except two failures in `__tests__/app/one/setup/` that **reproduce identically on `origin/main`** (same assertion, same numbers) — pre-existing, unrelated.
- `tsc --noEmit` clean, `eslint --max-warnings=0` clean on every touched file.
- `verify:surface-map` and `verify:route-orchestration-index` both report current.
- New `e2e/one-location-onboarding-copy.layout.spec.ts`: 5 assertions across 4 iPhone widths, measuring the real strings against the real compiled CSS. **Mutation-checked** — reverting the copy module to the old strings turns all four width tests red with the exact overflow numbers in the table above.


ankitkumarsingh1702:
### Heads up: this PR also unblocks `main` for everyone

CI was red on the first two pushes, and it was not this branch's doing.

`Web Targeted Contracts` was failing on:

```
native-static-parity: native-route-inventory.json is missing ROUTES entries: /one/profile/preferences/voice
```

The voice-settings work (#5346–#5348) added that route to `ROUTES` without adding it to `native-route-inventory.json`. Proof it predates this branch — on `origin/main`:

- `lib/navigation/routes.ts` contains `preferences/voice` → **1 match**
- `native-route-inventory.json` contains it → **0 matches**

This branch changes no route, no voice file, and no Python file at all. Since that gate feeds the required `CI Status Gate`, **every open PR is currently red on it**, so I fixed it here rather than leave main blocked.

The entry needed no judgement: `app/one/profile/preferences/voice/page.tsx` is byte-identical to its siblings — every one of them is `export { default } from "@/app/profile/profile-workspace-page"` — so it is the `/one/profile/preferences/kai` entry with the path substituted.

Regenerating cascades here, as it always does: inventory → surface map → route orchestration index → Kai action gateway. All four generators were re-run; all four `--check` gates now pass locally. The generated diff is exactly one route's `native` block moving from `null` to populated, plus content hashes.

`excluded_routes` is deliberately left at `10`. It already disagreed with the array (which holds 14) before this branch — correcting someone else's stale counter isn't this PR's business.

For the record on the earlier run, everything else was already green: **Web Core (Next.js) ✅, Integration ✅, Governance ✅, Base Freshness Gate ✅** — only the parity gate and the status gate that depends on it were red.

PR #5376, opened 2026-08-17
fix(app-shell): clear the header's fade for anchor jumps and the section rail

ankitkumarsingh1702:
Follow-up to #5375, same defect class (safe-changes R21), two more surfaces.

Both measured clearance against `--top-shell-reserved-height` — the height the
top mask is **solid** to — rather than `--top-shell-mask-visible-height`, the
height it actually **paints** to.

| Surface | Cleared | Header paints to | |
|---|---|---|---|
| `scroll-margin-top` for `[id]:target` / `[data-page-primary]` / `[data-section-header]` | 60px | 82px | **22px under** |
| Desktop section rail (sticky) | 76px | 82px | **6px under** |

The anchor one is app-wide: every jump to a heading, on every route with
anchors, parked the heading inside the header. Its own comment read *"Anchor
clearance: hash-targeted elements must clear the fixed top shell"*.

## The test I nearly shipped that could not fail

The first version of the rail check **passed with the bug reintroduced**.

The fixture built the shell from `resolveSignedInShellContentOffset` and let
everything else inherit from `app/globals.css`. That looks faithful and isn't:
CSS substitutes a custom property using the values present where it is
**declared**, so `:root`'s `--top-shell-mask-visible-height` bakes in `:root`'s
own `--top-fade-active` (8px) and carries that computed value down, whatever the
shell sets below. The real route shell re-declares the whole derived block, so it
resolves against 22px.

The fixture was measuring a header **14px shorter** than the shipped one — and a
rail pinned 6px too high cleared it comfortably.

Fixed at the cause: the derived geometry is now one exported function,
`resolveTopShellGeometryStyle`, which `app/providers.tsx` spreads and the layout
contract builds its fixture from. Neither can drift without failing.

## Test changes

`ambient-chrome-tabs.contract.test.ts` asserted that block as **literal source
text** inside `providers.tsx`, so moving it broke the test. The greps are
replaced with assertions on the resolved values — those survive a refactor and
prove more than a string match ever did.

## Verification

- Both new browser cases mutation-checked: restore the old values, both go red.
- `npm run test:layout-contracts` — 125 chromium + 64 webkit passed.
- `npx tsc --noEmit`, `eslint` on every changed file — clean.
- `npm run test:ci` — 27 failures across 17 files, **every one also failing on
  untouched origin/main** (the suite is flaky by ±2 files; three baseline runs
  gave 17/27, 18/31, 17/27).
- All four generated indexes verified current — no artifact churn.

Adds R22. No user-facing copy changes.

### DIFF
--- a/.claude/skills/safe-changes/SKILL.md
+++ b/.claude/skills/safe-changes/SKILL.md
@@ -816,6 +816,36 @@ The first fails closed for any `flow` route that does not name the file owning
 its clearance. The second measures the pixels at eight widths in Chromium and in
 WebKit — WebKit being the engine the iOS app actually runs.
 
+### R22 — A fixture that inherits `:root` is measuring a different app than the one that ships
+
+**Incident (2026-08-17, extending the R21 contract to the sticky section rail.)**
+The new browser test passed with the bug deliberately reintroduced. The fixture
+built the shell from `resolveSignedInShellContentOffset` and let everything else
+inherit from `app/globals.css`, which looked faithful and was not: CSS
+substitutes a custom property using the values present where it is **declared**,
+so `:root`'s `--top-shell-mask-visible-height` bakes in `:root`'s own
+`--top-fade-active` (8px) and keeps that computed value as it inherits. The real
+route shell re-declares the whole derived block, resolving the same token
+against 22px. The fixture's header was therefore **14px shorter** than the
+shipped one — and a rail pinned 6px too high cleared it comfortably.
+
+A passing new test is not evidence. Only the mutation is.
+
+**Rule.** When a test reproduces a runtime surface outside the app, it must
+build it from the **same exported function** the app uses, never from a
+hand-picked subset of tokens. If a value is re-declared at a scope below
+`:root`, inheriting it instead of re-declaring it silently changes the number.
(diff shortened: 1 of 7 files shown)


ankitkumarsingh1702:
Correcting my own description above, after measuring on deployed UAT.

**The rail half is real and measured.** On `/research/protocol` (a tabbed route,
so reserved 130 / visible 150), the rail computed `top: 146px` against a header
painting to 150px — 4px under, on the live app.

**The anchor half is latent, not a live user-visible fix.** The rule was wrong and is
now right, but today it barely matches anything:

- `[data-section-header][id]` matches **nothing**. No component sets that attribute;
  `SectionHeader` in `components/app-ui/page-sections.tsx` uses
  `data-ui-role="section-header"`.
- `[data-page-primary][id]` matches **nothing**. `PageHeader` sets `data-page-primary`
  but never an `id`.
- `[id]:target` only applies during real hash navigation, and the app scrolls with
  `scrollIntoView` rather than hash links.

So the corrected token is right, and it will be right whenever those selectors do
start matching — but I should not have implied it fixed something users hit today.
The attribute-name mismatch is a separate pre-existing issue I found and did not
change, because changing it would alter behaviour for every section header and I
have no measurement to justify that here.

PR #5390, opened 2026-08-17
fix(one): the empty band under every screen, and five UI reports from UAT

ankitkumarsingh1702:
## What was reported

Fourteen annotated UAT screenshots. Six were already fixed on `main` and are listed at the bottom — the rest are here.

## The one that mattered

**Every page reserved the fixed bottom bars twice.** `app/providers.tsx` gives the scroll root `padding-bottom: var(--app-scroll-bottom-pad)` and its own comment says why it owns that: *"so feature routes do not need to guess at device safe areas or bar geometry."* `.app-page-shell` guessed anyway.

Measured in a browser: **112px from the page + 132px from the scroll root, under a 132px bottom chrome.** Profile carried a third copy. That surplus is also the answer to *"header overlay ho rha"* — it is what makes a short screen scrollable, so cards travel up under the top bar on a page that had nothing to scroll.

The second reserve was not even the right number. CSS substitutes a custom property using the values present where it is **declared**, and `--bottom-chrome-stack-height` is declared on the route shell, not on `:root` — so `--app-bottom-content-clearance` froze the `--app-bottom-inset` fallback and ignored the height `AppBottomShell` actually measures.

## The rest

| Reported | Change |
|---|---|
| *"TIME SELECTION ke liye itna space justified nahi"*, *"how long section is not appropriate"* | The ladder was a 3-column grid, so on a desktop each cell stretched to 258px in the Location shell and 440px in the Share card. From `sm` it is a wrapping row of content-width chips — one 44px row instead of 148px. Phones keep the grid. |
| Same, Custom open | The wheel was `mx-auto` in a 790px card with ~265px dead on either side. Left-aligned from `sm` up. |
| *"request mein bhi same aana chahiye… bas until i stop nahi dena"* | Ask had a two-column scroll wheel jammed into a settings-row trailing slot. Ask and Share now share one card and one ladder; Ask has no open-ended rung. Two lines of helper text that explained the obvious went with it. |
| *"location on / location pause toggle ke just neeche lao"* | The status was under the title, reading as a subtitle for the screen. Its own full-width row, right-aligned, directly under the switch — which costs the title no width, so it cannot bring back the wrap that once put a bare "On" on iOS. |
| *"phone number +91 9682…9352"* | Was `919682 •• •• 9352`. Now `+91 9682••9352` — calling code split off, one bullet per hidden digit. |
| *"max show ho rha phir sahi na??"* | Yes — ten end times on the hub would be a crowd. But "Ends 10:38 AM" over "Sharing with 2 people" reads as one window when the first may stop in 29 minutes. It says **"Last ends"** when there is more than one. |

## Tests

- **New** `e2e/app-shell-bottom-clearance.layout.spec.ts`, 320–1440. It measures the reserve, not unused viewport — a short page in a tall window has dead space that no padding change can fix. Failed at **244px against a 164px budget** before the fix. Enumerated in `test:layout-contracts` *and* added to the WebKit `testMatch`, because a spec that is not listed in both silently runs nowhere.
- The duration ladder contract grows a desktop half (768/1024/1440): one row, no cell over 170px, wheel aligned to the ladder. **Mutation-checked** — reverting either class string fails it. The ladder now also runs under WebKit, the engine iOS ships.
- Ask asserts the full Share ladder and the absence of any wheel spinbutton.
- 6 unit tests for the phone formatter, 1 for "Last ends".
- **Fixed a pre-existing CI flake**: `one-location-agent-page.test.tsx` asserted "capture has not happened yet" at an unsynchronised instant, failing ~1 run in 4 on untouched `main` (reproduced 1/4 on a clean baseline worktree). The real rule is an order — ask before you read — so it asserts invocation order. 6/6 green after.

## Verification

- `tsc --noEmit`, `eslint --max-warnings=0`: clean.
- `test:layout-contracts`: 138 chromium + 89 webkit passed.
- `test:ci`: **21 failed / 4457 passed** on this branch; **21 failed / 4449 passed** on `origin/main`. Identical failing set, same test names — all pre-existing.

## Already fixed on main, no change needed

Feed "Needs you" description over the timestamp (`5ff834908`), the Circle invite icon floating above its title (`84f4324a6`), the bare spinner on Create link — it names the wait now, and the wait itself is a GPS acquisition — and the "Done" row under the Custom wheel.


ankitkumarsingh1702:
Two things found while writing the tests for this batch, both pushed here:

**1. A fourth copy of the bottom clearance.** The R23 check written for the ledger lists every `:root` token that reads a shell-scoped var. Run against this tree it returned `--bottom-chrome-stack-height`, which led to `profile-stack-navigator.tsx` — pushed Profile screens were reserving that band a fourth time (scroll root + `.app-page-shell` + `.profile-home-screen` + this).

**2. `test:layout-contracts` has never run in CI.**

```
$ grep -rn "layout-contracts" --include=*.yml --include=*.sh --include=*.json .
hushh-webapp/package.json:17: "test:layout-contracts": ...
```

One hit — the script that defines it. 227 assertions across 11 spec files, each written after a reported pixel defect, running only on the author's laptop. JSDOM performs no layout, so these are the only web tests that can fail on overlap, clipping, tap-target height or a stretched control.

Now a pack in **Web Targeted Contracts**, matched on layout specs, `playwright.config.ts`, `globals.css` and the component trees the specs measure. Both browsers install — WebKit is the engine iOS ships, and a Chromium-only run passes controls that are visibly broken on an iPhone. Proven against this PR's own file list: 138 chromium + 89 webkit.

Issue #5387, opened 2026-08-17
App feels slow to load/refresh — suspected caching or DB-load issue


parthmawai:
## Summary
Screens feel slow to load and refresh -- data does not reflect instantly and takes noticeable time to catch up. Reported hypothesis: a caching issue where the cache is not loading instantly.

## Sector
Hushh Research

## Environment
Local

## Reproduction steps
Use the app normally (Location and related screens) and observe that loading/refreshing state takes longer than expected to settle.

## Expected result
Screens load and refresh close to instantly, using cached state first and reconciling in the background.

## Actual result
Noticeable delay before data reflects -- feels like a stale cache or slow round trip rather than an instant read.

## Possible related cause (needs triage, not yet confirmed)
There is a previously-identified, deliberately deferred performance issue in Location's state/consent loading: 23-77 SQL queries per call, 26-47s+ per load, which starves the shared DB pool. That is a DB-load problem, not literally a caching one, but it would produce exactly this symptom (slow to load, slow to refresh) and is a strong candidate root cause here. Whoever triages this should check that path before assuming a client-side cache bug.

## Impact
General usability -- affects how responsive the app feels across normal use, not tied to one specific action.

## Evidence
Reported directly by the user during this session; not yet reproduced/measured by an agent.

Jhumma-hushh:
@parthmawai 
please work closely, i can't see progress at my end

parthmawai:
Progress: root-caused and fixed the real query-volume problem behind this.

Traced it to `/api/consent/center/summary` (called by every screen refresh, not just Consent Center itself): its default code path computes pending/active/previous counts via 3 concurrent calls, and for the common case (investor, consents) each of those 3 independently re-fetched the exact same location + marketplace data — a straight 3x re-fetch of identical results, not per-item loops. That's what actually produced the "~77 queries per call" figure, and matches the "slow to load/refresh" symptom since it fires on regular use, not one specific action.

Fix: fetch that data once per summary call instead of once per surface, and reuse it three ways. Same authorization/filtering, ~3x fewer round trips for this path. Up in #5466, CI-green, not merged yet — flagging progress here per your ask rather than waiting for merge.

[...]

PR #5466, opened 2026-08-18
perf(consent): fetch investor location/marketplace buckets once per summary, not per surface

parthmawai:
## Summary
Addresses #5387 ("App feels slow to load/refresh — suspected caching or DB-load issue"). Not auto-closing on merge — issue stays open until this is verified on UAT.

Root-caused the "~77 queries per call" figure to `consent_center_service.py`'s default (V1, `CONSENT_CENTER_SUMMARY_V2_ENABLED` unset) `get_center_summary` path. `pending`/`active`/`previous` counts are computed via `asyncio.gather` over 3 concurrent `_get_surface_count` calls, and for `investor` + `consents` (the common case), **each of those 3 independently re-fetched the exact same `_location_buckets_async`/`_marketplace_buckets_async` result** (~24 queries each). This is a straight 3x re-fetch of identical data, not per-item N+1 loops — and it alone reconstructs the reported query count almost exactly.

## Why this fix is safe
Both bucket-fetch functions take only `user_id` and are read three different ways by the caller (`incoming_requests` / `active_grants` / `history`) — there's no per-surface filtering happening *inside* them. Hoisting one fetch of each above the `gather` and passing the already-computed dict through requires no re-implementation of authorization/filtering logic; it's the same object read three ways instead of fetched three times.

## Changes
- `_get_surface_count` accepts optional `location_buckets`/`marketplace_buckets` kwargs (default `None` preserves the original fetch-it-yourself behavior for any other caller — there's no other caller today, but the signature stays backward compatible).
- `get_center_summary`'s V1 branch, gated to `actor == "investor" and mode == "consents"` specifically (the only case that ever touched these buckets), pre-fetches both once via `asyncio.gather` and threads them through all three surface calls.
- The already-optimized V2 path (behind the still-off `CONSENT_CENTER_SUMMARY_V2_ENABLED` flag) already did this correctly on its own — left untouched.

## Not addressed here (flagged, not blocking)
`_owned_user_identifiers` memoizes per `ConsentCenterService` instance with no lock, so concurrent `gather` branches can still race past the cache before the first write lands — a smaller, separate query-count source. Lower impact than the fix above, and needs its own care (a lock or memoized coroutine) rather than a drive-by change bundled into this PR.

## Test plan
- [x] New `test_consent_center_summary_query_dedup.py`: asserts `_location_buckets_async`/`_marketplace_buckets_async` are each called exactly once (not 3x) for an investor+consents summary, and asserts an RIA-actor summary never touches either bucket function at all (confirms the gate is correctly scoped).
- [x] `pytest tests/services/test_consent_center_summary_query_dedup.py tests/services/test_consent_center_connection_requests.py tests/test_consent_center_query_bounds.py tests/test_consent_center_page_cap.py tests/test_consent_center_pure_helpers.py` — 101 passed.
- [x] `ruff check` / `ruff format --check` clean.
- [x] `mypy` clean on changed lines (5 pre-existing unrelated errors elsewhere in the file, outside this diff's line ranges).

### DIFF
--- a/consent-protocol/hushh_mcp/services/consent_center_service.py
+++ b/consent-protocol/hushh_mcp/services/consent_center_service.py
@@ -1289,7 +1289,18 @@ class ConsentCenterService:
         actor: str,
         surface: str,
         mode: str = "consents",
+        location_buckets: dict[str, list[dict[str, Any]]] | None = None,
+        marketplace_buckets: dict[str, list[dict[str, Any]]] | None = None,
     ) -> int:
+        """
+        ``location_buckets``/``marketplace_buckets`` let a caller that already
+        fetched them for the investor+consents case (get_center_summary calls
+        this once per surface) pass them straight through instead of each
+        call independently re-fetching the same ~24-query bucket 3x over
+        (#5387 -- this alone accounted for most of the query volume behind
+        "app feels slow"). ``None`` (the default) preserves the original
+        fetch-it-yourself behaviour for any other caller.
+        """
         normalized_actor = "ria" if actor == "ria" else "investor"
         normalized_mode = "connections" if mode == "connections" else "consents"
         if normalized_mode == "connections":
@@ -1308,11 +1319,12 @@ class ConsentCenterService:
                 ]
             )
         if normalized_actor == "investor":
-            location_buckets = (
-                await self._location_buckets_async(user_id)
-                if normalized_mode == "consents"
-                else None
-            )
+            if location_buckets is None:
+                location_buckets = (
+                    await self._location_buckets_async(user_id)
+                    if normalized_mode == "consents"
+                    else None
+                )
             location_count = 0
             if location_buckets:
                 if surface == "pending":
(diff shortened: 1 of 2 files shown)


parthmawai:
Closing — this duplicated an optimization that already exists correctly, in the wrong place.

`consent_center_service.py` already has a "V2" path (`_consent_summary_v2_enabled()`) that fetches the location/marketplace buckets once and derives all three surface counts from it — the code's own comment describes solving this exact bug, and the flag defaults ON when unset. The actual reason the slow 3x-redundant path runs in UAT is that `deploy-uat.yml` explicitly pins `CONSENT_CENTER_SUMMARY_V2_ENABLED` to `'false'` by default.

This PR instead duplicated similar logic into the "legacy" (V1) path — which directly fights `test_consent_center_summary_explicit_opt_out_uses_legacy_surface_counts`, a test whose entire purpose is keeping V1 byte-for-byte untouched as a deliberate rollback lever. Caught that on CI failure rather than papering over the test.

Correct fix (config, not code) coming in a follow-up PR: flip the UAT deploy default so the already-built, already-tested fast path actually runs.

PR #5467, opened 2026-08-18
feat(location): offer Saved Places after checkout, neutralize Check out button

### LINKED ISSUE #5457: Location Check-Out: post-checkout 'Save to Saved Places' prompt & replace red button color with neutral styling
## Feature & UX Design First-Principles Summary
Two refinements for the Location Check-Out lifecycle:
1. **Post-Checkout Saved Places Prompt:** Prompt users upon checkout with a convenient follow-up action to save the location to their "Saved Places".
2. **Button Color Semantics (UX First Principles):** Change the "Check Out" button color from red (destructive) to neutral/secondary styling, reserving red exclusively for dangerous or emergency triggers.

### 1. Post-Checkout Saved Places Follow-up
- Upon checking out from a location:
  - Present a sleek, non-blocking post-checkout banner/micro-sheet:
    *"Checked out from <Location Name>. Save this to your Saved Places?"*
  - Provide a 1-tap `[ ⭐ Save to Places ]` button with optional nickname/category tag, alongside a quick dismiss.
  - Automatically suppress the prompt if the location already exists in the user's Saved Places.

### 2. Action Color Semantics (Non-Destructive)
- **Problem:** The "Check Out" button currently renders in red. In design systems, red is reserved for destructive, irreversible, or emergency alerts (e.g. Delete Account, SOS Trigger).
- **Expected:** Checking out of a location is a standard, safe operational lifecycle event. Style the "Check Out" button using neutral secondary, outline, or brand dark tokens with high contrast.

## Acceptance Criteria
- [ ] Update "Check Out" button color from red/danger to neutral/secondary design tokens.
- [ ] Implement post-checkout follow-up prompt to save the checked-out location.
- [ ] Prevent duplicate prompts if the location is already saved.
- [ ] Verify responsive behavior on mobile and desktop viewports.

### LINKED ISSUE #5760: feat(location): offer Saved Places after checkout, neutralize Check out button
Tracking item for **PR #5467** — https://github.com/hushh-labs/hushh-research/pull/5467

Filed as an issue because every view on [board 79](https://github.com/orgs/hushh-labs/projects/79) filters `is:issue`, so a PR added to the board directly is invisible in all three views.

See the pull request for the full description, diff, and test results.

Closes when #5467 merges.

anoushkauoc:
Closes #5457.

## 1. Check out is no longer destructive

`nearby-check-in-sheet.tsx` rendered the button as `variant="destructive"`. Checking out is a safe, reversible step in the lifecycle — you can check back in. Red is reserved here for the dangerous and irreversible (SOS, delete), and spending it on a routine action drains the signal from the places that need it.

Now `variant="outline"`: neutral fill, bordered, full-contrast foreground — the "neutral secondary / outline" the issue asked for, using existing design-system tokens rather than a new one.

## 2. Post-checkout Saved Places offer

On a successful checkout the sheet offers to save the venue, with a one-tap **Save to Places** and a dismiss.

Everything needed already existed — `placeLabel`/`placeLat`/`placeLng` on the presence record, and `findDuplicateSavedLocation` with a 25 m radius — so this is mostly wiring, plus a handful of decisions that were easy to get wrong:

- **The anchor is snapshotted before checkout runs**, because checkout clears the presence record. Reading it afterwards would always yield null.
- **`placeLat`/`placeLng` are the venue the owner picked, not their live position.** That distinction is what makes this safe to save at all — it saves the bar you chose, not wherever you happened to be standing.
- **Suppression reuses `findDuplicateSavedLocation`** rather than inventing a second notion of "same place".
- **Category is pinned to `"other"`.** `defaultSavedLocationCategory` returns `"home"` when no home is set, so the first venue anyone checked out of would have been silently filed as their **home** — a wrong answer that is tedious to undo.
- **The banner renders above the active/setup branch.** Checkout flips the sheet between those two views, so a banner living inside either one would be unmounted by the very action that creates it.
- **Every failure path is silent.** This follows a checkout the person already saw succeed; an unasked-for nicety must never raise an error toast of its own. Locked vault, failed read, or already-saved all just skip the offer.
- **The offer is retired** when the sheet closes or a new check-in starts, so it never sits above an active check-in naming a different venue.

### On `vaultKey`

Saved Places needs an unlocked vault, which the sheet did not have. I pass `vaultKey` as a **prop** rather than calling `useVault()`, matching how this sheet already receives `ownerId` and `vaultOwnerToken`. The parent (`location-immersive-map.tsx`) already calls `useVault()`, so it just destructures one more field.

I tried context first: it throws without a provider and broke all 38 existing tests. Props keep the sheet provider-free and its suite intact.

## Testing

6 new tests: offer shown, offer suppressed when already saved, no offer when vault is locked, saves as `"other"`, dismiss without saving, and Check out is not destructive.

The suppression test drives the **real** duplicate-radius logic — only the two vault-backed calls are stubbed, via `importOriginal`, so it exercises the actual 25 m rule instead of a mock that agrees with whatever the component asks it.

- Sheet suite: 44 passed (38 existing + 6 new).
- Wider area: 28 files, 423 tests, all green.
- `tsc --noEmit` and `eslint --max-warnings=0` clean.

Pre-existing failures in `lib/one-location/__tests__/drive-recents.test.ts` and `emergency-numbers.test.ts` (13 total) are **unrelated and already red on clean `main`** — verified by stashing.

## Deliberately not built

The spec mentions an "optional nickname/category tag" on the save button. I kept it a genuine one-tap save: Saved Places already supports renaming and recategorising via `updateSavedLocation`, so rebuilding an editor inside a banner meant for a single decision would work against the "sleek, non-blocking" framing. Happy to add it if you'd rather it be there.

Not yet verified on a live surface.

Closes #5760


anoushkauoc:
@RGlodAkshat — flagging this one for you since @Jhumma-hushh routes Location work your way.

Rebuilt on current `main` just now (branch was 210 commits behind; all four touched files had been substantially rewritten in the meantime). Git reported the old branch as `MERGEABLE`, which was the risk — it would have merged cleanly while my feature no longer matched the reshaped code. Re-verified after the merge: 45 tests in the sheet suite, 368 across `components/one-location/`, `tsc` and `eslint` clean.

The parts genuinely worth your eyes, rather than the diff as a whole:

1. **`vaultKey` as a prop, not `useVault()`.** Saved Places needs an unlocked vault, which the sheet did not have. I pass it from `location-immersive-map.tsx`, matching how the sheet already receives `ownerId` and `vaultOwnerToken`. Context throws without a provider and broke all 38 existing tests. If you'd rather this sheet take the provider, that's a bigger call than I wanted to make unilaterally.

2. **The anchor is snapshotted before `checkoutNearby` runs**, because checkout clears the presence record. Worth a sanity check that I'm reading `placeLat`/`placeLng` (the venue picked) and never the owner's live position.

3. **Category pinned to `"other"`.** `defaultSavedLocationCategory` returns `"home"` when no home is set — so without the pin, the first venue anyone ever checked out of would be silently filed as their **home**.

4. **Every failure path is silent** — locked vault, failed read, already-saved. This follows a checkout the person already saw succeed, so I did not want a follow-up nicety raising its own error toast. Tell me if you'd rather a read failure surfaced.

Not verified on a live surface — I can't dispatch UAT. Once this merges, one deploy of the resulting `main` SHA covers this plus three already-merged fixes.

PR #5472, opened 2026-08-18
fix(security): CSP/security headers, native-test-bridge lockdown, vault KDF hardening

smirthi-dharma:
## What this fixes

From the 2026-08-17 security assessment:

- **Content-Security-Policy + security headers.** Added a nonce-based CSP plus X-Frame-Options, X-Content-Type-Options, Referrer-Policy, Permissions-Policy, and HSTS (`proxy.ts`). Served SVGs are sandboxed. Verified against a real production build (`npm run build` + `npm run start`): all headers present, zero CSP violations in the browser console on a fresh load.

- **Native test-mode bridge locked down.** `window.__HUSHH_NATIVE_TEST__` is now only trusted when `Capacitor.isNativePlatform()` is true, everywhere it's read — previously any script running on the page could set it directly. Auditing every call site also found it could skip phone verification and attach a callable persona-switch hook to the page; both closed by the same fix.

- **Passphrase stretching raised 100,000 → 600,000 PBKDF2 rounds** for new vaults, with full backward compatibility — the round count travels inside the (already backend-opaque) salt field, so existing vaults keep unlocking at their original round count. New coverage in `__tests__/vault/passphrase-key-kdf-versioning.test.ts`.

- **Removed a dead recovery-key code path** that derived a wrapping key from a salt hardcoded to the same value for every user. Confirmed via full-repo search it was unused.

- **Vault-wrapping key is now non-extractable** — it was extractable with unused wrapKey/unwrapKey capabilities; every call site only ever encrypts/decrypts with it.

- **Reviewed the cookie/CSRF surface** flagged in the assessment: no backend route reads `request.cookies` and nothing ever sets one, so there's no cookie today for a forged request to ride on.

## Verification

- Full production build + live header/console check against it
- `tsc --noEmit` clean, `eslint` clean
- Full vitest suite: 940/942 passing (2 pre-existing failures, unrelated, verified in isolation)

### DIFF
--- a/consent-protocol/tests/services/test_connections_service.py
+++ b/consent-protocol/tests/services/test_connections_service.py
@@ -376,9 +376,7 @@ def test_accept_request_never_imports_or_calls_location_service():
     from hushh_mcp.services.connections_service import ConnectionsService
 
     lines = inspect.getsource(ConnectionsService.accept_request).splitlines()
-    code_only = "\n".join(
-        line for line in lines if not line.strip().startswith("#")
-    )
+    code_only = "\n".join(line for line in lines if not line.strip().startswith("#"))
     assert "one_location_agent_service" not in code_only.lower()
     assert "OneLocationAgentService" not in code_only
     assert "auto_start_share_for_new_peer" not in code_only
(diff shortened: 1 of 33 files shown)


smirthi-dharma:
Closing in favor of a clean branch rebased directly on current main (this one carried stale/unrelated history from a fork branch that was behind).

Issue #5410, opened 2026-08-17
Circle [id] route: remove redundant pencil icon on editable circle name input


kushaltrivedi5:
## Bug / UI Polish Summary
On the Circle detail/management route (`/one/location/circle/[id]`), a decorative pencil (edit) icon is rendered on the circle name text input.

## Problem
- **Redundant Affordance:** The circle name is already an interactive text input field with native focus and typing states.
- **Visual Noise:** The screen already features a dedicated, well-placed "Save" button in the action bar, making the pencil icon superfluous and awkward.

## Expected Behavior
- Remove the redundant pencil icon from the circle name input field.
- Rely on clean input field styling (focus ring, placeholder, active cursor) and the existing Save button for a streamlined editing flow.

## Acceptance Criteria
- [ ] Remove the pencil icon element from the Circle name editable header/input on `/one/location/circle/[id]`.
- [ ] Verify clean alignment and focus states across mobile and desktop screens.

Jhumma-hushh:
@DamriaNeelesh

DamriaNeelesh:
@parthmawai can update the status of this issue if this needed or fixed , else remove it

PR #5474, opened 2026-08-18
fix(circle): remove redundant pencil icon on circle-name field

parthmawai:
## Summary
Addresses #5410. The Circle detail route (`/one/location/circle/[id]`) showed
a decorative pencil icon beside the circle-name input even though the field
is already focusable/clickable on its own and a "Save" button already appears
once the name is edited.

- Removed the pencil `Button` and its `onClick` focus handler.
- Kept the trailing slot's exact box (an invisible `aria-hidden` placeholder)
  so focusing the field still never shifts width when Save appears — this
  geometry is deliberately locked down by `circle-name-row-layout.ts` and
  `e2e/connect-circle-cta.layout.spec.ts`.
- Updated the two unit tests that asserted on the old "Edit Circle name"
  button to assert on the new placeholder instead.

## Test plan
- [x] `npx vitest run components/one-location/redesign/circles/__tests__/named-circle-flows.test.tsx` — 26/26 passing
- [x] `npx tsc --noEmit` — clean
- [x] `npx eslint` on both touched files — clean

### DIFF
--- a/hushh-webapp/components/one-location/redesign/circles/__tests__/named-circle-flows.test.tsx
+++ b/hushh-webapp/components/one-location/redesign/circles/__tests__/named-circle-flows.test.tsx
@@ -897,8 +897,11 @@ describe("named Circle flows", () => {
     const input = (await screen.findByLabelText(
       "Circle name",
     )) as HTMLInputElement;
-    // Untouched: the trailing control is the edit affordance, never a write.
-    expect(screen.getByRole("button", { name: "Edit Circle name" })).toBeTruthy();
+    // Untouched: no edit affordance is needed, and the trailing slot is never
+    // a write until the name actually changes.
+    expect(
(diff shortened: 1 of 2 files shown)


parthmawai:
## Obsolete — recommend closing rather than resolving the conflict

Checked this while clearing conflicts. **The redundant pencil this removes no longer exists.**

`main` has replaced the whole affordance: the circle name is now edited through an explicit **"Edit"** button that opens a rename sheet (`named-circle-flows.tsx:1750-1760`), and there is no `Pencil` import left in the file.

The PR also deletes `nameInputRef` — which `main` now uses to focus that sheet's input when it opens (`:1755`). Resolving the conflict mechanically would **introduce a bug**, not fix one.

The change was right when it was written; the UI moved past it. Left open for a maintainer to close — full audit of this and the seven sibling PRs is in #6208.

PR #5490, opened 2026-08-18
fix(one-location, connect): pagination row wrap + Check-In hijacked by Nearby's map consent

ankitkumarsingh1702:
## Summary

Two independent screenshot reports on One Location / Connect, from real UAT use:

- **Connect pagination footer** — Prev/Next dropped to a line below "Page N · Per page" on every phone. The row was `flex-col`, switching to `flex-row` only past Tailwind's `sm:` (640px) breakpoint, which this screen never reaches at phone width. Fixed to an unconditional row, dropped the redundant "Per page" label, and hardened the page indicator against word-wrapping on the narrowest supported width (320px) — caught by the new layout-contract spec itself.
- **"Check-In" tile / voice action opening a full-screen map consent screen on first tap** — traced to Nearby Check-In (a proximity, map-based feature) silently taking over the "Check-In" tile and its `?action=check-in` deep link whenever it's available (always outside production). The tile's own subtitle and the voice-contract's own `meaning` field both still describe the lightweight "send a one-off note" flow, so this was contract drift, not a deliberate conflation. Fixed by giving Nearby Check-In its own "Nearby" row and restoring Check-In to the private flow it always promised.

A third reported item — removing "Show me on their map" from Location Settings — was investigated and intentionally left unchanged: unlike "Pause my location" (already removed earlier as a true duplicate of the header switch), this toggle is the *only* place that control lives, deliberately surfaced there for discoverability. Removing it would hide map-visibility control with no replacement, so it stays.

## Changes

- `app/connect/page-client.tsx`, `app/connect/connect-pagination-layout.ts` (new) — pagination footer layout fix, geometry constants extracted for the layout-contract spec.
- `e2e/connect-pagination.layout.spec.ts` (new), `playwright.config.ts`, `package.json` — real-browser regression coverage (chromium + webkit), registered in `test:layout-contracts`. Includes a mutation check proving the spec fails against the original broken markup.
- `components/one-location/redesign/location-redesign-hub.tsx` — Check-In tile always opens the lightweight flow; new "Nearby" row opens Nearby Check-In; removed the `?action=check-in` → Nearby redirect.
- `__tests__/components/one-location-check-in-destination.contract.test.ts` — updated to lock in the new contract (Nearby has its own route; Check-In's action is never redirected into it).

## Known, deliberately deferred gap

One's voice `execution_target` for `location.open_check_in` (`contracts/kai/*.json`) still points at the dedicated Nearby route, so saying "check in" to Kai and tapping the Check-In tile now diverge. Fixing that needs a new voice action plus regenerating the governed route indices — out of scope here, flagged for a follow-up.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npx eslint` on every touched file — clean
- [x] Full `one-location-agent-page.test.tsx` (101 tests), `one-location-check-in-destination.contract.test.ts` (11 tests), `check-in-flow.test.tsx`, `connect page-client` contract + component tests — all passing
- [x] `npx playwright test e2e/connect-pagination.layout.spec.ts` on **chromium and webkit**, all 7 tests passing, including the mutation check against the original broken markup


ankitkumarsingh1702:
Tracked in #5495.

PR #5534, opened 2026-08-18
fix(git): detect a merge conflict resolution that silently drops a landed fix

ankitkumarsingh1702:
## Summary

Investigated tonight's location-save/check-in feature that appeared to "revert to its first version." Root cause: `feat/agents-screen-plain-copy-redesign` fell 90 commits behind `main`. Its working tree already had the fix (uncommitted), but the branch's own committed HEAD still carried the pre-fix behavior — nobody broke anything with a bad commit, the sync was just never done.

That surfaced a real gap: **Base Freshness Gate proves a branch contains main's commits as ancestors; it says nothing about whether resolving those commits' conflicts kept their content.** A developer or agent can sync main in, hit a real conflict, resolve it by keeping the stale side to make it merge, commit, and the freshness gate reports "fresh" — because ancestry is satisfied even though content silently dropped. Confirmed this is exploitable with a reconstructed end-to-end test (conflict → bad resolution → would have shipped undetected).

Also swept every merge into `main` since 8 PM IST tonight (56 real sync-merges across ~6 contributors) for the same pattern — no other instance found, after ruling out two false-positive shapes the tool itself was producing (see below).

## What this adds

- `scripts/git/check-merge-regression.sh <base> <upstream> [target|--worktree]` — for a file changed between a merge's old fork point and upstream, checks both directions: upstream's added lines missing from the result, and upstream's deleted lines still present in the result (the "stop doing X" fix shape — a redirect that should have been removed but survived a bad resolution — only shows up in the second check). Skips a file outright if the result is byte-identical to upstream, and drops any line that recurs more than once in its own source file — both fixes came from real false positives found while testing this against tonight's actual merge history (a shared "list" file two branches both append to; a repeated callback signature in a large file).
- `./bin/hushh sync main` now captures the fork point and runs the check automatically after a clean merge.
- `./bin/hushh sync verify-merge` — run after hand-resolving a conflict `sync main` stopped on; checks the resolution the same way.
- `scripts/git/scan-pr-merges-for-regression.sh` — walks every real sync-merge a PR introduces and runs the detector on each.
- `ci.yml`: new **"Merge Fidelity Gate"** job, advisory only (`continue-on-error`, kept out of `ci-status`'s `needs`, `MERGE_REGRESSION_MODE=warn`) — cannot block a merge, surfaces signal for review.

Everything defaults to `warn`, not `block` — it's a text heuristic with a known false-positive rate, not a semantic proof. `MERGE_REGRESSION_MODE=block` is available once someone trusts it for their own files.

A companion skill (`merge-regression-guard`, in `~/.claude/skills/`) documents when to use this and the known false-positive shapes.

## Test plan

- [x] Unit-style: ran the detector against ~10 real merges from tonight (clean and known-conflicted) — confirmed no false positives after two rounds of fixes.
- [x] End-to-end: reconstructed the actual incident in a disposable worktree — real conflict, bad resolution (kept the stale side) → caught; good resolution → passes clean.
- [x] Ran `scan-pr-merges-for-regression.sh` across the full range of tonight's ~56 sync-merges — 0 false positives after fixes, and it would have caught the incident's shape.
- [x] `bash -n` syntax check on all four scripts; tested under macOS stock bash 3.2 (no `mapfile`, unbound-array-on-empty-array quirk) since that's what this checkout runs locally.
- [ ] Not run: real-world usage by another contributor on their own branch (this is advisory-only by design, so no live PR is at risk either way).


ankitkumarsingh1702:
Superseded by #5638. Same work rebuilt on current main, plus two fixes: the detector could never print a finding (SIGPIPE from `git log | head -5` under `set -euo pipefail` killed it at the first flagged file, which behind `continue-on-error` looked exactly like a pass), and the job only scanned sync-merge commits, so it was blind to the revert shape that actually happened here — `a60c51dfc` has one parent and sits on main's tip.

PR #5560, opened 2026-08-19
fix(one-location): say WHICH share ended when an SMS share is stopped

DamriaNeelesh:
## Yes, this is really happening

Verified before changing anything. Stopping an SMS alert is **not a distinct code path** — `handleStopSos` calls `revokeGrant` for each grant, which lands in the same `revoke_grant` an ordinary share uses. The copy there never named the lane, so both produced the identical sentence:

> **"X removed your location access."**

For someone who has only ever received an emergency SMS, that is a message about access they don't know by that name.

## Why it stopped being merely vague

**#5552** split the lanes so a person can hold an ordinary share **and** an SMS share with the same counterpart at the same time. An unnamed *"your share ended"* is now genuinely ambiguous — exactly when someone is checking whether the emergency one is still running.

## The grant already knew

`share_kind` is on the row (both queries in `revoke_grant` select `*`), and #5552 established it as **the** discriminator between the emergency lane and everything else. This reads it rather than inventing a second signal.

## Fixed on all three surfaces the message reaches

| Surface | Before | After |
|---|---|---|
| Push notification (`revoke_grant`) | "X removed your location access." | "X stopped sharing their location with you over SMS." |
| In-app activity feed (`_activity_event_payload`) | "X stopped sharing" | "X stopped SMS location sharing" |
| Client fallback (`locationWorkflowNotificationCopy`) | "X removed your location access." | "X stopped sharing their location with you over SMS." |

Titles become **"SMS location sharing stopped"**.

**SMS, not SOS.** `"sos"` is the column's word; an SMS alert is how it reached the recipient, and the copy has to match what they remember receiving. Asserted in both test suites.

`share_kind` also travels in the notification payload — the grant is revoked by the time the client renders this, so it cannot look the kind up from a share that no longer exists.

## Scope held tight

Ordinary shares keep their exact wording. Only the emergency lane is renamed — a check-in share is still a share, asserted explicitly.

## Verification

- **437 backend**, **1001 frontend** tests pass
- `tsc`, `eslint`, `ruff` clean
- Pre-existing on main and untouched: 2 in `test_one_location_maps_routes` (stale `nearby_only` stub), 2 in `one-location-onboarding-flow`


DamriaNeelesh:
Tracked in #5569 — the naming guarantee added on this branch (Location notifications were dropping the sender's name, and could render a raw identifier in its place).

PR #5610, opened 2026-08-20
fix(circles): notify inviter when a Circle invite is accepted

parthmawai:
## Summary
- `accept_member_invite()` (targeted, named Circle invites) wrote the new membership silently — the inviter's Circles screen only reflected a new member after a manual reload, because nothing told their client anything had changed.
- This is the same class of gap as the code-based join path, which already notifies the inviter via `send_circle_code_joined_push`. The targeted named-invite path never had an equivalent.
- Adds `send_circle_member_invite_accepted_push` and calls it from `accept_member_invite` on a real `pending -> accepted` transition only (not on an idempotent replay of an already-accepted invite).
- Wires the new `location_circle_member_invite_accepted` type through the existing Location notification pipeline (`WORKFLOW_COPY`, section routing, copy generation, `isOneLocationWorkflowNotificationType`) so it dispatches `CONSENT_STATE_CHANGED_EVENT` exactly the way every other Location workflow event already does. `app/one/location/page.tsx` already listens for that event and calls `refresh({ background: true })`, so no new frontend polling/plumbing is needed — this is purely "produce the missing event," not new infrastructure.
- Also fixes an adjacent, pre-existing gap in the same guard: `location_circle_code_joined` was already being sent by `send_circle_code_joined_push` (the code-based join's inviter notification) but silently dropped by `isOneLocationWorkflowNotificationType`, so that notification never actually reached the toast/refresh path either.

## Test plan
- [x] Backend: `pytest tests/services/test_one_location_circle_service.py tests/test_one_location_named_circle_routes.py` — 56 passed, including a new test asserting the push fires with the correct kwargs on accept, and that the accept-success path is otherwise unchanged.
- [x] Backend: `ruff check` clean on changed files.
- [x] Frontend: `tsc --noEmit` clean.
- [x] Frontend: `eslint` clean on changed files.
- [x] Frontend: `vitest run` on the Location notification test suites (`one-location-notifications`, `consent-notification-provider-location`, `one-location-notification-reconciliation`, `consent-center-page-location-actions`, `location-consent`) — 52 passed.
- [ ] Manual: two accounts, A invites B to a named Circle, B accepts while A has the Circles screen open — A's screen should update without a reload.


DamriaNeelesh:
@parthmawai for adding any connected member inside the circle we have decided to remove the fritction of inviting them and than that user accept that invite as the member is already connected through the connect tab.

So is there any other reason behind this PR ? or anything that is breaking ?

parthmawai:
@DamriaNeelesh Actually this is one in a series of fixes i'm doing, basically the broader issue is that certain elements doesn't update/refresh unless tab is reloaded

parthmawai:
Here is the summary:
PR #5610 fixes the specific "Circles" reload bug you confirmed as the real target for the data-snapshot work.

Root cause I traced: when someone accepts a named Circle invite (accept_member_invite() in one_location_circle_service.py), the code writes the new membership to the database but sends zero signal to the inviter — no FCM push, no event, nothing. So the inviter's Circles screen has no way to know a new member joined until they manually reload. That's a genuine bug, not a caching/timing issue — literally nothing was ever sent.

I found this by comparing against the sibling flow: joining a Circle via a shareable code already notifies the inviter (send_circle_code_joined_push) via the same Location notification pipeline that everything else in the app uses (dispatches CONSENT_STATE_CHANGED_EVENT, which app/one/location/page.tsx already listens for and refetches on). The targeted/named-invite path just never had that call wired in — likely an oversight when that flow was built, not something the recent rollback touched.

The fix adds the missing notification (send_circle_member_invite_accepted_push) and only fires it on a real accept, not on a replay of an already-accepted invite. I also caught and fixed an adjacent bug in the same guard function while I was in there: the code-join path's notification type was already being sent but was silently dropped by a frontend type-check that never recognized it — so that "someone joined via your code" case was also silently failing to live-refresh.

parthmawai:
@DamriaNeelesh I'll hold on this pr from merging if you think there is an issue with this, please let me know how can i improve

PR #5624, opened 2026-08-20
restore(one-location): bring back the ten PRs the Monday rollback reverted

DamriaNeelesh:
## What this is

PR #5603 rolled `main` back to Mon 2026-08-17 14:00 IST after work from several people had to be undone. It reverted **every** PR merged after that point except Kushal's — by author: Ankit, Abdul, Parth, **Neelesh**, Smirthika, Roopmann, anoushkauoc.

This restores **only Neelesh's ten**, replayed in merge order on top of current `main`:

| Merged | PR | |
|---|---|---|
| 17 Aug 17:01 | #5367 | onboarding finale map |
| 17 Aug 21:33 | #5442 | requester name in connection toasts |
| 18 Aug 13:23 | #5456 | map chrome above the check-in drawer |
| 18 Aug 16:45 | #5477 | 100-member Circles + the contact picker |
| 18 Aug 21:00 | #5497 | SMS review pill under the header |
| 19 Aug 05:58 | #5505 | SMS contacts as a protected system Circle |
| 19 Aug 06:18 | #5552 | SMS alert no longer revokes a normal share |
| 19 Aug 11:22 | #5560 | say WHICH share ended |
| 19 Aug 21:03 | #5578 | the SMS Circle from the other member's side |
| 20 Aug 03:01 | #5601 | Circle roster columns |

Note these PRs all still read MERGED on GitHub — the rollback landed a commit that undoes their *content*, so the merge commits stayed in history while the code did not.

## Two commits that are not Neelesh's, and why

Both are dependencies, not features. Neither can be split off without shipping something broken.

- **#5531 — migration 159** (`880a2b33c`, Ankit). The restored manifest declares 158, 159, 160 and the contract policy is `exact`, so 158 and 160 cannot land without it. Its title says what it is: *"widen the invite-code max_uses ceiling to match 100-member Circles"* — the other half of Neelesh's own 158.
- **`9a91b526b` — the system-Circle deletion guard** (Ankit). Migration 160 puts a trigger on `one_location_circles` that refuses to delete a system Circle. Without this, account **Reset and Delete 500**. Migration 160's own test asserts `SET is_system = FALSE` lives in the account service.

## Where the replay needed judgement

A clean merge is not the same as a feature arriving. Four places accepted new inputs and dropped the code that used them, and each was repaired against attribution rather than against whatever silenced the compiler:

- `SharedWithMeCard` took `isSmsTriggered` and drew nothing — an SOS share looked identical to an ordinary one. Badge restored.
- The Shared-with-me list opened a per-owner wrapper whose closing tag this base never had, leaving the JSX unbalanced.
- `grant-lanes.ts` needs `isSmsTriggeredGrant`, which arrived in a change outside this restore. Re-added as the minimum dependency, delegating to the share-kind normalizer already in that module. The upstream signature took `Pick<OneLocationGrant, "shareKind">`; here that field is optional, so the type is structural and optional instead.
- `native-route-inventory.json` — the replay appended a route block that was already present, taking it to 166 entries against its own `total_routes: 113`. Neither side intended a change and no restored PR adds an app route, so the base's inventory stands.

**Dropped rather than dragged in,** each verified by `git log -S` before removal: `focusGrantId` deep-linking and the `nameSuffix` end time (both Abdul Rashid), the compact add-minutes chips (Abdul Rashid — the only reason `onEditGrantSave` had a second argument), and one test exercising Parth Mawai's connection-accepted push.

**`main` wins every shared component it still owns.** The first pass let the older lineage overwrite work `main` has moved on with, so it was inverted: `named-circle-flows.tsx` is `main` byte-for-byte through `CirclesSection`, `CreateCircleFlow` and `JoinCircleFlow`, and only `CircleMemberRow` and `CircleDetailFlow` come from the restore. `AskFlow` returns to `main` outright — its duration editor and per-row remove are covered by nine tests that pass there, and the restored lineage predates them.

## Verification

```
tsc --noEmit                                          clean
eslint (every changed web file)                       clean
vitest components/one-location + 2 contracts          567 passed / 2 failed
  └─ the same 2 fail on a clean origin/main checkout — no new failure
pytest (every backend file this touches)              282 passed
playwright circle-member-row.layout  chromium+webkit   14 passed
verify:design-system                                  passed
verify:capacitor:static                               passed
./bin/hushh protocol lint                             passed
```

Migrations 150→160 are contiguous, each with its rollback, and the manifest matches.

## Before this deploys

**The UAT database needs 158, 159 and 160 re-applied.** PR #5603 states they were rolled back there. The app code in this PR expects that schema — a 100-member Circle and the system-Circle trigger both assume it.

Rebased onto `origin/main`, 0 behind at push.


DamriaNeelesh:
## Superseded by a sequence, not abandoned

Raised as one PR first to prove the restore integrates and passes end to end. It does — 567/569 vitest with the same 2 failures as clean `origin/main`, 282 backend tests, both layout engines, every gate green.

But 87 files across ten PRs is not reviewable in one pass, and it has two specific problems this branch cannot solve:

1. **Deploy ordering.** Migrations 158/159/160 and the code that reads them land in the same merge, so the UAT database has to be migrated by hand at exactly the right moment or the app meets a schema that is not there.
2. **Revert blast radius.** If anything misbehaves on UAT, the only rollback is all ten PRs at once.

So it is being split into eleven, each based on `main` and merged in order. Not stacked — a PR whose base is not `main` skips the test lanes here and still passes the gate, so stacking would ship this with no coverage at all.

**#5631 is the first: the schema alone**, deliberately ahead of the code, so the UAT migration happens on its own.

This branch stays open as the reference for what the sequence adds up to, and as the place the cross-file reconciliation is already worked out — the SMS badge that rendered nothing, the unbalanced JSX wrapper, `isSmsTriggeredGrant`, the route inventory that double-counted, and the decision to let `main` win every shared component it still owns.

DamriaNeelesh:
Superseded and complete. The sequence this branch was split into has fully landed:

| restore | PR | merge |
|---|---|---|
| schema first | #5631 | `b7b4849a4` |
| replay-safety | #5641 | `15584fb36` |
| 2/9 — #5442 | #5636 | `982e6a1eb` |
| 3/9 — #5477 | #5651 | `c8112e41e` |
| 4/9 — #5497 | #5653 | `d12b04030` |
| 5/9 — #5505 | #5655 | `0d517877c` |
| 6/9 — #5552 | #5660 | `58c5817aa` |
| 7/9 — #5560 | #5663 | `086583e2e` |
| 8/9 — #5578 | #5669 | `c029aa4aa` |
| 9/9 — #5601 | #5673 | `b38faf896` |

Audited file-by-file against `main`'s tracked tree afterwards: **191 of 191 files** across all twelve of the affected PRs are present. #5367 and #5456 were re-landed by others with newer implementations and were deliberately not restored again — restoring them would have regressed that work.

Splitting was the right call. It surfaced defects a single merge would have buried: a migration that was not replay-safe and blocked the whole release, a route inventory double-counted to 166 entries against its own `total_routes: 113`, a manifest test pinned to a tail that had moved, and two components the replay left half-merged.

PR #5626, opened 2026-08-20
fix(one-location): make the save-a-place surface a real bottom sheet, and stop asking for what the address already says

### LINKED ISSUE #5619: Save-location sheet renders as a floating panel and asks for unnecessary address fields
The Pin your entrance and Address details flow presents as a centred web dialog rather than a native bottom sheet. On viewports between 421px and 639px the surface is capped at 420px and centred, leaving wide empty strips on both sides so it reads as a floating pad above the app instead of a sheet attached to the bottom of the screen. The flow also uses the Radix Dialog primitive directly instead of the shared bottom-sheet component, so it has no drag-to-dismiss gesture, no drag handle and no keyboard-aware sheet behaviour. Separately, the Address details step asks for more information than the product needs: Building colour, PIN / postcode and a Fill from this address checkbox are all requested even though the postcode is already derivable from the geocoded address and the checkbox is redundant once a pin has been chosen. Expected: the surface uses the shared bottom sheet at full mobile width with correct safe-area handling and working gestures, and the form asks only for Address, House or flat, Landmark and Kind of place.

ankitkumarsingh1702:
Closes #5619

## What was wrong

The `Pin your entrance` / `Address details` flow was a Radix **Dialog** wearing a sheet's corners, carrying an unconditional `mx-auto max-w-[420px]`.

On a 375–430px phone that cap is invisible, which is how it survived. Between **421px and 639px** — a large phone in landscape, a small tablet, an iPad split view, a browser window dragged narrow — it painted a 420px card centred on a wider screen with a dead grey strip down each side. A *bottom sheet* read as a rectangular pad floating above the app.

Being a Dialog also meant **no drag-to-dismiss and no grabber**, because the dialog primitive has neither.

Every other bottom sheet in this app already gates its width the other way round — `sm:mx-auto sm:max-w-md`, `sm:max-w-md`, or no cap at all. Full width below `sm`, constrained above it, is the house pattern. This surface was the one place that inverted it.

## What changed

**Presentation.** Below 640px it renders the canonical `components/ui/sheet.tsx`: full viewport width, flush with the bottom edge, top corners from `--app-card-radius-feature`, the home indicator inside the footer's padding. Above 640px it stays the centred dialog it already was. The step rail doubles as the grabber and is the drag surface.

**Fields.** Four optional boxes became two. Two of them the sheet already knew the answer to:

| Removed | Why |
|---|---|
| `PIN / postcode` | Sat directly under a line that contained it — the sheet showed `…, Pune, Maharashtra 411045, India` and then asked for `411045`. Now derived at save time from the line **as it stands**, so retyping the address carries its postcode with it (the old field could not: it was seeded from the pin and stopped listening the moment anyone touched it). |
| `Building colour` | Almost nobody answers it and nothing downstream reads it. Still saved for places that already carry one — it rides through `initialDetails` untouched. No longer requested. |
| `Fill from this address` | Governed those two fields and nothing else. |

Left: **Address** (required, prefilled, editable), **House or flat**, **Landmark**, **Kind of place**.

## Three defects found while measuring

Each fixed at the cause, each now covered.

1. **The Address box could not be cleared.** `addressLineEditedRef.current = true` plus `setAddressLineValue("")` when the state was *already* `""` made React bail out of the re-render, so the ref flip was never reflected and the detected address snapped straight back. Select-all-delete did visibly nothing. Now one nullable state.
2. **The map could collapse to a hairline.** It is a flex item whose only child fills it with `h-full`, so its min-content height is 0; a viewport short enough to overflow the pane resolved the overflow by deleting the map. `shrink-0`.
3. **The keyboard cap double-counted.** The sheet is pinned above the keyboard by `bottom-[var(--kb-height)]`, so `92dvh - kb` subtracts it twice. On an iPhone SE with a 300px keyboard the pinned footer was pushed 3px past the bottom of its own sheet, where `overflow-y-hidden` quietly clipped it. Now 92% of what is left.

## Shared primitive

Two additive props on `components/ui/sheet.tsx`, both defaulting to today's behaviour:

- `contentDragDismiss` — opt **out** of body-origin drag dismissal, for a sheet that is a fixed frame around its own inner scroller (its `scrollTop` is permanently 0, so every downward swipe would engage and `preventDefault()` the scroll it was meant to be).
- `useSheetDragHandle()` — the grabber's pointer handlers through context, so a header that already looks like a grabber can be the drag surface instead of stacking a second 44px handle above it.

`__tests__/components/shared-sheet-consumers.contract.test.tsx` asserts that for all **eleven** consumers, exercises the real drag thresholds, and pins the full-width-below-`sm` rule the other ten already follow.

## Backend

Unchanged. `SavedLocationAddressDetails` keeps all four fields; `buildSavedLocationAddress` composes the same line. The onboarding lane only uses `details` to compose the address string; the Settings/edit lane persists the object, and `buildingColor` + `postalCode` both survive an edit round-trip (tested).

## Evidence

| Gate | Result |
|---|---|
| `tsc --noEmit` | pass |
| `eslint . --max-warnings=0` | pass |
| `save-location-modal.test.tsx` | **56 passed** (was 50) |
| `shared-sheet-consumers.contract.test.tsx` | **10 passed** (new) |
| `one-location-agent-page.test.tsx` | 100 passed |
| `npm run test:ci` | 27 failed / 4472 passed — **same 17 files failing on `origin/main`** (baseline run in a clean worktree at `ce19523f3`) |
| `npm run test:layout-contracts` | **161 chromium + 131 webkit passed** |
| `verify:surface-map`, `verify:design-system` | pass |

The browser contract now measures **320, 360, 375, 390, 430, 480, 540, 639** — including the 421–639 band nothing used to test — plus the three named iPhones and the keyboard raised, and it runs in **WebKit** as well as Chromium.

Every new contract was mutation-checked: reintroducing `max-w-[420px]` fails 4 browser cases and 2 unit cases.

`save-location-modal.test.tsx` was in no PR lane at all. It is registered in `verify:one-location` now, or none of this would run on a pull request.


ankitkumarsingh1702:
## Native verification — iOS Simulator

Built and run on the real WKWebView, not emulated.

```
xcodebuild -project ios/App/App.xcodeproj -scheme App -sdk iphonesimulator
  ** BUILD SUCCEEDED **

xcrun simctl install / launch  →  com.hushh.app: 88352  LAUNCHED
```

| Check | Result |
|---|---|
| `cap sync ios` (10 plugins) | clean |
| Xcode build | **BUILD SUCCEEDED** |
| Install + launch on iPhone 17 Pro Max sim | launched, WebView renders, no native error screen |
| Backend baked in | `consent-protocol-f2gsa4kfsq-uc.a.run.app` (UAT) — not `127.0.0.1:8000` |
| `fatal / crash / uncaught / RefererNotAllowed` in app log | **0** |

**Shipped-artifact check** — the exported bundle actually carries the change, so this is not a stale-cache pass:

```
out/_next/static/css/*.css   clamp(160px,34dvh,340px)                        ✓
out/_next/static/css/*.css   max-height:calc((100dvh - var(--kb-height,0px)) * .92)   ✓
out/_next/static/css/*.css   min(56vh,420px)                                 0 occurrences (old map height gone)
out/_next/static/chunks/*.js "Building colour" / "PIN / postcode" / "Fill from this address"   0 files
```

## What I could NOT verify, and why

The save-a-place sheet sits behind sign-in **and** the vault. The reviewer identity has no vault on UAT, and every `/one/setup` route ejects it — so this surface is not reachable by an autonomous run. The on-device walk of the flow itself needs a human login.

What stands in for it: the WebKit layout contracts measure the real exported class strings in Safari's engine at 320/360/375/390/430/480/540/639, on the three named iPhones, and with `--kb-height` raised to 300px. That covers `dvh`-inside-`clamp()`, the full-bleed width and the keyboard geometry. It does not cover the native Google Maps SDK inside the sheet, which remains unexercised.

## Merge sync

`main` moved twice during this work. PR #5625 touched two of my files and both conflicts were the shape that silently drops a test registration:

| Conflict | main's side | mine | resolved |
|---|---|---|---|
| `verify:one-location` | +2 finale specs | +2 sheet specs | union of 8 |
| webkit `testMatch` | +`one-location-ready-panel` | +`save-location-sheet` | union of 11 |

Verified mechanically that every spec `origin/main` registers survives in both files and that `test:layout-contracts` lost nothing from either half. `npm run verify:one-location` on the merged tree: **8 files, 226 tests, all passing.**

PR #5628, opened 2026-08-20
fix(location): make Your Map fill the screen and show you as yourself

### LINKED ISSUE #5622: Your Map: blank band above the map, generic pin for the current user, and an over-long privacy paragraph
The /one/location/map screen has three problems on the renderer-consent state.

1. A blank band appears above the map. The container is correct (h-[100dvh], map absolute inset-0); the band is Google's out-of-world area, caused by the fixed pre-consent camera (center lat 20, zoom 2). A zoom-2 Mercator world is 1024 px tall, so any viewport taller than about 908 px at that centre latitude cannot be covered and the screen reads as unfinished.

2. The current user is drawn with the renderer's generic pin. The product already knows who the user is and already has their avatar in the frontend, so their own position should be immediately recognisable.

3. The copy below the map is a three-sentence paragraph that explains Google Maps internals and the Nearby Check-In feature. Location Settings already states the Check-In disclosure, so the map screen does not need to repeat it.

Expected: the map reads as one full-bleed surface at every supported iPhone height, the current user is shown as their own avatar, and the supporting copy is at most two short lines.

Frontend only. No backend, API, location or consent logic changes.

ankitkumarsingh1702:
Closes #5622

Frontend only. Three defects on `/one/location/map` — only one of which was a layout bug.

## 1. The blank strip above the map was the CAMERA, not the container

The surface is `h-[100dvh]` and the renderer is `absolute inset-0`, so the box always filled the screen. What did not fill it was the picture inside.

Google draws a Mercator world `256 * 2^zoom` px tall and paints its own backdrop everywhere else. The pre-consent camera was hardcoded to `{ lat: 20, lng: 0 }, zoom: 2`. That world is **1024 px** tall with latitude 20 sitting **453.9 px** down it, so it can cover at most **~908 px** of viewport height:

| Viewport | Backdrop exposed |
|---|---|
| iPhone SE — 375×667 | 0 px |
| iPhone 15 — 390×844 | 0 px |
| iPhone 15 Pro Max — 430×932 | **12 px** |
| the reported window — 552×1080 | **86 px** |

That is why it passed review on a simulator and showed up in a screenshot: the failure is a property of the device, not of the build.

`lib/one-location/map-world-view.ts` derives the neutral view from the box the renderer was actually given — smallest zoom whose world covers it, then slide the centre latitude only as far as it must. A resize re-fits, but **only while consent is pending**; once there is a real position, the camera belongs to the person and to `focusSelfPoint`.

This is a sibling of the web `setPadding` trap already documented in `location-immersive-map.tsx`. That fix stopped z2 dropping to z1; it could not fix a starting zoom already too low for the box.

## 2. The owner was drawn with the renderer's generic pin

`@capacitor/google-maps` exposes one styling knob per marker (`tintColor`), and its `iconUrl` takes an `https:` URL or a bundled `public/` asset — neither can be a locally composed avatar (circle + keyline + ring), and iOS re-fetches it on every marker pass.

So the marker follows the pattern the map already uses for name pills: project the coordinate with `projectToMapBox` and draw HTML over the renderer. Same camera, same staleness rule, one more layer — not a second mechanism.

- Avatar source: `useEffectiveAvatarUrl`, the hook the top bar and the profile screen already use. **No new endpoint.** It paints immediately from the Firebase photo the auth context already holds; the app-owned photo replaces it if the existing identity cache resolves one.
- Fallback: the app's existing order — initials, then the profile glyph.
- The self pin **and** the "My location" pill both go away with it. Three ways of saying "you" on one coordinate is two too many.
- A tap still selects and flies to zoom 15 — now at a 44 px target instead of a pin's tip.
- A renderer that never reports a camera keeps its plain pin rather than losing the marker entirely.

Nothing about the owner reaches the renderer: not the name, not the photo URL. Asserted.

## 3. The copy was three sentences of architecture

**Before**
> Private shares open only on this device. Google Maps gets only what it needs to draw them. Nearby Check-In is separate — it starts only when you do.

**After**
> Private until you share.

(the line the final onboarding screen already uses)

- The renderer sentence described how the map is fed. Nothing on this screen changes because of it.
- The Nearby Check-In sentence explained **another feature's** consent, one screen early. It is still stated in its strong form — "Nearby" and "only" intact — on Location Settings, the surface that actually turns Check-In on.
- The consent record is unchanged: Continue still writes `GOOGLE_MAPS_RENDERER_CONSENT_VERSION`.

## Also fixed on the way past

No Maps key **and** consent not yet given rendered a bare `bg-muted` canvas with a Continue button floating on it: the loading overlay bails on `unavailable`, and the fallback waited for consent, so neither drew. The placeholder now draws whenever the map is unavailable, and the panel stops asking for consent to a renderer that cannot start.

The consent panel reaches the bottom edge with a rounded top instead of floating a rem above it, and spends the safe-area inset on padding under the button rather than on a gap under the card. Desktop dialog narrowed from 832 px to the 430 px the onboarding ready panel uses.

## Evidence

| Gate | Result |
|---|---|
| `tsc --noEmit` | pass |
| `eslint --max-warnings=0` | pass |
| `verify:one-location` | 207 pass / 7 files (was 4 — the immersive-map suite ran in no PR lane before this) |
| `test:layout-contracts` | 154 Chromium + 105 WebKit pass |
| `verify:design-system` | pass |
| `verify:surface-map` | current |

New coverage, all mutation-checked (each fix reverted, correct tests went red):

- `__tests__/one-location/map-world-view.test.ts` — reproduces the reported 86 px band under the old camera, proves 0 px under the new one across every height 320–2000 px.
- `__tests__/components/map-self-avatar-marker.test.tsx` — projection, 44 px target, off-box hiding, photo/initials/glyph fallbacks, stale ring, gesture fade.
- `e2e/one-location-map-consent-panel.layout.spec.ts` — real browser, Chromium **and** WebKit: renderer reaches y=0 at 4 heights, panel geometry at 10 widths, copy is one rendered line at 320 px, CTA inside the viewport.

Registered in `verify:one-location` and `test:layout-contracts` so they actually run on a PR.

## Not changed

Backend, API routes, database, location/geolocation logic, the location hook, map state, Circle, Check-in, auth, Lock, onboarding. Zero backend files in the diff.


ankitkumarsingh1702:
### Shared-component check: where the avatar does and does not go

`LocationImmersiveMap` is used by two routes — `/one/location/map` (Your Map) and `/one/location/check-in`. The avatar is scoped to **Your Map only**.

Check-in asks a different question — *how far am I from the place I am checking in to?* — and answers it with two pins, a connector between them, and a header legend whose swatch **is** the owner's pin colour (`SELF_TINT`). Swapping one of those two pins for a photo breaks the comparison and leaves the legend's blue dot standing for nothing on the map. Extending the avatar there means redesigning that legend too, which is not this change.

Locked by a test: `leaves the check-in map's two pins and its colour legend alone`.

The other two map surfaces are different components and are untouched by construction:

| Surface | Component | Touched |
|---|---|---|
| Your Map | `LocationImmersiveMap` | yes |
| Check-in | `LocationImmersiveMap` (`surface="check-in"`) | avatar deliberately excluded |
| Share Location / sharing status | `LiveMap` (raw `google.maps.Marker`) | no |
| Onboarding pin-your-entrance | `OnboardingLiveMap` | no |

### One disclosure on "no new backend call"

`useEffectiveAvatarUrl` is cache-first. The marker paints immediately from the Firebase `photoURL` the auth context already holds — zero network, zero added latency. The only request it can ever make is the app's **existing** identity SWR on a genuinely cold cache, the same one the top bar issues on every other signed-in screen. No new endpoint, no map-specific fetch. Worth naming rather than claiming a flat "no requests".

PR #5631, opened 2026-08-20
restore 1/11: the One Location Circle migrations, schema first

DamriaNeelesh:
First of eleven. The Monday rollback (#5603) reverted every PR merged after 2026-08-17 14:00 IST except Kushal's; this sequence brings back the ten that were Neelesh's, **one PR at a time, each based on `main`**.

Stacking was rejected deliberately: a PR whose base is not `main` skips the test lanes in this repo and still passes the gate. So each of these merges before the next is rebased.

## This PR: schema only

| | |
|---|---|
| 158 | circle member limit → 100 |
| 159 | invite-code `max_uses` ceiling, widened to match 158 |
| 160 | system Circles (SMS Contacts), with the delete-refusing trigger |

All three carry their rollback. The manifest lists 150→160 with no gap — the contract policy is `exact`, and 158 and 160 cannot be valid without 159 between them. That is why 159 travels here rather than with the feature that motivated it, even though it is not Neelesh's commit.

**Schema deliberately lands ahead of the code that reads it,** so the UAT database can be migrated before any application change arrives.

## `account_service.py` is not optional here

Migration 160's trigger refuses to delete a system Circle. Without the `SET is_system = FALSE` step ahead of the cascade, account **Reset and Delete 500**. Shipping the trigger without the guard would be shipping a known outage, so they travel together.

## What is deliberately left out

The three migration test files. Each asserts that a schema ceiling and a *service constant* agree, so they are coupled to application code arriving in later PRs of this sequence — they travel with the code they check rather than sitting red here.

Until those land, these columns are inert. That is the same state migrations 155 and 156 were left in by the rollback itself.

## Verification

```
manifest contiguous, every entry resolves to a file on disk
tests/services/test_account_service_cleanup_tables.py      14 passed
./bin/hushh protocol lint                                  clean
```

## Before deploying

**The UAT database needs 158, 159 and 160 applied.** #5603 states they were rolled back there.

## The rest of the sequence

Standalone next (no shared-file contention): #5456, #5497, #5442, #5560.
Then the ordered chain, which all touch the same mega-files: #5477 → #5505 → #5552 → #5578 → #5601.

The fully-integrated, verified end state lives on `restore/neelesh-post-rollback-work` (PR #5624) as the reference for what this sequence adds up to.


ankitkumarsingh1702:
Heads up: the restored migrations block UAT backend deploys against the current UAT database — #5640.

`158_one_location_circle_member_limit_100.sql` clears every stale `member_limit` CHECK constraint except the one it is about to create (`AND con.conname <> 'one_location_circles_member_limit_bounds'`), which is exactly the name a previous run of this migration leaves behind. UAT already carries it, so `ADD CONSTRAINT` raises `DuplicateObjectError` and the deploy rolls both revisions back (run 32390666873).

Worth checking 159's invite `max_uses` constraint for the same pattern.

PR #5638, opened 2026-08-20
fix(git): make the merge-fidelity check able to report, and cover the revert shape that actually happened

### LINKED ISSUE #5637: Merge fidelity check cannot report a finding, and only looks for one of three revert shapes
The merge-regression guard built on 2026-08-19 was never merged, and as written it could not report anything. check-merge-regression.sh builds its report with 'git log | head -5' under 'set -euo pipefail'; head closing the pipe sends git SIGPIPE, the pipeline returns 141, and set -e kills the script at the first flagged file with more than five commits of history, before any output is produced. The report is accumulated in a temp file the EXIT trap then deletes. Run against the 2026-08-20 mass revert (a60c51dfc) it exited 141 with zero output. Behind continue-on-error in CI that is indistinguishable from a pass.

Separately, the CI job only scanned the PR's own sync-merge commits, which detects a bad conflict resolution and nothing else. The revert that actually cost this repository a day left no merge commit: a60c51dfc has a single parent and sits directly on main's tip, so both a merge-commit scan and a fork-point comparison have nothing to look at.

Expected: the check completes, prints its findings, and covers a clean revert as well as a bad conflict resolution. A run that could not complete is reported as inconclusive, never as clean.
Actual: silent death at the first finding, and no coverage of the revert shape that has actually occurred here.

ankitkumarsingh1702:
Supersedes #5534, which has sat unmerged since 2026-08-19 and is red only because its base is stale. This is that work, rebuilt on current main, plus the two things that were wrong with it.

Closes #5637

## The check could not report a finding

`check-merge-regression.sh` built its report with `git log --format=… | head -5` under `set -euo pipefail`. `head` closes the pipe after five lines, `git` takes SIGPIPE, `pipefail` surfaces 141, `set -e` kills the script — at the **first** flagged file whose history is longer than five commits, before anything reaches stdout. The report was accumulated in a temp file the `EXIT` trap then deleted.

Run against the 2026-08-20 mass revert (`a60c51dfc`):

| | exit | output | files flagged |
|---|---|---|---|
| before | 141 | *(nothing)* | 0 |
| after | 0 | full report | 81 |

Behind `continue-on-error` in CI, the "before" row renders as a green check.

Fixed with `git log -n 5` instead of a pipe. The trap now dumps whatever it collected when the exit code is neither 0 nor 1, so this class of failure can never again be silent.

## It only looked for one of three ways work goes missing

The old job scanned the PR's own sync-merge commits. That finds a bad conflict resolution and nothing else. The revert that actually cost this repo a day — `a60c51dfc` — has **one parent** and sits directly on main's tip, so a merge-commit scan has no merge to walk and a fork-point comparison has a zero-width window.

`scripts/ci/check-merge-fidelity.sh` now runs three passes:

| | what it catches | leaves a merge commit? |
|---|---|---|
| **A** | conflict resolution that kept the stale side | yes — 2026-08-19 shape |
| **B** | clean revert on the branch, no conflict | no |
| **C** | revert committed straight onto the base tip | no — 2026-08-20 shape |

Pass C widens the base to "main as of 14 days ago" and narrows the file set to this PR's own diff via the new `MERGE_REGRESSION_LIMIT_TO` — a file the PR never opened cannot have been reverted by it. That is what keeps a two-week window from being slow and noisy at the same time.

## Verified against the real incident, not a synthetic one

**Positive control** — pass C's exact parameter shape over `a60c51dfc` flags 190 files, including `consent-protocol/db/release_migration_manifest.json` and `.github/workflows/deploy-production.yml`.

**Negative control** — this branch, synced: clean in 1.3s, with both later passes printing why they skipped.

**Escape hatch** — three fixtures: label + ack line (clears), neither (does not), label alone (does not).

## Two false-positive fixes the local runs forced

Passes B and C ask "the base has this content — does HEAD still?". That only means "the PR dropped it" if HEAD already **contains** the base. Run on a branch merely *behind*, every commit the base gained since the fork point answers "missing" — testing on this very branch, one hour behind main, produced 4 flagged files and ~100 missing lines from a PR that only adds shell scripts. Being behind is the Base Freshness Gate's question, and that gate blocks, so it is the authority. Both passes now require `git merge-base --is-ancestor <base> HEAD`.

The `intentional-revert` lookup was a `grep` over the raw `gh pr view` JSON, which also matches the string appearing in the PR body — where this mechanism is documented and therefore likely to be quoted. It parses the labels array now.

## Scope

Still advisory: `continue-on-error: true`, out of `ci-status`'s `needs`, `MERGE_REGRESSION_MODE=warn`. It is a text heuristic and it cannot block a merge. What it can now do is say something.

A pass that did not finish is reported **INCONCLUSIVE**, never clean — each runs under a 600s budget, and a timeout or any exit code that is not 0 or 1 says so in the job summary instead of falling through to a green check.

No application code. Nine files: five scripts, two hooks, `bin/hushh`, and the CI job.


ankitkumarsingh1702:
### Reproduction: the 30% threshold does not cover the serialized-list shape

Ran the merged `scripts/git/check-merge-regression.sh` against the two `origin/main` syncs I did in #5628 today. Both were correct unions — verified entry-by-entry that every token from main's side survived — and both are reported as **100% MISSING**, on both files, in both syncs.

```
sync 1  base=ce19523f3 upstream=a0db23f01 target=ceb4ee65e
  hushh-webapp/package.json          1/1 of upstream's added lines are MISSING (100%)
  hushh-webapp/playwright.config.ts  1/1 of upstream's added lines are MISSING (100%)

sync 2  base=a0db23f01 upstream=fb5ff1536 target=b1f7fb9e0
  hushh-webapp/package.json          1/1 of upstream's added lines are MISSING (100%)
  hushh-webapp/playwright.config.ts  1/1 of upstream's added lines are MISSING (100%)
```

**Why the threshold cannot catch this.** `THRESHOLD` is a percentage over *upstream's added lines in that file*. `verify:one-location`, `test:layout-contracts` and the WebKit `testMatch` regex each serialize a whole list onto **one** line, so upstream's contribution to the file is a single line. A correct union replaces that line with a superset — the exact text is gone, so the score is 1/1 = 100%, never a small fraction. There is no threshold value that distinguishes this from a real drop; 30 vs 40 is irrelevant when the denominator is 1.

The header comment says these cases were "confirmed-benign … a shared test-list line". That holds when the file also has many other upstream lines to dilute the score. It does not hold when the list line is the *only* thing upstream touched in that file, which is the common case — two branches each adding one spec.

**Frequency.** Four false positives from one PR. Both of the repo's highest-collision list files, twice each, because main moved twice while I was resolving. Anyone adding a spec file hits this.

**Impact.** Advisory only — `MERGE_REGRESSION_MODE=warn`, exit 0, `continue-on-error` in CI — so nothing was blocked. The cost is that the detector cries wolf on the single most common legitimate conflict in this repo, which is how a warn-mode gate gets trained into background noise before it ever catches a real revert.

**Shape of a fix, if it's wanted:** compare by *entry* rather than by line for lines that are token lists — split the removed and added line on whitespace and report only tokens absent from the result. A correct union then scores 0 missing, while a genuine drop still scores the tokens it lost. That reads naturally as an extra check inside the existing per-file loop rather than a new exclude pattern, since excluding these paths outright would give up on the files most likely to be resolved wrongly.

Not opening a PR for this — flagging it here so it stays with the work rather than becoming a second competing change.

Issue #5405, opened 2026-08-17
Share location subroute: selecting one circle triggers loading across all circles & layout clips text inputs


kushaltrivedi5:
## Bug Summary
On the Share Location subroute, two issues occur:
1. **Unscoped Loading State:** When multiple circles exist (e.g., Circle "A" and Circle "B") and Circle "A" is selected, loading spinners trigger globally across both "A" and "B" rather than isolating loading to the selected circle.
2. **Layout Clipping:** On the same page, the text input box and displayed text are cropped/clipped by container boundaries.

## Reproduction Steps
1. Open the Share Location subroute with 2 or more configured circles.
2. Tap to select Circle "A".
3. Observe both cards showing active loading states.
4. Observe the message/text input box and descriptive labels getting clipped at the viewport/container boundary.

## Expected Behavior
- **Scoped Loading:** Selection loading/pending state must be keyed to the specific `circleId` so other circle items remain in their normal state.
- **Unclipped Responsive Layout:** Container overflow, padding, and safe-area insets must ensure all text inputs, action buttons, and displayed labels are fully visible and comfortable to interact with on all device sizes.

## Acceptance Criteria
- [ ] Refactor circle selection state to track `selectedCircleId` / per-circle loading rather than a global boolean.
- [ ] Fix CSS overflow/height constraints on the Share Location surface to eliminate text and input box clipping.
- [ ] Verify fix across mobile Web, iOS, and Android.

PR #5661, opened 2026-08-20
fix(one-location): scope Share Circle loading state per circle and fix flow clipping

parthmawai:
## Summary
Addresses #5405. Re-ships work that was reverted by #5603 (the 2026-08-20 Monday-baseline UAT rollback) -- same fix as the original #5450, reapplied fresh against current `main`.

Two independent bugs on the Share Location subroute:

1. **Unscoped loading state.** `BusyState`'s `"shareCircle"` was a flat literal, not id-scoped like its siblings `` `sms-contact:${string}` ``/`` `sms-circle:${string}` `` right next to it in the same union. Selecting Circle A set that one flag, which every circle row's render read -- so every card showed "Loading..." and became disabled at once, not just the tapped one.

2. **Layout clipping.** The full-screen task-flow wrapper in `LocationRedesignHub` (used by Share and every other flow -- Ask, CheckIn, SOS, etc.) has no bottom safe-area/nav clearance. Traced via `git log -p`: `AppPageShell`'s bottom padding was intentionally dropped for the Now/People/Links tab surface when it got its own clearance mechanism (`SwipeViews` `heightMode="active"`) -- but task flows bypass `SwipeViews` entirely and were never given an equivalent, so their bottom-most content (note textarea, Start sharing/Cancel buttons) sits flush against the viewport edge on every device.

## Changes
- `page.tsx`: `BusyState`'s `"shareCircle"` -> `` `shareCircle:${string}` ``; setter now calls `` setBusy(`shareCircle:${circleId}`) ``.
- `location-redesign-hub.tsx`: both `vm.busy === "shareCircle"` comparisons in `ShareFlow`'s circle list now compare `` `shareCircle:${circle.id}` ``; the flow wrapper div gets the same bottom-clearance class already used by ~6 other full-page surfaces in this app (`kyc/page.tsx`, `gmail-receipts-page.tsx`, `developer-docs-hub.tsx`, `pkm-settings-shell.tsx`, etc.) rather than reintroducing the tab surface's now-superseded padding.

**Note on the diff**: `hushh-webapp/cache-coherence-screen-manifest.generated.json`'s diff also includes two unrelated `/one/profile/preferences/voice*` entries that were already stale on `main` before this change, swept up by the same regeneration pass this fix requires -- not an intentional change here.

## Test plan
- [x] `tsc --noEmit` clean -- confirms no other call site still compares against the old flat `"shareCircle"` literal.
- [x] `eslint` clean on both changed files.
- [x] `vitest run` on `one-location-agent-page.test.tsx`, `one-location-sos-emergency.contract.test.tsx`, `saved-locations-section.test.tsx` -- 123 passed, no regressions.
- [x] All 5 generated artifacts regenerated and confirmed at a stable `--check`-clean fixed point.
- No existing test exercises this specific circle-selection interaction (none found for either the old or the sibling `sms-circle:`/`sms-contact:` scoped pattern) -- relied on the full relevant suite passing rather than adding new integration-test scaffolding disproportionate to a 5-line fix.


parthmawai:
## Half still valid, half obsolete — recommend trimming

Two unrelated fixes share this branch and they have aged differently.

**Keep: the `shareCircle` busy-state scoping.** Still broken on `main` today — `page.tsx:827`, `:7693`, `:8389` and `location-redesign-hub.tsx:4452`, `:4469` all still compare against a bare `"shareCircle"`, so selecting one circle shows "Loading…" on every circle. This fix is still needed.

**Drop: the flow-clipping hunk.** It rewrites `pb-[calc(150px+env(safe-area-inset-bottom))]`, but `main` now reads `pb-6` — already changed, a different way.

That obsolete hunk is also what the conflict is; trimming to the busy-state change alone should make this merge clean. Full audit of this and the seven sibling PRs is in #6208.

Issue #5459, opened 2026-08-17
Business Profile Claiming: verify ownership via Phone OTP & enable user-initiated SMS invites during Check-In


kushaltrivedi5:
## Feature & Verification Architecture Summary
When checking into a place or browsing local businesses (e.g., hotels, cafes, venues), many locations may not yet have claimed their verified Hussh business profile. We need an OTP-backed business claiming flow and a user-initiated invite mechanism during Check-In.

### 1. Phone OTP Verification Identifier for Business Claiming
- Business ownership and profile administration are verified using **Phone OTP** sent to the business's registered public phone number.
- Entering the validated OTP grants the business owner administrator rights over the place profile (allowing them to manage venue details, operating hours, verified badges, and privacy rules).

### 2. User-Initiated Invite on Behalf of Hussh
- During the **Check-In** flow (or venue profile view), if the selected business/hotel is marked as `unclaimed`:
  - Provide a clean CTA: *"This business profile is unclaimed. Send an invite on behalf of Hussh to claim profile"*.
  - User can confirm the business phone number and trigger an automated, branded SMS invite.

### 3. Business Claiming Landing Flow
- The business owner receives the invite SMS with a secure claim link.
- Clicking the link triggers the Phone OTP challenge to the business number.
- Upon successful verification, the business profile transitions to `claimed: true` with a verified badge.

## Acceptance Criteria
- [ ] Implement backend OTP generation and verification for business place claiming.
- [ ] Add "Send invite to claim business" action in the Check-In flow for unclaimed places.
- [ ] Implement business claim landing page with phone OTP authentication.
- [ ] Verify SMS dispatch, OTP validation, and profile claim status transitions.

smirthi-dharma:
Opened a PR covering the Check-In CTA piece of this issue: #5715

Send invite to claim business — detects the unclaimed state during Nearby Check-In, shows the CTA, and adds a confirmation dialog before marking the invite sent. OTP verification, SMS dispatch, and the rest of the claiming system are separate follow-up work.

PR #5715, opened 2026-08-21
Add invite-to-claim CTA for unclaimed businesses during Check-In

smirthi-dharma:
## Summary
Part of #5459.

- Nearby Check-In detects that the selected place has a real, provider-confirmed unclaimed business profile and shows a card: *"This business profile is unclaimed — Send an invite on behalf of Hushh to claim profile."*
- **CTA only surfaces on a real signal.** `isPlaceUnclaimed()` requires an explicit `businessClaimStatus: "unclaimed"` field on the place (added to `OneLocationNearbyPlaceSuggestion`). No current provider populates this field, so the CTA stays hidden on real data today and activates automatically once a provider starts sending it — "unknown" is never treated as "unclaimed".
- Tapping **Send invite to claim business** opens a confirmation dialog naming the business and explaining that sending isn't live yet.
- **Confirming does not claim success.** There is no SMS dispatch backend behind this yet, so confirming only notes the request locally (`toast.message`, not `toast.success`) and shows "Noted. Sending isn't available yet." — never a false "Invite sent" state.
- Canceling leaves everything untouched.

## Explicitly out of scope for this PR
- OTP generation/verification backend
- SMS dispatch
- The business claim landing page and profile transition
- Any other acceptance criteria from #5459 beyond the Check-In CTA

## Test plan
- [x] `npx vitest run components/one-location/nearby-check-in/__tests__/nearby-check-in-sheet.test.tsx` — 50 passed (50), including 5 in the unclaimed-business-invite suite: CTA never shows on the default (no-signal) fixture, CTA stays hidden when explicitly marked `claimed`, CTA appears only when marked `unclaimed`, confirming notes the request without claiming success, canceling leaves state and toasts untouched.
- [x] `tsc --noEmit` clean on the touched files
- [x] `eslint` clean on the touched files

### DIFF
--- a/hushh-webapp/components/one-location/nearby-check-in/__tests__/nearby-check-in-sheet.test.tsx
+++ b/hushh-webapp/components/one-location/nearby-check-in/__tests__/nearby-check-in-sheet.test.tsx
@@ -55,6 +55,7 @@ vi.mock("sonner", () => ({
   toast: {
     error: vi.fn(),
     success: vi.fn(),
+    message: vi.fn(),
   },
 }));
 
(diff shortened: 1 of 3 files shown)


smirthi-dharma:
Pushed a correctness fix on top of the initial version:

- **Unclaimed detection was a false positive.** \`isPlaceUnclaimed()\` returned \`true\` for every place unconditionally, since there was no backend claim data. Now it requires an explicit \`businessClaimStatus: "unclaimed"\` field from the provider — hidden on all real data until that field exists.
- **Invite confirmation claimed false success.** The post-confirm state said "Invite sent on behalf of Hushh" with a success toast, even though no SMS dispatch backend exists. It now says "Noted. Sending isn't available yet." via \`toast.message\`, never \`toast.success\`.

50 tests passing (5 in the unclaimed-business-invite suite), tsc and eslint clean.

PR #5845, opened 2026-08-23
fix(contacts): the discovery endpoint stops being an enumeration oracle

DamriaNeelesh:
Step 1 of the Contact Sync plan. Contact sync itself works on all three platforms — this closes the security hole underneath it.

## The hole

`POST /api/marketplace/contacts/match` answers *"is the person behind this phone number on Hushh"*, a thousand numbers at a time. It shipped with **no rate limit at all** — not its own, and not a global one to fall back on, because `SlowAPIMiddleware` is never installed and `RateLimits.GLOBAL_PER_IP` is referenced nowhere outside its own test.

Every One Location route is bounded. This one — the most identity-revealing query in the product — was not. An authenticated caller could walk the national number space against the user base at whatever rate their connection allowed, and unsalted SHA-256 over a ten-digit keyspace means each answer is a phone number, not a hash. That is the attack in *"All the Numbers are US"* (NDSS'21), available to anyone with an account.

## What changed

**Two ceilings**, because one number cannot describe the shape of the abuse:

| | | |
|---|---|---|
| `CONTACT_DISCOVERY_MATCH` | `4/minute` | stops a tight loop |
| `CONTACT_DISCOVERY_MATCH_DAILY` | `20/day` | stops the patient walk — the realistic version |

Keyed **per authenticated user**, not per IP: the route requires a Firebase identity, so a caller cannot shed the limit by changing address, and an office behind one NAT does not share a budget it never spent. Syncing an address book is a deliberate act — even a heavy multi-device day does not approach twenty.

**`phone_last4` is gone from the response.** The caller derived every digest it sent from its own address book, so it already holds the number behind each match — echoing four digits back tells it nothing it did not have, and costs a real leak the moment a response is logged, cached, or rendered. Both were happening at once: the marketplace deck rendered those digits into the DOM (`app/marketplace/page.tsx:1240`) while One Location stripped them on arrival (`contact-signals.ts:134`). A field one client has to defend against and another leaks is a field that should never have left the server.

The client-side strip **stays**, now keyed on the *shape* of the value rather than one field name. A deploy is not atomic, so through a rollout — and any rollback — this client can still be talking to a server returning the old payload. Removing the field from the TS type would otherwise have silently stopped protecting exactly the case that still needs protecting.

**`cap.contact.discovery` is declared**, so the capability finally has a name the consent surface can show and an audit row can carry. Deliberately **not enforced yet** — the route authenticates with a Firebase ID token and all three shipped clients call it that way, so requiring a consent token here would break contact sync on web, iOS and Android the moment it deployed. That is a coordinated client change and should be one, not a silent outage.

## Tests and docs, where there were none

The endpoint had **zero backend tests** and was **undocumented**. It now has both. The new file is registered in `consent-protocol/scripts/test-ci.manifest.txt` — the only pytest invocation in any lane, including the merge queue, so a guard absent from it is decoration.

## Verification

| | |
|---|---|
| New `tests/test_contact_discovery_hardening.py` | **9 passed** |
| Full CI backend manifest (locally, minus one Unix-only file) | **1366 passed** |
| Contact/marketplace/observability vitest | **175 passed** (20 files) |
| `tsc --noEmit`, `eslint`, `ruff check`, `ruff format --check`, `verify:docs` | clean |

Both limits confirmed registered on the route at runtime: `4 per 1 minute`, `20 per 1 day`.

## Next in the plan

Step 2 — the pepper (HMAC over plain SHA-256), the server's hardcoded `+1` normalizer, and the candidate cap that can make a user permanently invisible to their own contacts. Then the invite/growth half, which was computed and thrown away.

### DIFF
--- a/consent-protocol/api/middlewares/rate_limit.py
+++ b/consent-protocol/api/middlewares/rate_limit.py
@@ -166,6 +166,22 @@ class RateLimits:
     # Owners can rotate/revoke a code, but rapid churn is never a normal flow.
     ONE_LOCATION_CIRCLE_MUTATION = "6/minute"  # noqa: S105
 
+    # Contact discovery. This route answers "is the person behind this phone
+    # number on Hushh", a thousand numbers at a time, and it was the only
+    # identity-revealing route in the product carrying no limit at all --
+    # neither its own nor a global one, because SlowAPIMiddleware is never
+    # installed and GLOBAL_PER_IP is referenced nowhere outside its own test.
+    # An authenticated caller could walk the national number space against the
+    # user base at whatever rate their connection allowed.
+    #
+    # Two ceilings, because one number cannot describe the shape of the abuse.
+    # The per-minute bound stops a tight loop; the daily bound stops a patient
+    # one, which is the realistic version of this attack. A person syncing
+    # their address book does it once, and again occasionally after adding
+    # someone -- even a heavy multi-device day does not approach twenty.
+    CONTACT_DISCOVERY_MATCH = "4/minute"  # noqa: S105
+    CONTACT_DISCOVERY_MATCH_DAILY = "20/day"  # noqa: S105
+
     # The owner's own position heartbeat onto their live public link. Unlike
     # every other mutation on this router this one is SUPPOSED to repeat: the
     # web client publishes on a twenty-second heartbeat plus a movement watch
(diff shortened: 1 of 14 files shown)


ankitkumarsingh1702 (approved):

PR #5848, opened 2026-08-23
fix(contacts): say why the person moved up the list

DamriaNeelesh:
Step 3 of the Contact Sync plan. Independent of [#5845](https://github.com/hushh-labs/hushh-research/pull/5845) — no overlapping files.

## The contact scan's entire visible output was an unexplained sort

A matched contact gets a **+8 recommendation boost**, so they jump the People list. `enrichRecipientsWithContactSignal` (`app/one/location/page.tsx:958`) sets an `"In your contacts"` reason *and* a summary to explain that jump.

Neither ever reached the screen for most people. The summary is written as:

```ts
recommendationSummary: recipient.recommendationSummary || "Matched from your private mobile contact scan"
```

— it only fills when the server left it **empty**. And `recipientRecommendationLine` reads the summary first. So for anyone the server already had a line about (which is most recipients), the scan reordered the list and **nothing anywhere said why**.

The subtitle now prefers the contact reason outright, read off `recommendationReasons` rather than the summary — the reason is the fact, the summary is prose about it. Nothing is overwritten; the server's own line still serves everyone else.

## Two more on the same surface

**A picker refusal stopped blaming the platform.** `contacts-web.ts:88` collapsed every non-`AbortError` rejection into *"Contacts are only available in the native mobile app"*. The three ways `navigator.contacts.select` rejects are a call outside a user gesture, an insecure context, and a cross-origin iframe — all things the page did, none of them a browser lacking contacts. Telling somebody whose browser *does* support contacts to install an app is advice that cannot help them. The genuine "no picker here" path keeps its message.

**"Find contacts" became voice-addressable.** It was the only contact control with no `data-voice-control-id` and no entry in `page.voice-action-contract.json` — Kai could open the invite screen and add connections, but not run the scan.

## `contacts-web.ts` had no test at all

The only path that runs in a browser, and the one whose decisions are what desktop and iOS Safari users experience as the feature being *absent* rather than broken. Eight now, including the two that pin `unavailable` vs `denied` (they drive different remedies — a Settings button on desktop opens nothing) and the one that pins a dismissed sheet as an empty pick rather than a failure.

## Verification

| | |
|---|---|
| New `contacts-web.test.ts` | **8 passed** |
| `verify:one-voice` | **29 passed** |
| `tsc --noEmit`, `eslint` | clean |
| surface-map / route-orchestration / cache-coherence / topology `--check` | all current |
| `verify:docs`, doc brand check | pass |

The regression guard for the main fix is an added assertion in the existing contact-sync integration test: after a sync, the People list must show **"In your contacts"**. Before this change it rendered the generic summary instead.

## Note on the commits

Three: the fix, one unrelated `ruff format` drift in `test_one_adk_live_protocol.py` that the consent-protocol pre-commit hook blocks every commit on (third time this session — not this branch's code), and one regenerating the two content-hashed artifacts that conflicted on rebase, resolved by re-running the generators rather than hand-merging.


ankitkumarsingh1702 (approved):

ankitkumarsingh1702 (approved):
Approved for governed queue after full-Web contract refresh. Exact pushed head has focused local 70/70 and PR Validation is running.

ankitkumarsingh1702 (approved):
Approved after targeted One Location rerun: 29 files, 598/598; focused full-Web failure pack 70/70.

PR #5849, opened 2026-08-23
fix(location): the chat clamp test was asserting the value the clamp removes

DamriaNeelesh:
**This test is red on `main` right now.** It broke in #5830 — mine — and went unnoticed because `__tests__/components/use-location-chat.test.tsx` is in none of the path-scoped vitest packs a PR runs.

## What happened

#5830 clamped the chat path's `create_public_link` duration:

```ts
durationHours: Math.min(PUBLIC_LINK_MAX_DURATION_HOURS, Math.max(PUBLIC_LINK_MIN_DURATION_HOURS, Number(action.durationHours) || 1))
```

Because `CreatePublicInviteRequest.durationHours` is `gt=0, le=1` and the service refuses anything above `PUBLIC_INVITE_MAX_DURATION_HOURS` — 24 was the *private* share ceiling, copied. That is the one route to `createPublicInvite` which does not pass through the screen's own duration control, and its value arrives from a **model-authored directive**, so an unclamped number reached an endpoint that would 422 it *after the person had already said yes*.

The test still asserted `durationHours: 2` — the exact number the clamp exists to remove.

## The fix

Corrected, and given the coverage it should have had in the first place: a table of four proposals across both bounds.

| proposed | expected | why |
|---|---|---|
| `2` | `1` | above the public ceiling |
| `24` | `1` | the private-share ceiling, copied — the original bug |
| `0.1` | `0.25` | below the shared minimum |
| `0.5` | `0.5` | already inside the range, untouched |

So the next person to touch the clamp finds out from a test rather than from a 422.

## Verified

`use-location-chat.test.tsx` — **22 passed** (was 17 passed / 1 failed).

I also ran the **entire** web suite (5024 tests) and diffed the failures against a worktree at `cd8688fc9`, the commit before #5830 merged, to be sure this was the only thing I broke. It was. Details in the PR discussion below.

### DIFF
--- a/consent-protocol/db/connection.py
+++ b/consent-protocol/db/connection.py
@@ -279,8 +279,7 @@ def _get_acquire_timeout_seconds() -> float:
 def _pool_exhausted_error(elapsed: float) -> "DatabaseUnavailableError":
     """Build the 503 raised when our own acquire deadline actually elapsed."""
     logger.warning(
-        "db.pool_acquire_timeout: no connection available after %.1fs "
-        "(DB_POOL_MAX_SIZE=%s)",
+        "db.pool_acquire_timeout: no connection available after %.1fs (DB_POOL_MAX_SIZE=%s)",
         elapsed,
         os.getenv("DB_POOL_MAX_SIZE", "<default>"),
     )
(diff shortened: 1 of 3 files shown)


ankitkumarsingh1702 (approved):

PR #5850, opened 2026-08-23
feat(contacts): let the people who are not on Hushh be invited

DamriaNeelesh:
Step 4 of the Contact Sync plan — the growth half, which was computed and thrown away.

## A contact scan answers two questions and the product only ever asked one

`inviteCandidateCount` — how many of your contacts are **not** here — has been computed on every sync since `contact-signals.ts` shipped, and read by nothing except an analytics dimension (`page.tsx:6316`, `:6405`, `:6458`). The app learned *"forty of your contacts are not on Hushh yet"*, recorded it, and told the person nothing. There was **no per-contact invite anywhere**.

`describeContactSyncOutcome` now names them and offers the invite:

| state | title | description | action |
|---|---|---|---|
| full read, some matched | `2 people added as a contact signal.` | `48 contacts are not on Hushh yet.` | **Invite them** |
| full read, none matched | `No One users matched from this contact scan.` | `30 contacts could be invited.` | **Invite them** |
| everyone already here | *(count)* | — | none |
| partial read | *(scoped count)* | *(scoped)* | **Check more** / **Open Settings** |

`invite` only ever appears where `remedy` would have been `null`. A partial read keeps its own remedy — widening the read is a better next step than inviting out of a list the person has not finished choosing from, and the toast has one action slot.

## It shares the referral link, not the generic one

This is the whole difference between a share that is attributed and one that is not. `/r/<slug>` resolves server-side and records the attribution **before** the recipient reaches sign-in, so nothing downstream has to trust a slug the client hands back afterwards. `buildInviteToOneShare` is a bare origin with no token — correct for its own purpose, and unattributable — so it stays only as the fallback for the case the referral system cannot serve (no link yet, or the summary call fails). Better a share that works and is not counted than a dead control.

`rememberLocationInviteSource("contact_sync")` is also finally written. That value has existed in `GrowthLocationInviteSource` (`lib/observability/events.ts:63`) since the enum shipped and had **never once been recorded**, so every person who arrived because a friend's contact scan found them was attributed to nowhere. First-touch wins inside the helper, so it cannot overwrite an earlier, truer source.

## No schema change

"Who did I invite, and who joined since" needs a table and is deliberately **not** here. The loop works without it, and the table can be designed once there is traffic to shape it — rather than guessing at a shape now and carrying a migration for it.

## Verification

| | |
|---|---|
| `verify:share-ladder` | **62 passed** (5 files) |
| `__tests__/lib/one-location/` + agent page | **277 passed** (17 files) |
| `tsc --noEmit`, `eslint` | clean |

Five new cases in `contact-sync-outcome.test.ts` cover both invite states, the "everyone is here" silence, singularisation, and — the one that matters most — that `invite` never takes the action slot from a remedy that widens a partial read.

Depends on nothing; independent of #5845 and #5848.

---

## Added after review of the finished flow

**A dismissed picker no longer gets answered with a result it never had.**
Closing the web Contact Picker was reported as:

> None of the 0 contacts you shared are on Hushh yet

A sentence shaped like a finding, reporting one nobody asked for, on the most
common way out of the flow. `contacts-web.ts` deliberately turns an `AbortError`
into an empty read — dismissing is not an error — but that empty read then
travelled down the partial-read branch, which assumes a subset was *chosen* and
phrases the count as a search result. Zero shared and zero matched are not the
same fact, and only one of them is worth telling somebody.

iOS limited access with nothing selected lands in the same place, so both are
handled, each with the remedy its surface can actually act on: the picker on web,
Settings on iOS, where a sticky empty subset is widened.

Verified failing before the fix, not merely passing after — reverting the branch
produces `expected 'None of the 0 contacts you shared are...' not to match /are
on Hushh yet/`. The assertions are written against the *shape* of the mistake (no
`0 contacts`, no `are on Hushh yet`) rather than only pinning the new string, so
rewording the copy later cannot quietly reintroduce a count-of-nothing.

### DIFF
--- a/consent-protocol/db/connection.py
+++ b/consent-protocol/db/connection.py
@@ -279,8 +279,7 @@ def _get_acquire_timeout_seconds() -> float:
 def _pool_exhausted_error(elapsed: float) -> "DatabaseUnavailableError":
     """Build the 503 raised when our own acquire deadline actually elapsed."""
     logger.warning(
-        "db.pool_acquire_timeout: no connection available after %.1fs "
-        "(DB_POOL_MAX_SIZE=%s)",
+        "db.pool_acquire_timeout: no connection available after %.1fs (DB_POOL_MAX_SIZE=%s)",
         elapsed,
         os.getenv("DB_POOL_MAX_SIZE", "<default>"),
     )
--- a/consent-protocol/tests/test_one_adk_live_protocol.py
+++ b/consent-protocol/tests/test_one_adk_live_protocol.py
@@ -97,18 +97,24 @@ async def test_close_quietly_swallows_a_close_failure_on_an_already_gone_client(
 
 @pytest.mark.asyncio
 async def test_runtime_bootstrap_accepts_only_the_authenticated_byok_frame():
-    mode, credential, transport, project, location, _resume, _voice = (
-        await _receive_runtime_bootstrap(
-            _BootstrapSocket(
-                {
-                    "type": "runtime_bootstrap",
-                    "runtime_credential_mode": "byok",
-                    "runtime_credential": "test-key",
-                    "runtime_credential_transport": "developer_api",
-                }
-            ),
-            uid="user_1",
-        )
+    (
+        mode,
+        credential,
+        transport,
+        project,
+        location,
+        _resume,
+        _voice,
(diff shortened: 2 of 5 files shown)


ankitkumarsingh1702 (approved):

PR #5851, opened 2026-08-23
fix(contacts): two silent ways a real person went unfound

DamriaNeelesh:
Step 2 of the Contact Sync plan — the correctness half. Both bugs return an ordinary empty list, so nothing on any screen could tell an honest *"none of your contacts are here"* from a broken one. Neither had a test.

## 1. The normalizer invented a country

```python
if len(digits) == 10:
    return f"+1{digits}"   # ← gone
```

Same bug the **client** normalizer removed and documented (`lib/contacts/phone-normalization.ts:11-15`) — and worse on this side. On the client a wrong guess only produced a digest that missed. Here it **fabricates an identity**: a stored `9876543210` on an Indian account hashed as `+19876543210`, so a requester holding the real US number `+19876543210` was told *that stranger is on Hussh*.

A guess that can disclose somebody's membership to a person who does not know them is not a fallback.

Every writer of `actor_identity_cache.phone_number` in this repo stores E.164 with the leading `+` — the Firebase mirror, the token claim path, and the UAT test-confirm path (which forces a `+` unconditionally) — and migration 047 added the column with **no backfill**. So the branch should have been unreachable.

Rather than assert that from reading the writers, a `logger.warning` carrying the **digit count and never the number** now proves it from production traffic. A row predating all of them would go *unmatched* rather than *wrongly matched* — a miss, never a false positive about a stranger.

## 2. The candidate cap decided who was findable

```python
candidate_row_cap = min(max(limit_safe * 50, 500), 5000)   # ← keyed off the RESULT limit
```

Applied as `LIMIT` after `ORDER BY aic.user_id ASC` — an ordering the query's own comment calls arbitrary — and **anything past it was never digest-compared**.

A last4 bucket holds roughly one ten-thousandth of the user base, so a full address book of 1000 buckets expects `users / 10` candidates:

| users | expected candidates | old cap |
|---|---|---|
| 10,000 | 1,000 | fine |
| **50,000** | **5,000** | **at the wall** |
| 100,000 | 10,000 | half of them never compared |

Past that, a person whose Firebase uid sorts late was dropped from *every* large lookup — deterministically and permanently. They could be invisible to their own contacts forever, with nothing anywhere to see.

Now `min(max(len(last4_values) * 50, 500), 50_000)` — sized against the buckets actually asked for. Still bounded: `phone_lookups` is capped at 1000 by the route, so no request can pull more than 50k rows.

## What I deliberately did NOT do

Pushing the digest comparison into SQL (`encode(digest(...,'sha256'),'hex') = ANY(...)`) is the more elegant fix and would remove the cap's relevance entirely. **I left it alone.** It is not verifiable without a live Postgres, and a subtly wrong predicate there filters *everything* out — a worse version of the bug being fixed. The Python comparison stays the only thing that decides a match.

## Verification

| | |
|---|---|
| New `tests/test_contact_match_correctness.py` | **11 passed** |
| Full CI backend manifest (locally, minus one Unix-only file) | **1368 passed** — identical failure set to the pre-change baseline |
| `ruff check`, `ruff format --check` | clean |

Registered in `scripts/test-ci.manifest.txt` — the only pytest invocation in any lane, so a guard absent from it is decoration.

One of the two commits is the recurring `ruff format` drift in `test_one_adk_live_protocol.py` — fourth time this session, not this branch's code, and the pre-commit hook blocks every commit until it is flattened.

Independent of #5845, #5848, #5849, #5850.

### DIFF
--- a/consent-protocol/db/connection.py
+++ b/consent-protocol/db/connection.py
@@ -279,8 +279,7 @@ def _get_acquire_timeout_seconds() -> float:
 def _pool_exhausted_error(elapsed: float) -> "DatabaseUnavailableError":
     """Build the 503 raised when our own acquire deadline actually elapsed."""
     logger.warning(
-        "db.pool_acquire_timeout: no connection available after %.1fs "
-        "(DB_POOL_MAX_SIZE=%s)",
+        "db.pool_acquire_timeout: no connection available after %.1fs (DB_POOL_MAX_SIZE=%s)",
         elapsed,
         os.getenv("DB_POOL_MAX_SIZE", "<default>"),
     )
--- a/consent-protocol/hushh_mcp/services/ria_iam_service.py
+++ b/consent-protocol/hushh_mcp/services/ria_iam_service.py
@@ -8922,14 +8922,44 @@ class RIAIAMService:
 
     @staticmethod
     def _normalize_contact_phone_for_hash(value: Any) -> str | None:
+        """A stored phone number as the E.164 string the client hashed.
+
+        No country guessing. This used to read a bare ten-digit number as North
+        American -- `if len(digits) == 10: return f"+1{digits}"` -- which is the
+        same bug the client-side normalizer removed and documented at
+        `lib/contacts/phone-normalization.ts:11-15`.
+
+        It is worse here than it was there. On the client a wrong guess only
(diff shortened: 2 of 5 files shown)


ankitkumarsingh1702 (approved):

PR #5852, opened 2026-08-23
docs(contacts): the Play declaration this app now needs, and why we qualify

DamriaNeelesh:
Step 5 of the Contact Sync plan. Docs only — no code, no behaviour change.

## Why now

`READ_CONTACTS` is declared in `AndroidManifest.xml:59` and read by the first-party `HushhContacts` plugin, so this app is in scope for Google Play's **Contact Permissions policy** of 15 April 2026. **Nothing in this repo covered it** — I grepped `ship-android-playstore.md` and the workflow for `data safety` / `contacts` / `permission` and the only hits were unrelated.

| When | What |
|---|---|
| 15 Apr 2026 | Policy announced |
| **Sept 2026** | Play Console prompts for declarations |
| **Jan 2027** | Mandatory; non-compliant apps subject to removal |

## The load-bearing fact

**Our use case is on Google's approved list.** They name *"friend matching with server-side processing"* explicitly, and that is exactly what contact sync does. The declaration has to be read from **that** sentence, not from "we sync contacts".

The distinction matters: **inviting and referring do not justify the permission** and must use the system picker. Ours already is picker-driven — the share offered after a scan carries no contact data at all, only the sender's referral link.

## The Data safety form, field by field

A mismatch between declared and actual behaviour is a primary removal trigger, so the doc writes out the honest answers — which are unusually good ones:

| | |
|---|---|
| Collected | phone numbers, as **SHA-256 digests + last four digits** |
| Raw numbers / contact names | **never leave the device** |
| Stored | **nothing** — `match_marketplace_contacts` performs zero writes; a contact who is not a Hussh user leaves no trace on any server |
| Shared with third parties | no |
| Optional | yes — the flow works fully for someone who grants nothing |

## Prominent disclosure

Play requires an in-app disclosure **before** the prompt, inside the app, not buried in a menu. The onboarding contacts screen already satisfies this, and the doc records *why* so the next person to move that copy knows what they would break: the privacy line renders **above** the button that triggers the OS prompt, and the prompt fires **on tap rather than on mount**.

## iOS counterpart, flagged not fixed

`PrivacyInfo.xcprivacy` declares `PhoneNumber` and `Name` but not `NSPrivacyCollectedDataTypeContacts`. Whether hashes-only egress counts as "collecting contacts" is a real question with a defensible answer either way — so the doc records it as a decision to make rather than making it. `release-ios-appstore.md` already calls the privacy-manifest reconciliation publish blocker #1; this is part of it.

## Verification

`verify-doc-brand.cjs`, `verify:docs`, and `bin/hushh docs verify` all pass.

### DIFF
--- a/consent-protocol/db/connection.py
+++ b/consent-protocol/db/connection.py
@@ -279,8 +279,7 @@ def _get_acquire_timeout_seconds() -> float:
 def _pool_exhausted_error(elapsed: float) -> "DatabaseUnavailableError":
     """Build the 503 raised when our own acquire deadline actually elapsed."""
     logger.warning(
-        "db.pool_acquire_timeout: no connection available after %.1fs "
-        "(DB_POOL_MAX_SIZE=%s)",
+        "db.pool_acquire_timeout: no connection available after %.1fs (DB_POOL_MAX_SIZE=%s)",
         elapsed,
         os.getenv("DB_POOL_MAX_SIZE", "<default>"),
     )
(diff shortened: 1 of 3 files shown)


ankitkumarsingh1702 (approved):

PR #5863, opened 2026-08-24
feat(contacts): Google Contacts for the browsers that have no address book

DamriaNeelesh:
Ships **dark**. `googleContactsAvailability()` returns `"unconfigured"` without `NEXT_PUBLIC_GOOGLE_OAUTH_CLIENT_ID`, which is set in **no environment** — so behaviour on web, iOS and Android is byte-identical to today until the console work behind it is done.

## The gap it closes

`navigator.contacts.select` ships enabled by default in **Chrome on Android and nowhere else**. iOS Safari has it behind a flag; no desktop browser has it at all. On those, "Find contacts" had nothing to read and correctly said so — the onboarding step removed itself entirely.

A Google account is not a device capability. It works everywhere a browser does.

## The People API is called from the browser, never from us

A token minted by Google Identity Services, held in module memory, used for one read, discarded. Our servers never see it and never see a contact.

**The alternative is genuinely tempting and that is the danger.** A backend implementation is ~20 lines, reuses the working OAuth refresh path, and needs no public client id, no JS origins, no Capacitor gate. It would read as a *simplification* in review. It also puts the phone number of **every non-user in somebody's address book** onto our infrastructure.

So the invariant is now a test. `consent-protocol/tests/test_contacts_never_reach_the_server.py` asserts:
- no backend module references `people.googleapis.com`
- nothing asks `GoogleConnectionService.access_token` for the contacts scope
- the matching path still contains no `INSERT`/`UPDATE`/`DELETE`

Until now that property lived only in comments, and comments do not fail a build.

**Why not `access_token` specifically.** It refreshes with **no `scope` parameter** while `start()` sets `include_granted_scopes: "true"` — so the stored refresh token accumulates the union of every scope the person ever granted. Anyone who connected Calendar with `manage` has `calendar.events` **write** in that token. Handing it to JavaScript to read an address book hands the browser the ability to cancel their real meetings. Narrowing on the refresh grant *fails open*: if Google ignores the narrowed scope you get a broader token, not an error.

## One pipeline, not two

`buildMarketplaceContactLookups` gained one optional `source`. Normalization, dedupe, mobile-first ordering, the 1000 cap and the hashing are untouched and still have exactly one implementation.

Three details that are load-bearing and not obvious:

| | |
|---|---|
| `defaultRegion: null` | **Better** than the picker path, not merely different. The web picker passes a browser locale, so a US-locale browser overrides an Indian account's own verified number when deciding what `9876543210` means — the exact bug `phone-normalization.ts` was written to kill. Null lets the account's real number win. |
| `limited: false` | A People read is the whole address book. That flag is the sole gate on the partial-read copy and its **"Check more"** remedy, whose action would re-read and return the identical set. |
| `canonicalForm` **unused** | Google returns its own E.164 beside the typed value. Using it is the obvious shortcut and it is wrong: both sides of the hash must reach byte-identical E.164 through **one** implementation. Google's parser and libphonenumber-js need not agree at the edges, and the moment they disagree the digest misses silently and the person is told nobody matched. |

## A fallback, not a second button

It fires only where the device address book is unavailable. The device book stays primary — it is the person's actual phone contacts rather than whichever of them Google happens to hold, and it needs no consent sheet. **No control is added to any screen.**

Native returns `"unconfigured"` unconditionally: `capacitor.config.ts` sets `iosScheme: "App"`, so the shell origin is `App://localhost`, which Google will not accept as an Authorized JavaScript Origin — and native already has the real address book.

## Verification

| | |
|---|---|
| New `google-people-source.test.ts` | **13 passed** |
| New `test_contacts_never_reach_the_server.py` | **3 passed** |
| `__tests__/lib` + `lib/contacts` + `lib/capacitor/plugins` | **616 passed** (82 files) |
| `tsc --noEmit`, `eslint`, `ruff` | clean |

The new tests pin the request shape itself: `personFields` is exactly `names,phoneNumbers` (no photos, no addresses — those belong to people who are not our users), `sources=READ_SOURCE_TYPE_CONTACT` (saved contacts, not profiles Google infers from mail traffic), `requestSyncToken` is never sent (a durable handle to an address book we have nowhere to keep), the token never appears in a URL, and nothing but a `HushhContactsReadResult` leaves the module.

## Three human steps before it can be switched on — none of them code

1. **Read the OAuth consent-screen publishing status for `hushh-pda`.** `deploy/.env.backend.uat:144` is a **localhost** callback in a deployed UAT env, which is what a screen still in *Testing* looks like. If so, nothing ships to real users on any Google scope today — Gmail included — and turning contacts on means a first production submission that necessarily includes restricted `gmail.readonly` and its CASA assessment. Contacts did not cause that debt but is what makes you pay it. **Either way the feature works today for allowlisted test users**, so dogfooding is not blocked.
2. **Enable `people.googleapis.com`** and register **Authorized JavaScript Origins** on the OAuth client. Keep the browser client in the same Cloud project — a separate project restarts consent-screen verification from zero.
3. **Plumb `NEXT_PUBLIC_GOOGLE_OAUTH_CLIENT_ID`** through Secret Manager, `frontend.cloudbuild.yaml`, the Dockerfile `ARG`/`ENV`, and `verify-env-secrets-parity.py` — or parity CI will not catch a miss and a user will find a dead button.

One commit is the recurring `ruff format` drift in `test_one_adk_live_protocol.py` — fifth time this session, not this branch's code, and the pre-commit hook blocks every commit until it is flattened.

---

## Added after review of the finished flow

**Closing the consent sheet no longer reports a failure.** The sheet can be
closed, and that path had no handling: the rejection landed in the same catch as
a real failure, so changing your mind produced a red error toast, a card stuck on
"scanning", and an analytics row saying the sync failed. Desktop and iOS Safari
have no Contact Picker at all, so Google is not a nicety there — it is the only
contact source those browsers have, which makes its consent sheet the single most
likely place for somebody to back out.

`isGoogleContactsConsentCancelled` ships beside the code that creates the
condition, because `error_callback` stamps `name = "AbortError"` on a dismissal
precisely so a caller can tell the two apart. Only a cancellation is swallowed;
anything else still rethrows.

Verified failing before the fix, not merely passing after — reverting the
try/catch produces `expected "vi.fn()" to not be called at all, but actually been
called 1 times` on `toast.error`. The page test mocks only the Google call and
pulls the predicate through `importOriginal`, so the real one is exercised rather
than a copy that would keep passing after it broke.

**`docs/guides/google-contacts-enable.md`** now carries the three console steps
in runnable form, with the ordering hazard first: Cloud Build resolves every
`availableSecrets` entry before the build starts, so referencing the secret
before creating it fails the frontend build in all three environments, on every
branch. The access note in it is measured rather than assumed —
`testIamPermissions` against `hushh-pda` and `hushh-pda-uat` returns
`resourcemanager.projects.get` and nothing else for an ordinary member
credential, while secret *names* still list fine, which reads like access and is
not.

### DIFF
--- a/consent-protocol/db/connection.py
+++ b/consent-protocol/db/connection.py
@@ -279,8 +279,7 @@ def _get_acquire_timeout_seconds() -> float:
 def _pool_exhausted_error(elapsed: float) -> "DatabaseUnavailableError":
     """Build the 503 raised when our own acquire deadline actually elapsed."""
     logger.warning(
-        "db.pool_acquire_timeout: no connection available after %.1fs "
-        "(DB_POOL_MAX_SIZE=%s)",
+        "db.pool_acquire_timeout: no connection available after %.1fs (DB_POOL_MAX_SIZE=%s)",
         elapsed,
         os.getenv("DB_POOL_MAX_SIZE", "<default>"),
     )
--- a/consent-protocol/scripts/test-ci.manifest.txt
+++ b/consent-protocol/scripts/test-ci.manifest.txt
@@ -74,6 +74,15 @@ tests/test_capability_health.py
 tests/test_managed_vertex_readiness_script.py
 tests/test_verify_uat_release.py
 
+# The privacy property contact sync is built around, as code instead of prose.
+#
+# A raw phone number belonging to somebody who is NOT a Hushh user must never
+# reach a Hushh server. Adding Google Contacts is ~20 lines if the backend calls
+# the People API — it reuses the working OAuth refresh path and needs no client
+# id, no JS origins, no Capacitor gate. It would read as a simplification in
+# review. This is what makes it a failing test instead of a judgement call.
+tests/test_contacts_never_reach_the_server.py
+
 # One Location Circles, and the migration guards that protect them.
 #
 # Every file below was written to lock a behaviour that had already broken
@@ -90,12 +99,3 @@ tests/test_one_location_system_circle_kinds_migration.py
 tests/test_migrations_are_replay_safe.py
 tests/services/test_account_service_cleanup_tables.py
 tests/test_pool_acquire_is_bounded.py
-
-# The privacy property contact sync is built around, as code instead of prose.
-#
-# A raw phone number belonging to somebody who is NOT a Hushh user must never
-# reach a Hushh server. Adding Google Contacts is ~20 lines if the backend calls
-# the People API — it reuses the working OAuth refresh path and needs no client
-# id, no JS origins, no Capacitor gate. It would read as a simplification in
-# review. This is what makes it a failing test instead of a judgement call.
-tests/test_contacts_never_reach_the_server.py
--- a/consent-protocol/scripts/test-ci.manifest.txt
+++ b/consent-protocol/scripts/test-ci.manifest.txt
@@ -67,13 +67,6 @@ tests/test_ria_dossier_mail_contract.py
 tests/test_ria_claim_location.py
 tests/test_ria_brochure_profile.py
 
-# Release-gate classification. These guard the rule that a third-party outage
-# must not block a healthy release, and that a candidate defect still must.
-tests/test_dependency_health_classifier.py
-tests/test_capability_health.py
-tests/test_managed_vertex_readiness_script.py
-tests/test_verify_uat_release.py
-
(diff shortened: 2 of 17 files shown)


ankitkumarsingh1702 (approved):

PR #5873, opened 2026-08-24
style(consent-protocol): make the subtree ruff-format-clean so pushes work again

DamriaNeelesh:
Two files on `main` fail `ruff format --check`, and between them they block **every push from every branch in this repo** — including branches with no Python in them.

| File | Drift | Came from |
|---|---|---|
| `consent-protocol/db/connection.py` | acquire-timeout warning split across two adjacent string literals | #5869 |
| `consent-protocol/tests/test_one_adk_live_protocol.py` | seven-name tuple unpack the formatter wants one-per-line | pre-existing |

Neither is a behaviour change.

### Why this is not cosmetic

`consent-protocol/ops/monorepo/pre-push.sh:293` runs:

```sh
"$LINT_PYTHON" -m ruff format --check .
```

over the **whole subtree**, unscoped. One drifted file anywhere under `consent-protocol/` fails the push of any branch. That is the state `main` has been in.

### Verification

- `ruff format --check .` → `972 files already formatted` (was `1 file would be reformatted`)
- `ruff check .` → `All checks passed!`
- `pytest tests/test_one_adk_live_protocol.py` → **46 passed** after the reformat
- Local ruff is `0.15.11`, the version pinned in `requirements-dev.txt:26` — so this is not a version-skew artefact

### The thing actually worth fixing, separately

CI merged both of these while the local hook rejects them. So the backend lint lane is either not running `ruff format --check` at all, or running it scoped to changed paths while the hook runs it unscoped. As long as those two disagree, a green `main` can keep blocking every developer's push, and this PR will be needed again. Filing that as its own issue rather than widening this one.

### DIFF
--- a/consent-protocol/db/connection.py
+++ b/consent-protocol/db/connection.py
@@ -279,8 +279,7 @@ def _get_acquire_timeout_seconds() -> float:
 def _pool_exhausted_error(elapsed: float) -> "DatabaseUnavailableError":
     """Build the 503 raised when our own acquire deadline actually elapsed."""
     logger.warning(
-        "db.pool_acquire_timeout: no connection available after %.1fs "
-        "(DB_POOL_MAX_SIZE=%s)",
+        "db.pool_acquire_timeout: no connection available after %.1fs (DB_POOL_MAX_SIZE=%s)",
         elapsed,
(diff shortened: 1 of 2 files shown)


ankitkumarsingh1702 (approved):

PR #5931, opened 2026-08-24
feat(welcome): fix hero clipping on "/", rebuild it on React Bits glass

RudraO2:
## The bug

The signed-out welcome screen clips its own content.

`IntroStep.module.css` pinned the composition to `block-size: 100svh` on a shell with `overflow: hidden`, but wrote its vertical rhythm in fixed pixels. On any viewport shorter than those pixels assume — short laptops, and every phone in landscape — the excess is cut rather than absorbed:

- the brand mark lands on top of the 🤫 quiet mark;
- "Nothing moves without your yes." is sliced in half by the CTA.

Reproduced on `uat.one.hushh.ai` at both 1440×900 and 390×844. This is the first screen a visitor sees before being asked for a phone number.

## Layout fix — the actual bug

- Every vertical step and display size is now `clamp(floor, <viewport unit>, ceiling)`, so the column compresses to the height it actually has instead of overflowing it. **The page still fits on exactly one screen and still does not scroll.**
- The hero is bounded on **both** axes (`min(24vw, 14vh)`). A `vw`-only hero grew on a short landscape viewport at exactly the moment the vertical room for it shrank — that is the mechanism behind the brand-mark collision.
- `.stage` gets `min-height: 0`. Without it a grid row refuses to shrink below its content's intrinsic height, which is what turned overflow into overlap.

The existing `intro-step-layout.contract.test.ts` assertions still hold unchanged.

## Material

- **Brand bar → React Bits `GlassSurface`.** A real refracting pane (three per-channel `feDisplacementMap` passes, hence the chromatic fringe at the rim), not a blurred rectangle. Distortion softened from the `-180` default to `-92` so the wordmark stays legible.
- **Pointer lens → derived from React Bits `FluidGlass`, `lens` mode.** Upstream owns an entire page via `ScrollControls` and renders its own demo text and images inside the WebGL scene, with the host page portaled through `<Scroll html>`. That is unusable here: the welcome copy must stay real DOM because the voice surface, the a11y tree and the SEO markup all read it. This keeps the `lens.glb` geometry, the pointer damping (`easing.damp3` at `0.15`) and the `MeshTransmissionMaterial` parameters, and drops the demo shell.
- **`LandingAurora`** gives the glass something to refract. A flat canvas makes every pane read grey. CSS gradients only — no JS, no shader.

Content, copy, routes and layout are otherwise unchanged. `MOTIONS`, the approved product line, and the Research/Blog/Developers group are all as they were.

## Performance

- The lens is `dynamic(..., { ssr: false })` and decides for itself whether to mount. On touch (nothing to follow) and under `prefers-reduced-motion` (the effect *is* the motion) it never mounts, so the three/fiber/drei chunk is never fetched and no WebGL context is allocated. Confirmed as a separate lazy chunk, present only in the route's dynamic-import registry and not in the initial entry.
- Nothing in the aurora animates layout or paint — `transform` on composited layers only.
- `lens.glb` is 16 KB.

## ⚠️ One change that is not local to this screen

`@react-three/fiber` globally augments `React.JSX.IntrinsicElements` with every three.js object, and there is **no opt-out**:

```ts
declare module 'react' {
  namespace JSX { interface IntrinsicElements extends ThreeElements {} }
}
```

Once that is in the program, TypeScript's `LibraryManagedAttributes<T, …>` resolution degrades to `never` for `as`-prop components, so `tsc` breaks in four files that never touch 3D:

```
components/app-ui/app-page-shell.tsx        (3 sites)
components/app-ui/fullscreen-flow-shell.tsx
components/app-ui/typography.tsx
lib/morphy-ux/ui/segmented-control.tsx
```

Added `lib/morphy-ux/polymorphic.ts` (a type, not a helper function — a function returning a component trips `react-hooks/static-components`) and applied the cast at the six render sites. Public prop types are untouched, so callers keep full per-element checking; the looseness stops at the JSX tag.

**This is the price of R3F and reviewers should weigh it deliberately.** If it is not worth paying, the lens can be dropped and the `GlassSurface` bar kept — that half needs no new dependencies at all.

Also guarded `CSS.supports` and `ResizeObserver` in the vendored `GlassSurface`; upstream calls both unconditionally and throws during render in jsdom.

## Licensing

React Bits is **MIT + Commons Clause**. Vendored under `components/react-bits/` with provenance headers and every local change marked `HUSHH` at the line. Recorded in `THIRD_PARTY_NOTICES.md`, including the reasoning that inclusion here is use-as-part-of-a-product, which the license permits, rather than redistribution of the components, which it does not. Worth a maintainer's own read before merge.

## Dependencies added

`three`, `@react-three/fiber`, `@react-three/drei`, `maath` — all reachable only from the lazily-loaded lens chunk.

## Verification

- `npx tsc --noEmit` — clean
- `npx eslint` on all touched files — clean
- `npx next build` — succeeds
- `intro-step`, `intro-step-layout.contract`, `home-auth-entry` suites — 8/8 pass

Visual verification is by hand rather than automated; screenshots are worth taking on a short laptop viewport and in landscape, since those are the cases that were broken.

Commits are DCO signed off (`git commit -s`).


RudraO2:
Closing — opened prematurely. The welcome-screen layout is still being worked through, so this doesn't reflect where it's heading. Will raise a fresh PR once the direction is settled. Apologies for the noise.

PR #5957, opened 2026-08-25
fix(referrals): stop referral counters from capping at 50

### LINKED ISSUE #5956: [Bug]:  Referral counters incorrectly capped at 50 users
The Your referrals page stops showing more than 50 In progress referrals, even when additional new users successfully join through the referral link.

Current behavior
Referral link continues to receive new sign-ups.
In progress count reaches 50 and stops increasing.
Newly referred users beyond 50 are not reflected in the total count.
The Recent list also contains only the latest referrals.
Root cause

In consent-protocol/hushh_mcp/services/one_referral_service.py, get_referral_summary() retrieves referral relationships using:

ORDER BY r.created_at DESC
LIMIT 50

The qualified_count, in_progress_count, and under_review_count are then calculated from those same limited rows.

Because only 50 relationships are loaded, the counters can never exceed 50.

Expected behavior

Referral counters should represent all referrals belonging to the referrer, regardless of how many items are displayed in the Recent list.

The Recent/detail list can remain limited to the latest 50 records for performance.

Example:

73 total referrals
68 In progress
5 Qualified

UI should show:

In progress: 68

Qualified: 5

while the Recent list may still return only the latest 50 users.

Fix

Separate the aggregate counter query from the Recent referral query.

Calculate status counts across all one_referral_relationships for the referrer.
Keep LIMIT 50 only for the referral detail/Recent list.
Do not change referral eligibility, attribution, qualification, or privacy rules.
Acceptance criteria
Referral counters can exceed 50.
Counts reflect all referral relationships for the current referrer.
Recent referral rows remain capped at 50.
Existing referral statuses and qualification logic remain unchanged.
Add a regression test with more than 50 referrals.
Existing referral tests continue to pass.

smirthi-dharma:
## Summary
- `get_referral_summary()` computed `qualified_count` / `in_progress_count` / `under_review_count` from the same 50-row `LIMIT` used for the Recent referral list, so counters silently stopped climbing at 50 even when the referrer had more relationships.
- Split the query: a lightweight status-only query now covers every `one_referral_relationships` row for the referrer to compute counts; the existing `ORDER BY r.created_at DESC LIMIT 50` detail query still backs the Recent list unchanged.
- No change to attribution, eligibility, qualification, engagement, privacy rules, `public_status()` mapping, or the API response shape.

Fixes #5956

## Test plan
- [x] Added `tests/services/test_one_referral_service_summary.py` covering the issue's exact scenario (63 total referrals, 60 in progress, 3 qualified) asserting `in_progress_count == 60`, `qualified_count == 3`, `len(referrals) == 50`, plus an `under_review_count` case and a below-cap parity case.
- [x] `pytest -k referral` — 92 passed, no regressions.
- [x] `mypy hushh_mcp/services/one_referral_service.py` — no issues.
- [x] Verified live: ran the real `get_referral_summary()` behind the real FastAPI route (auth dependency overridden, DB layer replaced with the same 63-referral fixture) and hit `GET /api/one/referrals/summary` — response confirmed `qualified_count: 3, in_progress_count: 60, under_review_count: 0, referrals: 50 rows`.

### DIFF
--- a/consent-protocol/hushh_mcp/services/one_referral_service.py
+++ b/consent-protocol/hushh_mcp/services/one_referral_service.py
@@ -213,6 +213,17 @@ def get_referral_summary(user_id: str) -> dict:
     required_minutes = policy.required_active_seconds // 60
 
     with get_db_connection() as connection:
+        all_statuses = connection.execute(
+            text(
+                """
+                SELECT r.status
+                  FROM one_referral_relationships r
+                 WHERE r.referrer_user_id = :uid
+                """
+            ),
+            {"uid": user_id},
+        ).fetchall()
+
         rows = connection.execute(
             text(
                 """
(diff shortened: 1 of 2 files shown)


smirthi-dharma:
Also fixes #5969 (duplicate of #5956).

PR #6067, opened 2026-08-27
fix(brand): pin the emoji font, and stop the agent header shipping Noto artwork

anoushkauoc:
## The bug

On UAT the 🤫 brand mark renders as the **Android (Noto) shushing face on macOS**. Two independent causes, both fixed here.

## Cause 1 — no emoji tier in the font stack (intermittent, most surfaces)

The product font stack is `"InterVariable", "Inter", system-ui, sans-serif`. **None of those carry emoji glyphs**, so every emoji codepoint fell through to the browser's *last-resort* font matching, which has no defined priority order. On a Mac with Noto Color Emoji installed — Google Fonts, Docker and several design tools all ship it — Chrome can pick the Android artwork. Because `InterVariable` loads with `font-display: swap`, the resolution can differ between the first paint and the swapped one, which is why it was wrong *often* rather than always.

Only one of six render sites had pinned the emoji fonts: the top app bar, via a one-off inline style.

**Fix**
- New `--font-app-emoji` token (Apple → Segoe → Noto), appended to `--font-app-display` / `--font-app-body` — the only two consumers of `--font-family-product`. Every surface inherits a deterministic emoji tier.
- `--font-family-product` is left **byte-identical**: `verify-apple-hierarchy` pins its exact literal to keep Inter the product typeface. An earlier revision of this PR folded the emoji tier into it and that guard correctly failed the build. Emoji fonts carry only emoji, so Latin text still resolves to Inter exactly as before.
- `.hushh-brand-mark` class pinning the stack for the logo specifically — it holds even if a component sets its own `font-family` on an ancestor. Applied at every render site; the top bar's inline stack is removed so there is one source of truth.

## Cause 2 — the agent chat header shipped Noto artwork as a raster (always)

The avatar beside "One / Your private agent" in the agent chat header — the surface the **Email agent** opens into — rendered `/one-quiet-emoji.png`. That file *is* the Noto shushing face baked into a PNG, so it stayed Android on a Mac no matter what the font stack resolved to. It was the one brand mark in the app that could not follow the platform, and it is why the Email agent still looked wrong after Cause 1 was fixed.

**Fix** — render the mark as text through `.hushh-brand-mark`, exactly like the top bar, sidebar, intro gate and onboarding, at sizes matching the old raster box (22px desktop, 30px at `max-sm`). Drops the file's last `<Image>`, so the `next/image` import goes too. `/one-quiet-emoji.png` is now unreferenced; left in place rather than deleted, since removing a public asset is a separate call.

## Known limit

This stabilises the mark **per platform** — Apple's artwork on every Mac and iOS device on UAT. It cannot give a Windows or Linux tester Apple's glyph: Apple Color Emoji isn't installed there and isn't redistributable. Byte-identical artwork on every OS would need a drawn SVG brand mark instead of a text emoji — a design decision, not made here. `app/icon.svg` and the iOS app icon source embed the emoji as SVG `<text>` and carry the same caveat.

## Verification

All 13 CI checks green on a branch freshly merged with `main`. Locally: `eslint` clean, `npm run verify:design-system` OK, 40 tests pass across `intro-step`, `hushh-intro-gate`, `auth-step-layout`, `intro-step-layout`, `top-app-bar`, `agent-chat-shell` and `register-phone-safe-area`.

Tracked as #6068.


anoushkauoc:
Tracked as #6068.

Issue #6066, opened 2026-08-27
[Bug]:  Marketplace CTA incorrectly displays "View Profile" instead of "Send Request"


Divya-rajendran009:
## Description
In RIA → Clients → Connected → Browse the Marketplace, the primary CTA is inconsistent across marketplace client cards.

Some cards display:

"Send Request"

while other cards display:

"View Profile"

However, both CTAs currently perform the same Send Request action.

Since the intended action in this marketplace state is to send a connection request, displaying "View Profile" is misleading and does not accurately communicate what will happen when the user clicks the button.

This should be corrected as a frontend UI/UX consistency issue by displaying "Send Request" for the applicable marketplace cards while preserving the existing request functionality.

## Steps To Reproduce
1. Open RIA → Clients.
2. Select **Connected**.
3. Open **Browse the Marketplace**.
4. Compare multiple marketplace client cards.
5. Observe that some cards display **"Send Request"** while others display **"View Profile"**.
6. Click a card displaying **"View Profile"**.
7. Observe that it performs the Send Request action rather than opening a profile.

## Expected Behavior
Marketplace cards whose primary action sends a connection request should consistently display:

**"Send Request"**

The CTA label must accurately describe the action that occurs when it is selected.

After the fix:

- "View Profile" should be replaced with "Send Request" for the affected marketplace state.
- Existing Send Request functionality should remain unchanged.
- No new profile-navigation behavior should be introduced.
- Request state and connection logic should remain unchanged.
- CTA styling and alignment should remain consistent across marketplace cards.

## Scope
Frontend UI/UX only.

Do not modify:
- Backend
- APIs
- Send Request functionality
- Connection/request logic
- Client relationship rules
- Marketplace data
- Authentication
- Authorization
- Data fetching
- Business logic

Only correct the misleading CTA presentation so the displayed label matches the existing action.

Divya-rajendran009:
Linked PR: #6088

PR #6088, opened 2026-08-27
fix(ria): correct marketplace request CTA label

Divya-rajendran009:
Description
In RIA → Clients → Connected → Browse the Marketplace, the primary CTA is inconsistent across marketplace client cards.

Some cards display:

"Send Request"

while other cards display:

"View Profile"

However, both CTAs currently perform the same Send Request action.

Since the intended action in this marketplace state is to send a connection request, displaying "View Profile" is misleading and does not accurately communicate what will happen when the user clicks the button.

This should be corrected as a frontend UI/UX consistency issue by displaying "Send Request" for the applicable marketplace cards while preserving the existing request functionality.

Steps To Reproduce
Open RIA → Clients.
Select Connected.
Open Browse the Marketplace.
Compare multiple marketplace client cards.
Observe that some cards display "Send Request" while others display "View Profile".
Click a card displaying "View Profile".
Observe that it performs the Send Request action rather than opening a profile.
Expected Behavior
Marketplace cards whose primary action sends a connection request should consistently display:

"Send Request"

The CTA label must accurately describe the action that occurs when it is selected.

After the fix:

"View Profile" should be replaced with "Send Request" for the affected marketplace state.
Existing Send Request functionality should remain unchanged.
No new profile-navigation behavior should be introduced.
Request state and connection logic should remain unchanged.
CTA styling and alignment should remain consistent across marketplace cards.

### DIFF
--- a/hushh-webapp/__tests__/lib/marketplace/primary-card-label.test.ts
+++ b/hushh-webapp/__tests__/lib/marketplace/primary-card-label.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveMarketplacePrimaryCardLabel } from "@/lib/marketplace/primary-card-label";
+
+describe("resolveMarketplacePrimaryCardLabel", () => {
+  it("labels RIA primary actions as send requests when the handler sends a request", () => {
+    expect(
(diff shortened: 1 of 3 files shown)


Divya-rajendran009:
Fixes #6066

Issue #6134, opened 2026-08-28
[Bug]: Minimizing the agent returns to /one instead of the page it was opened from


anoushkauoc:
## Description

Open the agent from a feature page (Email agent, Gmail workspace, others), maximize the chat, then minimize it — you land on the **Agents roster at `/one`**, not the page you came from. Reported across many pages, not just the Email agent.

## Root cause (full-page `/agent` route)

`AgentChatWorkspace.handlePageMinimize` (`components/agent/agent-chat-workspace.tsx`) decides where to go by reading `document.referrer`:

```ts
const referrer = document.referrer ? new URL(document.referrer) : null;
const isSameOriginReferrer =
  referrer?.origin === window.location.origin &&
  referrer.pathname !== ROUTES.AGENT;
if (isSameOriginReferrer && window.history.length > 1) { router.back(); return; }
router.push(ROUTES.ONE_HOME);   // ← /one, the Agents roster
```

**`document.referrer` is not updated by Next App Router client-side navigation.** Every entry point into this route is a client `router.push(ROUTES.AGENT)` — `app/one/email/email-agent-page-client.tsx:68` and `components/gmail/gmail-receipts-page.tsx:807`, both of which fall back to that push whenever the popover provider isn't available. So `document.referrer` still holds whatever the document was *loaded* with (usually empty, or an external URL), `isSameOriginReferrer` is false, and the handler always falls through to `router.push(ROUTES.ONE_HOME)`.

That is exactly the reported symptom, and it is page-agnostic — which matches "happening across many other pages too". Any page that hands off to `/agent` loses its origin on minimize.

## Suggested directions

- Carry the origin explicitly — `/agent?from=/one/email` — and minimize back to it. Survives reloads and deep links, unlike history heuristics.
- Or store the origin route on the handoff in `one-conversation-session` when `createHandoff` is called, since every entry point already goes through it.
- Keep `ROUTES.ONE_HOME` as the fallback only for genuine direct links into `/agent` with no recorded origin.
- Whatever the mechanism, `document.referrer` should not be part of it in an App Router app.

## Open question — the popover path may be a second defect

[...]

PR #6135, opened 2026-08-28
fix(agent): return to the page the agent was opened from on minimize

anoushkauoc:
Fixes the `/agent` half of #6134.

## The bug

Open the agent from a feature page, maximize, minimize → you land on the Agents roster at `/one`, not the page you came from. Reported across many surfaces, not just the Email agent.

## Root cause

`handlePageMinimize` inferred the destination from `document.referrer`:

```ts
const referrer = document.referrer ? new URL(document.referrer) : null;
const isSameOriginReferrer = referrer?.origin === window.location.origin && ...
if (isSameOriginReferrer && window.history.length > 1) { router.back(); return; }
router.push(ROUTES.ONE_HOME);
```

**`document.referrer` is never updated by App Router client navigation.** Every entry point into `/agent` is a client `router.push` — `email-agent-page-client.tsx` and `gmail-receipts-page.tsx` both fall back to one when the popover provider isn't available. So the referrer held whatever loaded the document (usually nothing), the guard was always false, and minimize fell through to `ROUTES.ONE_HOME` every single time. Page-agnostic by construction.

## The fix

Carry the origin explicitly: `/agent?from=/one/email`. Chosen over a history heuristic because it survives a reload, a shared link and a restored tab — none of which `history.length` or a referrer can speak for.

- `lib/navigation/agent-origin.ts` — `agentRouteWithOrigin()` to build the href, `readAgentOrigin()` to read it back
- `handlePageMinimize` checks the recorded origin first, then keeps the existing referrer/`router.back()` path and the One-home fallback for direct links carrying no origin
- Both hand-off sites record where they came from. Gmail's page is hosted on **both** `/one/gmail` and `/one/setup/gmail`, so it records the live `usePathname()` rather than a route constant

### On the untrusted input

The origin arrives from the query string, so a shared link can carry anything. `readAgentOrigin` accepts a single-leading-slash app path only and returns null otherwise — rejecting protocol-relative (`//evil.com`), scheme-carrying (`https://evil.com`), backslash-escaping (`/\evil.com`) and `javascript:` values, all of which would leave the origin if pushed unchecked. `/agent` itself is rejected too, so minimize cannot loop back to the screen it is leaving. `agentRouteWithOrigin` drops an unsafe origin rather than encoding it, degrading to the plain route.

## Still open

The popover-path question in #6134 is **not** addressed here. In the reported screenshot the URL reads `/one/gmail` with the chat maximized — that is the popover at `sizeMode: "fullscreen"`, whose window controls only call `setSizeMode` / `requestClose` and never navigate. If minimizing *there* also redirects, it is a separate defect; #6134 records how to tell the two apart.

## Verification

8 new tests for the origin helper (round trip, missing/blank input, and each off-site escape vector). 36 existing tests pass across the agent chat shell, popover provider, command handoff, and all four Gmail page suites. `eslint`, `tsc --noEmit` and `npm run verify:design-system` clean.

### DIFF
--- a/hushh-webapp/__tests__/lib/agent-origin.test.ts
+++ b/hushh-webapp/__tests__/lib/agent-origin.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  agentRouteWithOrigin,
+  readAgentOrigin,
+} from "@/lib/navigation/agent-origin";
+import { ROUTES } from "@/lib/navigation/routes";
+
+describe("agentRouteWithOrigin", () => {
+  it("records the originating page on the agent route", () => {
+    expect(agentRouteWithOrigin(ROUTES.EMAIL_AGENT)).toBe(
+      "/agent?from=%2Fone%2Femail",
+    );
+  });
+
+  it("keeps a query string on the origin intact through a round trip", () => {
+    const href = agentRouteWithOrigin("/one/gmail?tab=receipts");
+    expect(readAgentOrigin(new URL(href, "https://one.hushh.ai").search)).toBe(
+      "/one/gmail?tab=receipts",
+    );
+  });
+
+  it("degrades to the plain route rather than encoding an unsafe origin", () => {
+    for (const unsafe of ["//evil.com", "https://evil.com", "", null]) {
+      expect(agentRouteWithOrigin(unsafe)).toBe(ROUTES.AGENT);
+    }
+  });
+});
+
+describe("readAgentOrigin", () => {
+  it("reads the recorded origin", () => {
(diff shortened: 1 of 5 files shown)


anoushkauoc:
Superseded by #6573.

This branch carried the right fix but sat ~690 commits behind `main`, so it could not merge past the Base Freshness Gate without a refresh. #6573 is this exact commit rebased onto current `main` (authorship preserved), with two conflicts resolved: the Email Agent page's dependency array gained `user?.uid` from the first-run intro gate (#6527), and the Gmail page's import line had drifted.

#6573 also answers the open question this PR deliberately left: the popover path does **not** navigate on minimize, and two new tests hold that at both sizes. Closing here so the work lands in one place.

PR #6248, opened 2026-08-30
fix(one): make a Circle roster say who a menu belongs to, and who owns the Circle

DamriaNeelesh:
Three defects reported from phone QA on Circle Detail, all the same shape: a screen correct in every part and wrong as a whole.

## 1. The kebab's actions landed on the next member's name

Reported with a screenshot of the SMS Circle roster: tapping the kebab on **Ankit Kumar Singh** opened a small floating card that painted over **JHUMMA KUMARI**'s name in the row below. Nothing on the surface said whose actions those were, and the one cue available — proximity — pointed at the wrong person. *"Dusre connection ke name ke upar aa gaya."*

That is not a positioning bug to nudge with an offset. An anchored popover is a **pointer** idiom: it works because a cursor is already at the anchor and the eye follows it there. On a phone the finger covers the anchor, the menu paints into whatever row is below, and a roster is a list of near-identical rows — so the one piece of context that would disambiguate it is exactly what gets covered.

The surface now follows the pointer:

| | phone (<640px) | desktop (≥640px) |
|---|---|---|
| surface | bottom action sheet over a scrim | anchored menu |
| identity | member's own avatar + name + status in the header | member's name as a menu header |
| rows | 56px, full-bleed | 44px (was 30px) |
| destructive confirm | second pane of the **same** sheet | AlertDialog, and the menu now **closes** first |

The confirm no longer stacks. On the phone it cannot: vaul's content sits at `z-712` and the alert at `z-500`, so a dialog opened over a closing sheet would spend its entrance animation behind the thing it replaced. On the desktop, the menu used to be left **open** behind the confirm — two surfaces for one decision.

New module: `circle-member-actions-menu.tsx`. It owns the kebab, both surfaces, the confirm, and the inert 44px spacer that holds the kebab column open on rows with no menu — so "does this row have a menu" cannot be answered two ways in one render.

## 2. The owner sat wherever the alphabet put them

Reported on a Trusted Circle reading *Ankit Kumar Singh / JHUMMA KUMARI (You · Owner) / Neelesh Meena* — *"owner hamesha sabse upar hi rahega, sabhi kind ke circles ke liye."*

Nothing was mis-sorted: the roster is A–Z, and J falls between A and N. But A–Z is the right rule for a list of **peers**, and a Circle's roster is not one — the owner is the only row whose position carries information.

`sortCircleMembersOwnerFirst` is a **stable partition**, not a second sort key, applied to the rendered list. So A–Z survives beneath the owner, and so does the server's order on a paged roster. It hoists only a row already in the list — it never fetches one.

It stops at the search box, deliberately. This branch rebases onto #6244, which gave the server-paged member search the same one-letter ranking the client already had. Hoisting the owner through that ranking would answer a typed question with the wrong person: `"ku"` would return Jhumma Kumari above Ankit Kumar Singh, when both begin a word with it and A–Z inside the rank says otherwise. A roster has no question to answer, which is exactly why the owner leads it.

## 3. The Add people sheet had 48px of gap above its search field

Reported as *"extra space between the section is bad"*, *"search bahut jyada neeche aa raha"*, *"divs truncated bhi dikh rahe hain."*

None of that gap was written on this screen. It was four sensible defaults stacking, each defaulting for a context it could not see:

```
SheetContent   flex flex-col gap-4   +16px between every child
SheetHeader    p-4                   +16px below the description
                                     (and +16px to its LEFT, on top of the
                                      sheet's own px-4 — so the title sat
                                      indented from the field under it)
the body       mt-4                  +16px again
SettingsGroup  mt-7                  +28px before "Your connections"
```

`SettingsGroup`'s `mt-7` separates one section from the section **above** it — correct on a settings screen, pure gap in a sheet whose header has already introduced the list. It now takes a `headingClassName` (additive; every existing caller is untouched), the same escape hatch `shellClassName` and `contentClassName` already are.

The "truncated divs" half is the scroll area: an `overflow-y-auto` box clips at its own padding edge, so a card flush against it lost the outer pixels of its rounded corner and all of its shadow, and read as a row cut in half rather than a list that scrolls. One pixel of gutter, pulled back with a negative margin so nothing moves.

`circle-sheet-layout.ts` states the rhythm once and all three Circle sheets (Add people, Invite code, Rename Circle) share it.

**On scrolling, since it was asked:** yes, everywhere. `SheetContent side="bottom"` caps at `max-h-[calc(85dvh - --kb-height)]` with `overflow-y-auto overscroll-contain`, so any bottom sheet scrolls rather than running off the screen. Add people additionally owns an inner scroll region so its "Add N people" button stays pinned while the roster moves under it — that is now asserted, not assumed.

## Verification

- `npm run verify:one-location` — 674 tests. The one failure is the known 5s-timeout flake in `one-location-agent-page.test.tsx` under parallel load; 126/126 pass isolated.
- 78 tests across the Circle suites, 14 of them new: both action surfaces, the owner partition (both circle kinds), the search carve-out, and the sheet spacing contract.
- `tsc --noEmit`, `eslint`, `verify:design-system`, `verify:accent-tokens`, `verify:apple-hierarchy`, `verify:surface-map` — all clean.
- Protected behaviour `location-people-search-finds-a-person-from-one-letter` untouched: both required titles still present, assertion count only rises.
- No schema change, no migration.

### DIFF
--- a/hushh-webapp/components/app-ui/settings-ui.tsx
+++ b/hushh-webapp/components/app-ui/settings-ui.tsx
@@ -176,6 +176,7 @@ export function SettingsGroup({
   embedded = false,
   separatorInset,
   className,
+  headingClassName,
   shellClassName,
   contentClassName,
   testId = "settings-group",
@@ -218,6 +219,21 @@ export function SettingsGroup({
    */
   separatorInset?: boolean;
   className?: string;
+  /**
+   * Tunes the heading block's own spacing.
+   *
+   * The block carries `mt-7` because a settings SCREEN stacks groups down a
+   * page and 28px is what separates one section from the last. Inside a
+   * bottom sheet there is no previous section -- the group is the first thing
+   * under a header that has already introduced it -- so that 28px lands as a
+   * gap between the sheet's own title and the list it is titling. Reported on
+   * the Add people sheet as "search bahut jyada neeche aa raha".
+   *
+   * A caller-supplied class rather than a `compact` boolean: the value that is
+   * wrong here is a margin, and the surfaces that need to change it are the
+   * ones that own their own vertical rhythm.
+   */
+  headingClassName?: string;
   /** Lets a bounded manager make the shared group shell a flex viewport. */
   shellClassName?: string;
   /** Lets a bounded manager make the shared row stack the scroll owner. */
@@ -258,7 +274,12 @@ export function SettingsGroup({
   return (
     <section className={cn("w-full", className)} data-testid={testId}>
       {eyebrow || title || description || titleAction ? (
-        <div className="mb-2 mt-7 flex items-start justify-between gap-3 px-[6px]">
+        <div
+          className={cn(
+            "mb-2 mt-7 flex items-start justify-between gap-3 px-[6px]",
+            headingClassName,
+          )}
+        >
           <div className="min-w-0 flex-1 space-y-[var(--settings-heading-stack-gap)]">
             {eyebrow || title ? (
               <SectionLabel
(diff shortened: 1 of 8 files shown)


DamriaNeelesh:
Tracked by #6249 (Hussh Action Items → In Progress).

PR #6280, opened 2026-08-31
feat(one): rate the place you just checked out of

DamriaNeelesh:
Check-In ends flat today: you check out, the sheet offers "Save for faster check-ins", and nothing asks the only question that moment is good for. This is **PR 1 of 2 — capture**. The history screen and the public averages land in PR 2.

**Do not merge yet** — we are testing this locally first.

---

## Two facts that shaped every decision here

**1. Google has no write API for consumer reviews.** The Places API is read-only on them (max 5), and the Business Profile API only lets a *business owner reply* on its own listing. There is no OAuth scope for "post as this user", and text cannot be prefilled. The deep link into Google's own composer is the entire sanctioned integration.

**2. This repo has a written contract forbidding what a rating table needs.** [consent-scope-catalog.md](docs/reference/iam/consent-scope-catalog.md): *"Plaintext coordinates, exact distance, **provider place ids, place labels** … are **forbidden** in nearby-presence persistence, logs, analytics, and audit metadata."* `place_id` appears in **zero** SQL columns repo-wide; `check_in()` validates one and drops it; `checkout()` NULLs all seven anchor columns. There was no record anywhere that a check-in had ever happened.

So the exception is **stated in the catalog in this same PR** and bounded to two tables, rather than shipped quietly against a document that says otherwise.

## The storage split

```
VAULT (client-encrypted, next to Saved Places)
  visit_notes -> placeId, label, rating, note, visitedAt
  The note NEVER reaches a server. The endpoint has no field for it.

SERVER  one_location_nearby_visits     encrypted, 7 days, purged.
                                       Venue readable only with the key,
                                       plus a keyed equality token.
SERVER  one_location_place_ratings     PLAINTEXT place id, permanent,
                                       UNIQUE(author, place).
SERVER  one_location_place_rating_aggregates   count + sum. No user reference.
```

The durable person-to-venue link is created at exactly one moment: when its author accepts `one-location-place-rating-v1`. `consentVersion` is **sent by the client and checked server-side**, not stamped there — server-stamping is fine for a 60-minute presence, but a permanent record must not be writable by a build still showing last version's words.

**Deliberately absent:** no review-text column, no author display name, no author id in any read projection. Averages are withheld below **5 raters** and reported in **buckets** — an exact `(count, average)` pair lets an observer recover each new rating by subtraction, at any n. Health, worship, legal, funeral and shelter categories never carry an average at all: excluded, not warned about, because nobody picks a hospital by a star rating and the worst harm this feature could do lives there.

## The UI

One pane, fixed-height slots, inside the existing `completed` branch — a flag on `CompletedCheckIn`, **not** a new `NearbyCheckInViewState`. That enum is written from server responses in six places including a 15s poll, which would otherwise be free to knock someone out of the step mid-note.

```
arrival                          star tapped (nothing moves)
+----------------------------+   +----------------------------+
| v Check-in ended           |   | v Check-in ended           |
|   Bag Maker                |   |   Bag Maker                |
| How was Bag Maker?         |   | How was Bag Maker?         |
|  o  o  o  o  o             |   |  *  *  *  *  o    Good     |
| +------------------------+ |   | +------------------------+ |
| |Tap a star. Only you    | |   | |Anything worth          | |
| |see this.               | |   | |remembering             | |
| +------------------------+ |   | +------------------------+ |
| [      Save     ] (off)    |   | [      Save     ] (on)     |
| [     Not now   ]          |   | [     Not now   ]          |
+----------------------------+   +----------------------------+
```

The height never changes after arrival. That is not tidiness: the map puts a `ResizeObserver` on this sheet and republishes `map.setPadding()` on every height change, so an "append a block" design settles the camera three times for one interaction. A Playwright contract measures it, because JSDOM cannot see any of it.

**Where the place id comes from:** the server, on its checkout response. Not the device — the presence payload carries a label and a point and never a place id, and a remembered one is how a Google review gets attached to the wrong venue. No place id means the hand-off row is simply absent.

**The Google half, honestly:** lowercase `placeid` (camelCase silently lands on Google's generic search page — pinned by a test, not described); an anchor, so iOS Capacitor hands it to `UIApplication.open` and the universal link reaches the Maps app, and Android fires `Intent.ACTION_VIEW`; and it appears **only after** the local save succeeds, says "**Also** post on Google", and the line above it says the words are not carried over.

`"Only you see this."` is the most important sentence on the pane. Without it a star row above a Google button reads as though it publishes somewhere.

## Three pre-existing defects fixed on the way past

1. **Check-out disarmed the anti-spoof guard.** `checkout()` NULLs `anchor_algorithm` -> `_decrypt_anchor` raises on its first line -> `_require_reachable_from_last_check_in` catches and returns. **So after any explicit check-out the next check-in had no travel-plausibility check at all**, and "check in anywhere, check out, repeat" cost nothing — a loop this feature would have trained every user into. The visit ledger survives checkout, so the guard now has a second place to look. Its fail-open discipline is unchanged.
2. **`cap.location.nearby.*` were declared but unresolvable.** In `ConsentScope` and `capability_scopes()`, missing from `_AGENT_SCOPE_MAP`, so `resolve_scope_to_enum` raised `ValueError` on all three — and missing from `_STATIC_SCOPE_META`, so the consent surface would have shown "Cap Location Nearby Publish".
3. **Five nearby check-in test files had never run.** None was in `scripts/test-ci.manifest.txt`, the only pytest invocation in any lane.

## Verification

| | |
|---|---|
| Backend | **120 green** (39 new). Pinned ruff 0.15.11 clean on check and format. |
| Frontend | **95 green** across check-in / star / maps-url (16 new). The five pre-existing post-checkout tests pass **unchanged** — the real regression guard. |
| `verify:one-location` | 756/757. The one failure is the known 5s-timeout flake in `one-location-agent-page` under parallel load; **126/126 isolated**. |
| Gates | `verify:analytics` (41), `verify:design-system`, `verify:accent-tokens`, `verify:apple-hierarchy`, `verify:surface-map`, `tsc --noEmit`, `eslint` — all clean. |

**Not verified locally, and worth saying:** the four new routes are registered, error-mapped and `extra="forbid"`-checked by direct import, but `tests/test_one_location_nearby_presence_routes.py` could not be collected on this machine — `ag-ui-protocol` is a declared dependency missing from the local venv. CI has it. The aggregate has no display in PR 1, so the k-threshold and the sensitive-category denylist are unit-tested only. And production stays dark behind `_require_nearby_presence_simulation` until the mode and a cohort are both set — correct, and someone will file a bug about it.

`scripts/ci/web-targeted-check.sh` needs **no** edit: every new file lands inside a tree it already matches.

## Not built, on purpose

Public review text, author identity on anything, a "verified visit" badge (the position is client-supplied and `one-location-agent.md` already forbids the claim), Google's own ratings shown next to ours (Enterprise SKU plus redistribution terms), third-party moderation APIs, and any editorial ranking.


DamriaNeelesh:
PR 2 is up: #6325 (Places you've been + anonymous averages on the nearby list). It is stacked on this branch, so merge this one first and I'll rebase it onto `main`.

PR #6324, opened 2026-09-01
fix(connect): one segmented material, and make My connections a disclosure

DamriaNeelesh:
Three reports from the Location and Connect screenshots.

---

## 1 · One segmented control, not three looks

> "in sb places pe inka UI and layout design type pura consistent hona chahiye … abhi sb mai alag alag feel aa rha hai"

There were three looks, and the cause turned out not to be styling drift — it was **two components and two global `!important` rules pulling in opposite directions.**

| | Location `Now / People / Links` | Connect `Connections / Circles` |
|---|---|---|
| Component | `TopShellTabs` | `SegmentedTabs` (+13 other strips) |
| Track | grey `--app-neutral-fill`, no border | near-white `--app-card-surface-compact`, card border |
| Pill | raised, soft shadow | outlined, `shadow: none` |
| Active label | **accent blue** | **`--app-label` dark** |

Same control, two materials, on consecutive screens.

### The colours were the worse half

**Neither component's own active colour ever applied.** `.ui-text-agent-tab-label` and `.ui-text-form-label` both set `color: … !important`, so `TopShellTabs`'s `text-[color:var(--app-label)]` and `SegmentedTabs`'s `text-[color:var(--app-segmented-active-foreground)]` were dead classes.

What actually decided it was this, in `globals.css`:

```css
[role="tab"][aria-selected="true"] :is(.ui-text-agent-tab-label) {
  color: var(--app-accent) !important;
}
```

`TopShellTabs` uses `role="tab"`. `SegmentedTabs` uses `aria-pressed`. So Location's active tab rendered blue and Connect's identical control rendered dark, **from rules neither file mentions.** This is exactly the shape of R19 in `safe-changes` — a CSS rule that "obviously" wins may not.

### The fix

At the **token layer**, so all fifteen call sites move together rather than fifteen edits — `safe-changes` R14, one shared implementation, not one fix per address:

- new `--app-segmented-track-surface`
- `--app-segmented-active-*` repointed at the raised-pill values
- the `[role="tab"]` accent rule scoped away from `[data-top-shell-tab-set="location"]`, an attribute `TopShellTabs` already writes

**Accent stays where it belongs** — the *underline* tab bars (Consent Center, RIA, Finance), where the tint is the only selection signal there is. A segmented strip marks the selection with the pill; tinting the label too is what made a segmented control read as a tab bar. Two coherent families now, instead of three arbitrary looks.

**Heights deliberately not unified.** Location's strip is 36px, the in-page control 44px. That difference is *density*, not material — and shrinking fifteen live touch targets to 36px for cosmetic parity is an accessibility regression, not consistency.

---

## 2 · Select and Sync were the wrong way round

> "the positioning of the select button and sync button got interchanged"

They were, and what gives it away is **what each one acts on**:

- **Sync contacts** fills the People directory from the address book → belongs beside the picker that says *which* directory you're looking at.
- **Select** turns that directory's rows into checkboxes → belongs on the directory's own header, not a scroll away above a list it does not touch.

Swapped, each keeping its own gate (Sync stays People-only and hidden where there's no readable source; Select stays disabled on an empty list).

---

## 3 · My connections is a disclosure

> "Connections wala ek accordion tab ki tarah ho jo click krne pe he open ho … currently it looks long for me"

Two stacked lists of people on one screen — the ones you already know, and the ones you're looking for — and the first pushed the second below the fold on a phone, **although the second is what the screen is for.** Searching is why someone opens Connect; reviewing who you already have is the occasional visit.

- **Closed by default**, and deliberately *not* remembered per person: the default is the answer to "what is this screen for", and it shouldn't drift.
- **The heading is the toggle**, carrying `aria-expanded` and `aria-controls` — a chevron is not an affordance to anyone who can't see it. It's the title node itself rather than a button *inside* the title, which is the mistake the group's own `titleAction` comment already records (a control inside `role="heading"` gets folded into the heading's name and is never offered as a control).
- `SettingsGroup` gained a `contentId` so the panel `aria-controls` names is a stable node whether or not it's showing.
- Rows are `hidden`, not unmounted — scroll position and in-flight row state survive a close/open.
- **Refresh renders only while open**: a refresh control over a collapsed panel reloads something nobody is looking at.

---

## Verification

| | |
|---|---|
| webapp typecheck + eslint | clean |
| `npm run verify:design-system` | design-system ✓ accent tokens ✓ Apple hierarchy ✓ |
| `e2e/one-location-tab-strip.layout.spec` (Chromium) | **10 / 10**, unchanged |
| vitest — connect, one-location, settings-ui, tab contracts | **957 / 958** |
| runtime topology index | current |

The one failure is the known load flake (`resets every abandoned share field`) — fails identically on a clean tree, passes isolated.

`e2e/tab-title-integrity`'s fixture was carrying the *old* container classes (`rounded-full p-1 border`), so it had been measuring a strip nobody ships. It now carries the shipped ones.

## Still needs a human

Put Location and Connect side by side on a phone — the two strips should now read as the same control: grey track, raised white pill, dark semibold active label. Then Consent Center, which should still be the underline bar with a blue active label, untouched.

On Connect: **Sync contacts** top-right beside `People ⌄`, **Select** on the People directory header, and **My connections** collapsed until you tap it.


DamriaNeelesh:
**Web Core is red from pre-existing breakage on `main`, not from this PR.**

The three errors are in `components/agent/agent-chat-workspace.tsx`, which is **byte-identical to `origin/main`** on this branch (`git diff origin/main` for that path is empty) and fails `eslint` there too:

```
91:3     'getPkmAutoSaveCards' is defined but never used
1331:10  'pkmAutoSavePolicy' is assigned a value but never used
2570:9   'saveEligiblePkmCardsInBackground' is assigned a value but never used
```

It arrived with `30be4abcd feat(one): make AG-UI chat Morphy-native`. `main`'s own gates never caught it because **Main Post-Merge Smoke does not run `npm run lint`** — only the PR lanes do, so lint breakage can sit on main indefinitely. Same class of problem as the ruff-format drift fixed in #6139.

**Deliberately not touched here.** `setPkmAutoSavePolicy` *is* called (4 sites) — only the read is unused, and `saveEligiblePkmCardsInBackground` is a complete callback nobody invokes. That is a **half-wired PKM auto-save feature**, not dead code: deleting it would remove someone's in-flight work, and silencing it with a disable comment would mask a real "you forgot to call this". It needs its author, in its own PR.

**The gap that red lane left was closed by hand instead:** `npm run build` run locally on this merged branch, exit 0, full route manifest generated. Plus `tsc`, `eslint` on every touched file, `verify:design-system` (3/3), the Location tab-strip browser spec 10/10, and 260/260 vitest across connect, settings-ui and the tab contracts.

PR #6325, opened 2026-09-01
feat(one): Places you've been, and anonymous averages on the nearby list

DamriaNeelesh:
**Stacked on #6280.** Base is `feat/check-in-rate-your-visit`, not `main`, because this reads the tables and service that PR 1 creates. Merge #6280 first; I'll rebase this onto `main` straight after.

**Do not merge yet** — both PRs are for local testing first.

PR 1 collects a rating and then never shows it again. This is the half that reads it back, on the two surfaces where it means something.

---

## 1. Places you've been

A new `?action=places-visited` flow: every place you rated, its stars, your note, and a per-place link back to Google.

```
Location > Places you've been
+--------------------------------------------------+
| [pin]  Bag Maker                     Google  [X] |
|        * * * * o   Good                          |
|        Quick and friendly.                       |
+--------------------------------------------------+
| [pin]  Red Eagle Army Canteen        Google  [X] |
|        * * * o o   Okay                          |
|        Unlock your vault to see your note.       |
+--------------------------------------------------+
```

It is assembled from **two stores that never meet on a server**:

| half | where | why there |
|---|---|---|
| the star | Postgres | an average and one-vote-per-place cannot be computed on a device |
| the note | your vault, client-side encrypted | free text about a named business, beside a venue and a timestamp, is a movement log with commentary |

**The join happens on the reader's own device**, keyed on `placeId`. That is the only place the two halves are ever in one memory — and it is why there is no "recent reviews" endpoint for somebody to build on top of later by accident.

- A **locked vault costs the note and nothing else.** The ratings still list; the row says so rather than looking broken.
- **Deleting removes both halves together.** A note that outlives its rating brings the place back with no stars on the next load.
- Entry point is a Settings row, `Places you've been` — without one the ratings are write-only.

## 2. Averages on the nearby list

```
Nearby places
+--------------------------------------------------+
| [pin]  Bag Maker                            9 m  |
|        Store · *4.5 · 5+                         |
+--------------------------------------------------+
| [pin]  Shree Ambba Luggage                 62 m  |
|        Store                                     |
+--------------------------------------------------+
```

`POST /location/place-ratings/summaries` takes up to 25 place ids and returns `{placeId, average, countBucket}` for the ones that have earned a public average.

- **One call for the whole list**, not one per row. Twenty round trips for a decoration is waste, and per-row calls would hand an observer a per-place timing signal they do not otherwise get.
- A place below five raters **returns nothing at all**, and the row shows nothing — not "no ratings yet", which is both noise on a list built to be scanned *and* a confirmation of exactly which places nobody has rated.
- **Counts stay bucketed** (`5+`, `10+`, `50+`) for the reason PR 1 gives: an exact count beside an exact average lets a watcher recover each new rating by subtraction, at any n.
- A POST rather than a GET because twenty provider place ids do not belong in a URL — a place id in a URL is a place id in every access log between here and the browser.
- The whole effect is **silent on failure**. An average is an ornament on that list, and a list that refuses to render because an ornament could not load is worse than a list without it.

## Header contract

`places-visited` is registered in `oneLocationActionLabel` as **"Places you've been"**, which is exactly the flow's `TaskFlowHeader` title — the rule `location-header-system` exists to enforce. The `__tests__/utils/top-shell-breadcrumbs.test.ts` table gains the case.

## Verification

| | |
|---|---|
| Backend | **107 green** (4 new on the batch projection: threshold, dedup, bound, junk-tolerance). Pinned ruff 0.15.11 clean. |
| Frontend | **764 green** across the whole `verify:one-location` pack — 7 new on the flow, 1 new breadcrumb case, and the 73 check-in tests still pass. |
| Playwright | **24/24** on the check-in panel layout contract. |
| Gates | `tsc --noEmit`, `eslint` clean. |

**Caveat worth stating plainly:** on a stacked base, this repo's CI lanes under-report — the change-detection compares against the base branch, so lanes can skip and the PR can look green while running very little. That is why everything above was run locally and is quoted with numbers. Once #6280 merges I rebase onto `main` and CI runs properly against it.

One known flake seen once and not reproducible: `one-location-agent-page.test.tsx` timed out on a single run under parallel load and passed on the immediate re-run of the same pack (764/764).


ankitkumarsingh1702:
## Deployed to UAT

### Why It Matters

This PR is included in the current healthy UAT release state. The latest successful release deployed main SHA 91913f01b0d12ddf44a5c78ed88281ed78ab2fb3 (which contains this PR) through https://github.com/hushh-labs/hushh-research/actions/runs/33527817922.

UAT is serving frontend revision hushh-webapp-00733-gcj at 100% traffic with backend revision consent-protocol-00574-zcq. Runtime parity, deployment provenance, passkey-domain associations, and semantic health checks passed with no blocking classifications.

PR #6327, opened 2026-09-01
fix(connect): give Connections the same row rhythm as Circles

DamriaNeelesh:
> "in the connect tab the UI of circle and Connections dono ka thoda alag alag feel ho rha hai — I like the UI and design and layout guidelines in Cicle wale mai so I want the same consistency on Connection tab as well"

## It wasn't a styling choice — it was arithmetic

Both tabs already use the same primitives: `SettingsGroup` + `SettingsRow` at `density="compact"`, same `space-y-4 sm:space-y-5` container, same `separatorInset`. So the difference had to be somewhere else, and it was.

`SettingsRow` draws its inset separator from a fixed offset per density:

| Density | Inset | Where the number comes from |
|---|---|---|
| `compact` | **58px** | 16px row padding + **28px leading** + 12px gap |
| `comfortable` | **62px** | 16px row padding + **34px leading** + 12px gap |

Those aren't arbitrary — each is measured around the leading visual that density expects. A compact row expects **28px**, which is exactly what its own icon well is (`h-7 w-7`), and exactly what the Circles rows use.

**`ConnectionPersonAvatar` was 34px — the *comfortable* size — at every call site, including inside Connect's compact rows.**

So on the Connections tab every hairline started at **58px** while the text it was aligning to started at **62px**. Every separator in the list sat 4px shy of its own column. The Circles tab beside it, same primitive, same density, lined up exactly.

```
Circles (28px leading)              Connections (34px leading)
┌──────────────────────────┐        ┌──────────────────────────┐
│ ▢  Trusted            ›  │        │ ◯   Ankit Kumar   Remove │
│    ─────────────────────  │  58px  │  ──────────────────────  │  58px
│ ▢  SMS Circle         ›  │   ✓    │ ◯   Anoushka      Remove │   ✗ 4px short
└──────────────────────────┘        └──────────────────────────┘
```

Same screen, same component, two rhythms. That is what "alag alag feel" was.

## The fix

At the number's source, not by nudging a margin at the call sites: the avatar takes a `size` matched to its row density, and all three Connect lists — connections, the People directory, and the batch-request sheet — ask for `compact`.

The verified badge scales with the face too, or a 15px badge swallows a 28px one.

**`comfortable` stays the default**, so the Location hub and the contact picker — which use comfortable rows and were already aligned — don't move.

## Verification

| | |
|---|---|
| webapp typecheck + eslint | clean |
| `npm run verify:design-system` | design-system ✓ accent tokens ✓ Apple hierarchy ✓ |
| vitest — connect, one-location, connect components | **510 / 510** |

The new case walks **every** rendered avatar rather than sampling one, so a fourth call site can't land at the wrong size unnoticed.

## ⚠️ Pre-existing on main, untouched here

`components/one-location/redesign/__tests__/share-circle-sections.test.ts` fails with:

```
TypeError: shareCircleSections is not a function
```

Both that file and its subject are **byte-identical to `origin/main`** on this branch — an export was renamed or removed without its test. Main's own gates miss it for the same reason they missed the lint break fixed in #6324: **post-merge smoke does not run the full vitest suite**, and PR lanes only run path-scoped packs.

Not fixed here — guessing at somebody's rename would be worse than reporting it. It needs its author.

## Still needs a human

Open Connect on a phone and flip between **Connections** and **Circles**. The separator hairlines should now start at the same place on both, and the two lists should read as one screen rather than two.


ankitkumarsingh1702 (approved):
Reviewed the exact current head for the compact-row avatar sizing contract. The change keeps the comfortable default, scopes compact sizing to compact Connect rows, scales the verification badge with the avatar, and adds coverage across rendered rows. Current required checks are green.

ankitkumarsingh1702:
## Deployed to UAT

### Why It Matters

This PR is included in the current healthy UAT release state. The latest successful release deployed main SHA 91913f01b0d12ddf44a5c78ed88281ed78ab2fb3 (which contains this PR) through https://github.com/hushh-labs/hushh-research/actions/runs/33527817922.

UAT is serving frontend revision hushh-webapp-00733-gcj at 100% traffic with backend revision consent-protocol-00574-zcq. Runtime parity, deployment provenance, passkey-domain associations, and semantic health checks passed with no blocking classifications.

PR #6517, opened 2026-09-04
Grant anoushkauoc dev and UAT deploy dispatch

kushaltrivedi5:
## What

Adds `anoushkauoc` to `dev.manual_dispatch_users` and `uat.manual_dispatch_users`.
Production is untouched and stays at its existing six.

## Why

Merge access and deploy access are separate lists here, and holding one has never
implied the other. She was already on `main`'s review and merge-queue bypass lists,
so she could land a PR, and was on neither `manual_dispatch_users` list, so every
deploy attempt was refused by `scripts/ci/assert-governed-actor.py`.

GitHub reported this as "maintainer but not admin", which points at the wrong
system: GitHub's role check passed and the button was there. What refused her was
this repo's own actor gate reading a list she was not on.

## Verified

With the enforcement script, not by reading the file:

```
dev manual dispatch policy passed for actor 'anoushkauoc'.
uat manual dispatch policy passed for actor 'anoushkauoc'.
Refusing production workflow dispatch for 'anoushkauoc'.
```

The GitHub Environments add no second gate: `dev`, `uat` and `production` each carry
zero required reviewers, and their branch policies are satisfied because a
`workflow_dispatch` always runs the workflow definition from `main`.

## Also here: a docs-gate fix

Both docs path gates (`verify-doc-links.cjs`, `verify-doc-runtime-parity.cjs`)
resolved a backticked token straight to disk after stripping only trailing
punctuation. A precise citation like `consent-protocol/api/routes/consent.py:1357`
therefore failed to resolve and was reported unresolved, while the vaguer
`consent.py` passed — backwards, since a line number makes the pointer more
checkable, not less.

Both resolvers now strip an optional `:N` / `:N-M` suffix through one shared
helper. The check itself is unchanged: a citation naming a file that does not exist
still fails, line number or not. Verified in both directions against a probe doc
citing one real and one missing path with line suffixes.

### DIFF
--- a/config/ci-governance.json
+++ b/config/ci-governance.json
@@ -129,7 +129,8 @@
       "azfx",
       "Jhumma-hushh",
       "DamriaNeelesh",
-      "Roopmann30"
+      "Roopmann30",
+      "anoushkauoc"
     ]
   },
   "uat": {
@@ -149,7 +150,8 @@
       "Jhumma-hushh",
       "DamriaNeelesh",
       "parthmawai",
-      "imsharukhan"
+      "imsharukhan",
+      "anoushkauoc"
     ]
   },
   "production": {
--- a/scripts/verify-doc-links.cjs
+++ b/scripts/verify-doc-links.cjs
@@ -88,6 +88,16 @@ function resolveCandidate(baseFile, token) {
   return path.join(repoRoot, cleaned);
 }
 
+/**
+ * `file.py:1357` and `file.py:478-489` are references to file.py. Stripping the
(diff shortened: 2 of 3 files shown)


kushaltrivedi5:
Superseded by #6520. This PR added one person to the UAT list; #6520 fixes the cause, which was that `uat.manual_dispatch_users` had drifted to 9 names against a 14-name merge cohort, violating the equality invariant that `verify-deployment-environment-governance.py` already asserts. The verifier had been reporting it on every advisory run.

PR #6599, opened 2026-09-07
fix(deploy): use --update-headers when updating a Cloud Scheduler job

ankitkumarsingh1702:
The UAT backend deploy reported **failure on a healthy release**.

Run [34154374598](https://github.com/hushh-labs/hushh-research/actions/runs/34154374598) applied the migrations, deployed the backend, promoted traffic, verified exact-SHA provenance, passed the schema contract gate and semantic verification, skipped rollback, and tagged the release last known good — then died on the final step:

```
ERROR: (gcloud.scheduler.jobs.update.http) unrecognized arguments:
  --headers=Content-Type=application/json (did you mean '--clear-headers'?)
```

## Why it only breaks in steady state

`deploy/account-deletion/setup_cleanup_scheduler.sh` built one `COMMON_ARGS` array and passed it to **both** subcommands:

```bash
if gcloud scheduler jobs describe "${JOB_NAME}" ...; then
  gcloud scheduler jobs update http "${JOB_NAME}" "${COMMON_ARGS[@]}"   # --headers rejected
else
  gcloud scheduler jobs create http "${JOB_NAME}" "${COMMON_ARGS[@]}"   # --headers accepted
fi
```

Only `create` accepts `--headers`. `update` spells it `--update-headers` — the CLI renames it to express intent, alongside `--remove-headers` and `--clear-headers`.

So the script worked perfectly **the first time the job was created** and failed every time after. The failure mode appears only once the resource already exists, which is steady state and exactly when nobody is watching.

## One defect, four addresses

The idiom had been copied into three more schedulers, all carrying the same latent break:

| Script | |
|---|---|
| `deploy/account-deletion/setup_cleanup_scheduler.sh` | the one that fired |
| `deploy/one-email/setup_kyc_retention_scheduler.sh` | latent |
| `deploy/one-location/setup_retention_scheduler.sh` | latent |
| `deploy/gmail/setup_personal_information_request_monitor_scheduler.sh` | latent |

Each now sends `--headers` on create and `--update-headers` on update. The create arms are unchanged.

**Deliberately left alone:** `setup_investor_replenisher_scheduler.sh` and `setup_gcp_observability.sh` send no headers on either path — they post `--message-body '{}'` and never needed a `Content-Type`. Consistent, not broken.

## Ledger

Adds **R29** — *a flag shared between `create` and `update` breaks only on the second run*, plus the corollary that a workflow failing on a post-deploy configuration step is not a failed release, so check where in the step list it died before rolling anything back.

**R29's check was wrong when I first wrote it.** It asserted that every `create` arm sends headers, and flagged the two scripts above — a deliberate choice reported as a defect. It now asserts the create and update arms *agree*, which is the real invariant. A check that cries wolf trains people to ignore it, so I fixed the check rather than the scripts.

## Verification

```
bash -n on all four scripts: syntax OK
R29 check: no output (passes)
create arms still send headers: all four confirmed
```

Shell-only change under `deploy/`. No application code, no migration, no IAM, no secret, no workflow YAML.

### DIFF
--- a/.claude/skills/safe-changes/SKILL.md
+++ b/.claude/skills/safe-changes/SKILL.md
@@ -1150,3 +1150,61 @@ would have broken.
 The Check line must be a command that **runs in this repo** and returns
 something meaningful. Run it before committing the rule. A rule with a command
 that doesn't work here is worse than no rule.
+
+### R29 — A flag shared between `create` and `update` breaks only on the second run
+
+**Incident (2026-09-07, UAT backend deploy run 34154374598.)** The release was
+healthy — migrations applied, the backend deployed, traffic promoted, provenance
+verified, rollback skipped, the release tagged last known good — and the workflow
+still reported **failure**. The last step died on:
+
+```
+ERROR: (gcloud.scheduler.jobs.update.http) unrecognized arguments:
+  --headers=Content-Type=application/json (did you mean '--clear-headers'?)
+```
+
+`deploy/account-deletion/setup_cleanup_scheduler.sh` built one `COMMON_ARGS`
+array and passed it to **both** `gcloud scheduler jobs create http` and
+`gcloud scheduler jobs update http`. Only `create` accepts `--headers`; `update`
+spells it `--update-headers`. So the script worked perfectly the first time a
+job was created and failed forever afterwards — the failure mode appears only
+once the resource already exists, which is exactly when nobody is watching.
+
+The same idiom had been copied into three more schedulers — One Email KYC
+retention, One Location retention, and the Gmail personal-information-request
+monitor — all with the same latent break.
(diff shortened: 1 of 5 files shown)


ankitkumarsingh1702:
Superseded by #6617.

Three of the four scheduler scripts in this PR were fixed independently on `main` while this sat open. Only `deploy/one-email/setup_kyc_retention_scheduler.sh` still carried the bug, and #6617 fixes exactly that against current main.

This branch is 50 commits behind and its ledger addition collides with rule numbers `main` has since published (R29 is now a different rule, and R32-R34 belong to the migration-retry work). Rebasing it would mean renumbering for no benefit, since the code fix is already covered.

PR #6740, opened 2026-09-13
fix(auth): prevent account session recovery lockout

DamriaNeelesh:
## Summary

This fixes the customer lockout shown as “Reconnect to continue securely.”

### Root cause

Production had an incompatible cross-service deployment:

- Frontend SHA `a3679dc9` required `/api/account/session-status`.
- Backend SHA `c1f90f4a` did not contain that route, so restored sessions received `404 {"detail":"Not Found"}`.
- The frontend mapped that response to an unavailable session and permanently rendered the recovery wall.
- Production allowed a frontend-only lifecycle deployment and only checked `/` and `/health`.
- Client timeout budgets were also inverted (8s outer client vs 9s production / 20s local backend), and transient failures had no automatic recovery.

### Changes

- Centralized account-session timeout policy with enforced request/orchestration nesting.
- Keep protected UI fail-closed, but automatically retry transient verification failures with bounded backoff.
- Avoid duplicate forced-token probes when the backend is unavailable.
- Add regression coverage for missing-route/version-skew responses and automatic recovery.
- Make production deployment scope use the deployed-SHA lifecycle resolver.
- Add a post-promotion semantic contract gate requiring unauthenticated session-status to return 401; incompatibility triggers rollback.

### Incident mitigation

Production frontend traffic was restored through the governed rollback workflow to compatible revision `hushh-webapp-00114-zd2`. The rollback and health checks succeeded.

### Verification

- `npm run verify:account-session`: 330/330 tests passed
- Playwright account-session browser suite: 6/6 passed (Chromium, WebKit, mobile Chrome)
- `npx tsc --noEmit`: passed
- ESLint on changed files: passed
- Next production build: passed
- Deployment resolver tests: 11/11 passed
- Workflow YAML and diff validation: passed

The backend endpoint remains intentionally fail-closed until the governed compatible backend rollout; this PR prevents another frontend/backend split from being promoted.


DamriaNeelesh:
Tracking issue: #6741 — detailed RCA, acceptance criteria, rollout follow-up, and project ownership.

PR #6755, opened 2026-09-14
fix(deploy): shorten the Gmail monitor scheduler service-account ID

parthmawai:
## What was happening

Both `deploy-uat.yml` runs today (`34805625032`, `34812828573`) failed. The second one failed cleanly at the "Activate personal Gmail monitor" step:

```
ERROR: (gcloud.iam.service-accounts.create) INVALID_ARGUMENT: The account ID
"gmail-personal-monitor-scheduler" does not have a length between 6 and 30.
```

That default is a hardcoded 32-character string in `deploy/gmail/setup_personal_information_request_monitor_scheduler.sh`, mirrored into `deploy-uat.yml`'s env block and asserted in a contract test. This step runs unconditionally whenever the release hasn't already failed and the backend is in scope — so it blocks every UAT deploy that reaches it, not just the two today.

No service account was ever actually created under the bad name (gcloud rejects the request before creating anything), so this is a clean rename with nothing to migrate.

## The fix

- Shortened the default to `gmail-personal-monitor-sched` (28 chars) in the script, the workflow env var, and the test assertion — same three places that reference it, checked with a repo-wide grep for the literal string.
- Added the same 6–30-char + pattern validation the sibling `deploy/account-deletion/setup_cleanup_scheduler.sh` already has, so a bad value in the future fails fast with a clear message instead of a raw `gcloud` error three layers deep.
- Checked `deploy-production.yml` / `deploy-dev.yml` for the same hardcoded string — neither references it, so this is scoped correctly to UAT only.

## Test plan

- [x] `bash -n deploy/gmail/setup_personal_information_request_monitor_scheduler.sh` — syntax OK
- [x] `pytest consent-protocol/tests/test_gmail_information_request_release_contract.py` — passes
- [x] Ran the script locally with the *old* 32-char name — now rejected immediately with a clear message instead of reaching `gcloud`
- [x] Ran the script locally with a too-short name — rejected the same way
- [x] Ran the script locally with the new default — passes validation, proceeds to the real `gcloud` call, fails only on local IAM permission (expected: this machine isn't the CI deploy identity)


parthmawai:
Superseded by #6756 (merged), which fixes the same 30-char service-account-ID overflow with the identical replacement name (`gmail-personal-monitor-sched`) plus a second bug this PR didn't catch: `gcloud scheduler jobs update` needs `--update-headers`, not `--headers`, which would have broken every second deploy once the account-name fix landed. Closing without merging.

PR #6872, opened 2026-09-18
fix(calendar): avoid Windows-incompatible strftime directives in confirmations

parthmawai:
## Summary

`_display_time()` (`consent-protocol/hushh_mcp/agents/calendar/tools.py`) formatted Calendar confirmation messages using `%-d`/`%-I` — glibc/BSD strftime extensions that strip a leading zero from day/hour. Windows' C runtime raises `ValueError` on those directives, and the caller's `except (ValueError, ZoneInfoNotFoundError)` clause silently swallowed it into a raw ISO-timestamp fallback — so every Calendar confirmation message on a Windows-hosted server showed something like `2026-08-11T09:05:00+05:30` instead of `Tue, Aug 11 at 9:05 AM IST`.

Fixed by formatting the weekday/month and minute/period/zone separately via strftime (directives that are portable), then hand-building the day/hour without a leading zero.

This fix was written earlier in an uncommitted worktree on a branch that has since merged (#6797) — it never actually made it to `main`. Recovered and landed properly here with a regression test.

## Test plan
- [x] New regression tests: `test_display_time_does_not_use_platform_specific_strftime_directives`, `test_display_time_strips_leading_zero_from_single_digit_hour_and_day`
- [x] `pytest tests/agents/test_calendar_agent_tools.py` — 17 passed
- [x] `ruff check` on touched files — clean


parthmawai:
Closing — prod runs on Cloud Run (Linux-only), so this never crashes in the real deployment. The crash I observed was pytest/the dev server run natively on a Windows dev machine, not a production concern. Not worth carrying the extra code for a platform we don't deploy to or develop against.

PR #6890, opened 2026-09-18
fix(location): UX batch - rating drawer, avatar markers, settings type, dialogs, SMS scope, accordions

DamriaNeelesh:
Single PR for all pending location UX tasks, all verified end to end (see below). Mobile-first and responsive throughout; desktop rail layouts unchanged in measure.


parthmawai:
@DamriaNeelesh flagging: this PR left the capability graph in a state that blocks CI for any other PR touching `consent-protocol/` (e.g. #6895), via `scripts/generate_capability_graph.py`'s breaking-change gate:

```
RuntimeError: Capability graph has unacknowledged breaking changes: workflows:workflow.setup.location. Add an applicable workflow migration or exact-revision deprecation.
```

Traced it down: `hushh_mcp/agents/location/knowledge.yaml`'s content-derived digest changed, but its declared `package_version` is still `2` — no version bump, so the graph compiler has no migration path to map the old cursor version forward and correctly treats it as breaking.

Reproduces on a clean, unmodified checkout of current `main` (confirmed via `uv run python scripts/generate_capability_graph.py --check` from `consent-protocol/`), so it's not specific to any downstream PR.

Fix is likely just bumping `package_version` in `knowledge.yaml` (and adding the corresponding migration entry) to match whatever content change landed in this PR — didn't want to guess at that myself since I don't have context on the location knowledge package's intended migration semantics. Happy to help verify once it's in.

parthmawai:
@DamriaNeelesh correction to my earlier comment — I was wrong, sorry for the noise. That "breaking change" was a false alarm caused by Windows CRLF line-ending corruption on my own local checkout (`core.autocrlf=true` smudges `knowledge.yaml` to CRLF, and the digest function hashes raw bytes), not a real issue with this PR. Verified on a byte-exact LF checkout: `CapabilityGraphV1 is current` — no breaking change at all, your `location.brain` package is fine as-is.

The actual CI failure on my downstream PR (#6895) turned out to be something else entirely: my own branch's route-contract changes needed a `npm run build:voice-gateway` regen that I'd skipped. Nothing here needs any action from you — apologies again for the false flag.

PR #6897, opened 2026-09-18
fix(onboarding): tie back navigation to where each setup screen was opened from

parthmawai:
## Summary
Audit of top-shell back-button targets across the onboarding screens touched this session, per your request. Three real gaps found and fixed:

1. **`/one/setup/connections`** — back always retraced to the `/one/setup` hub, but since #6892, a brand-new user lands here directly without ever visiting the hub. Now hides its back arrow, matching the hub's own existing precedent (it hides its own back arrow for the same "no confirmed previous step" reason).
2. **`/one/setup/capabilities`** (the dashboard's optional-capabilities checklist, PR #6882) — fell into the generic per-capability breadcrumb handler. The chevron ended up pointing at `/one` (via an unrelated dismissed-state rewrite) while the "Setup" breadcrumb label still linked to the hub — two different destinations on one screen. Gave it its own explicit case: both now point home.
3. **Each capability's own setup workspace** (Gmail, Calendar, Location, KYC, Finance, RIA, Connected Systems) — back always went to the hub with no memory of actually being opened from the capabilities checklist, so connecting Gmail from that checklist and tapping back dropped you on the old hub instead of back on your "6 left" list. Reused the `?from=` origin mechanism already established elsewhere in this same file (`resolveCapabilitySetupBackHref`) rather than inventing anything new.

Also: while in that file, made the checklist's "Connect X" buttons a uniform width — labels vary a lot in length ("Verify RIA" vs. "Connect Calendar"), which staggered the trailing edge down the list.

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `npx eslint` clean on touched files
- [x] All 41 existing `top-shell-breadcrumbs` tests still pass unchanged (verified the new origin-aware fallback preserves old behavior when no `?from=` is present)
- [x] Added 3 new tests covering: retrace-to-checklist via `?from=`, hidden back arrow on the AI-choice step, capabilities-checklist back-to-dashboard
- [x] Full onboarding/navigation/utils suites green (353/355 — the 2 failures are the pre-existing, unrelated `setup-warm-transition.contract.test.ts` failure already confirmed failing on unmodified `main`)


parthmawai:
Added one more commit: reverted the greyscale-until-onboarded icon treatment per request — agent icons now render full color regardless of setup state. The underlying `isOnboarded`/`isCapabilityOnboarded()` plumbing is untouched, just unused for icon color, so bringing it back later is a one-line change.

